PR #1, opened 2017-07-11
Page structure

DominikaK:
Adds mkdocs.yml with very initial config and page structure.

Note: some page names are different than the originals in Confluence and some pages will be combined. Comments show which existing pages should be included in which files.

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -1,0 +1,146 @@
+site_name: eZ Platform Developer Documentation
+repo_url: https://github.com/ezsystems/developer-documentation
+copyright: "Copyright 1999-2017 eZ Systems AS and others"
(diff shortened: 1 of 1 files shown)


DominikaK:
@mere-teresa @kusztel

mere-teresa:
I like this plan, except that some files are missing. When I export from Confluence, I have more files than that, because of Subpages.

DominikaK:
@mere-teresa: which is precisely what my comment referred to:
> some pages will be combined. Comments show which existing pages should be included in which files.

mere-teresa:
https://gist.github.com/mere-teresa/7b950219b6b6d0d87f3acb91ceab863e it's a load of work.

PR #15, opened 2017-08-05
Missing assetic dump from the manual on internationalization

jdcrecur:
Nobody could answer this Q, i had to find it by trial and error.


andrerom:
Looks ok @yannickroger ?

DominikaK:
@jdcrecur, thanks for this and the other contributions!

PR #18, opened 2017-08-07
EZP-27719 fixing invalid path in solr instructions

tomaszmadeyski:
It is a fix for https://jira.ez.no/browse/EZP-27719

Using the path which was here before ends up in `Specified configuration directory server/ez/template not found!` while trying to run the command

### DIFF
--- a/docs/guide/search.md
+++ b/docs/guide/search.md
@@ -380,17 +380,18 @@ Secondly, copy configuration files needed for eZ Solr Search Engine bundle, *her
 
 ``` bash
 # Make sure to replace the /opt/solr/ path with where you have placed Solr
-mkdir -p /opt/solr/server/ez/template
-cp -R vendor/ezsystems/ezplatform-solr-search-engine/lib/Resources/config/solr/* /opt/solr/server/ez/template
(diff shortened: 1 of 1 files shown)


andrerom:
Alternative would be to have instructions on changing current working directory, however this is the same used by our docker file for solr, so +1.

andrerom:
@tomaszmadeyski the alternative would make it consistent with the bundle doc btw. So that would be better potentially, see: https://github.com/ezsystems/ezplatform-solr-search-engine/blob/master/README.md

tomaszmadeyski:
@andrerom done in 92e8756

andrerom:
Nice, did you test it as well? ;)

tomaszmadeyski:
> Nice, did you test it as well? ;)

Of course I did 😃

andrerom:
@DominikaK all yours, and I guess it should be applies on 1.7 and up like the other PR.

DominikaK:
Added here: https://github.com/ezsystems/developer-documentation/commit/ed07d419b40818106687789b22397562ef60b04e

PR #24, opened 2017-08-10
EZP-27760: Add information about not allowing both basic and session auth at the same time

jacek-foremski:
JIRA ticket: [EZP-27760](https://jira.ez.no/browse/EZP-27760)

Excerpt from JIRA: 
> There is a missing feature to have both basic auth and session auth working at the same time on the REST API (covered in ticket EZP-22192). Enabling both of them at the same time leads to issues (EZP-27184). There is a warning in the code regarding this, introduced here: EZP-27222.
However, there is no mention of this in the documentation (https://doc.ez.no/display/DEVELOPER/REST+API+Authentication) and we could make it more clear there.

I copied the warning from the code.

### DIFF
--- a/docs/api/general_rest_usage.md
+++ b/docs/api/general_rest_usage.md
@@ -194,7 +194,7 @@ To enable HTTP Basic authentication, you need to edit app`/config/security.yml`,
 
 !!! caution
 
-    Until https://jira.ez.no/browse/EZP-22192(https://jira.ez.no/browse/EZP-22192) is implemented, enabling basic authentication in REST will prevent PlatformUI from working.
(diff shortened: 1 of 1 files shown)


docs/api/general_rest_usage.md
```diff
@@ -192,6 +192,10 @@ More information
 
 To enable HTTP Basic authentication, you need to edit app`/config/security.yml`, and add/uncomment the following block. Note that this is enabled by default.
 
+!!! caution
+
+    Until https://jira.ez.no/browse/EZP-22192(https://jira.ez.no/browse/EZP-22192) is implemented, enabling basic authentication in REST will prevent PlatformUI from working.
```
DominikaK: Could you fix the link formatting here to `[https://jira.ez.no/browse/EZP-22192](https://jira.ez.no/browse/EZP-22192)`?
Also, props for the use of a caution box ;)

natanael89: Fixed.

PR #26, opened 2017-08-21
EZP-27780: Remove unused table from legacy SearchIndex

andrerom:
Updated upgrade doc for 6.11 for fixing schema.

### DIFF
--- a/docs/releases/updating_ez_platform.md
+++ b/docs/releases/updating_ez_platform.md
@@ -133,13 +133,13 @@ in `app/AppKernel.php`.
 
 ## 4. Update database
 
-This step is only relevant for some releases:
+These steps are only relevant for some releases:
 
 **ezsearch_return_count table removal**
 
 V1.11
 
-v1.11.0 removes `ezsearch_return_count` table, aligning with eZ Publish legacy since 5.4/2014.11 to avoid issues for those using upgrade install with our withouth legacy bridge. To fix this apply the following database update script if your install is not an eZ Publish upgrade where it was already removed:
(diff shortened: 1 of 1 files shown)


DominikaK:
@andrerom: I reformulated the text for clarity. Could you take a look if the meaning has not been distorted?

docs/releases/updating_ez_platform.md
```diff
@@ -133,13 +133,13 @@ in `app/AppKernel.php`.
...
 
-This step is only relevant for some releases:
+These steps are only relevant for some releases:
 
 **ezsearch_return_count table removal**
 
 V1.11
 
-v1.11.0 removes `ezsearch_return_count` table, aligning with eZ Publish legacy since 5.4/2014.11 to avoid issues for those using upgrade install with our withouth legacy bridge. To fix this apply the following database update script if your install is not an eZ Publish upgrade where it was already removed:
+v1.11.0 removes the `ezsearch_return_count` table, which had been obsolete in eZ Publish legacy since 5.4/2014.11. This avoids issues which would occur when you upgrade using legacy bridge. Apply the following database update script if your installation has not had the table removed by an earlier eZ Publish upgrade:
```
andrerom: table was removed, not obsoleted. so wording there is a bit wrong. otherwise it looks ok

andrerom:
Note, the update SQL is missing form the release: https://github.com/ezsystems/ezpublish-kernel/compare/v6.11.0...6.11

So will need to resolve that with @bdunogier before we can merge this.

andrerom:
Issue above is resolved, the final build contains the files needed for this.

@DominikaK you can go ahead with this ;)

mere-teresa:
To be sure, could you confirm the _affected versions_ are those on the JIRA Issue, @andrerom?

>  dev-master, 1.7.4, 1.10.1

andrerom:
Yes, but this is upgrade doc telling people to upgrade, so it can only happen when upgrading to 1.11. So this is therefor pointing towards 1.11 branch.

PR #27, opened 2017-08-21
Create Translations with SiteAccess

jdcrecur:
This would have so helpful to me, it joins the 3 main dots of translation, siteaccess and permissions in one go, linking to the rel. full pages when required. I'm not sure if i have the naming of SiteAccess correct in all cases though :) Not sure if the siteaccess is only the section siteaccess or also the sections under system?


emodric:
This looks good, it's simple enough and could help people in simpler cases. What I would do though is write the doc in a way that we don't need to readd and then remove `site` siteaccess, but only mention at the end that users can now remove it as it superfluous. This way you can get rid of the whole "I can't access my site at all any more" situation.

docs/cookbook/Translations with SiteAccess
```diff
@@ -0,0 +1,119 @@
+## Description
+
+Being new to eZ and combining the translated content with multiple site access successfully can be a little tricky to get you head around.
```
mere-teresa: Please, could you:
eZ to replace by _eZ Platform_
site access need First Uppercase notation

docs/cookbook/Translations with SiteAccess
```diff
@@ -0,0 +1,119 @@
...
+
+To successfully achieve this for the most typical setups you will need all 3 of following items to be considered:
+1. Registering your translations in the database via the eZ admin.
+1. Creating at least two site accesses to deliver the right translated content at the right time in the `ezplatform.yml`.
```
mere-teresa: Same thing "site accesses" needs to be First Upper Cased, please. (I will not repeat that every time, so I consider that you will do it for every appearance of these words.)

docs/cookbook/Translations with SiteAccess
```diff
@@ -0,0 +1,119 @@
...
+1. Setting the correct permissions for the anonymous user to read the site accesses.
+
+Without all 3 steps complete your site access will either not work or you will be left with duplicate content from an SEO perspective.
+
+## Step 1 - Create your new translation
+Here is the full guide on [Internationalization](../guide/internationalization.md)
+1. Log into your ez admin
```
mere-teresa: "eZ Platform Admin"

docs/cookbook/Translations with SiteAccess
```diff
@@ -0,0 +1,119 @@
...
+## Step 2 - Creating the new site accesses
+Here is the full guide on [SiteAccess](../guide/siteaccess.md)
+
+The most typical setup for a site with translated content is to map the base of the domain to one language and using the first segment of the URI to match to another. For example:
+- www.mysite.com to match to English
+- www.mysite.com/fr to match to French
+
+To achieve this you will need to create at least two site accesses in your `ezplatform.yml` file.
+
+In the first bit of this working example, we declare `en` to be the default site access meaning if nothing else matches use the `en` site access. We then list the available site accesses for this site as well as add them both to a common group `site_group` (this will be used for shared settings such as api keys and cache locations etc.).
```
mere-teresa: API => uppercase, please (acronym)

jdcrecur:
I added the changes. I'm not the name change is the best option though, the name indicates you can create a translation via a SiteAccess.

andrerom:
True, it could be something else, like `How-to setup a multi language site` or `How-to setup multi language SiteAccesses and corresponding Translations`, first one might imply the form and scope of the article might need to adjust a bit.

DominikaK:
Merged in https://github.com/ezsystems/developer-documentation/commit/fcc11ab04e45badcbc73e380f923ec04e306adf2

PR #36, opened 2017-08-23
Applying the webdesign theme to the 1.8 version

mere-teresa:


DominikaK:
I think you only need https://github.com/ezsystems/developer-documentation/pull/36/commits/39749e0b7f6ae79311dbcb97b9d82e322714086e from this, but please don't trust me automatically :)

mere-teresa:
Do we maintain an 1.8 branch?

DominikaK:
I don't think we need to, it's not lts, and it's three versions back from the latest ft (at the moment).

PR #39, opened 2017-08-28
Add a note to pull the tag and not the branch

Plopix:
I got caught by that, I pulled the branch instead of the tag.
It is too easy to do that mistake and then you get issues when updating with composer and you don't understand.

This note could improve the developer procedure.

### DIFF
--- a/docs/releases/updating_ez_platform.md
+++ b/docs/releases/updating_ez_platform.md
@@ -14,7 +14,7 @@ If you intend to skip a version (for example, update directly from v1.3 to v1.5 
 
 # Update procedure
 
-## 1. Check out a version
+## 1. Check out a tagged version
 
 **1.1.** From the project's root, create a new branch from the project's master, or from the branch you're updating on:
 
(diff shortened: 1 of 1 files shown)


SylvainGuittard:
Very good point! Thanks @Plopix

docs/releases/updating_ez_platform.md
```diff
@@ -48,6 +48,8 @@ or
 git pull ezplatform-ee <version>
 ```
 
+> Don't forget the `v` here, you want to pull the tag `<version>` and not the branch `<version>` (i.e: `v1.11` and not `1.11`)
```
glye: The example should say `v1.11.0` since version tags are normally triplets. In addition to the `v` this also helps separating them from branches.

Good PR, thanks!

DominikaK: It would be also cool to format this as a "Pretty Tip Box"(tm) (like we have at the bottom of the front page: http://doc.ezplatform.com/en/master/). This is the way to do it:

```
!!! tip

    Don't forget the `v` here...
```

DominikaK:
Thanks for the smart contribution, @Plopix. Would you able to change the PR to go into branch 1.7 instead of master? (We'll then merge it forward).

Plopix:
Can you cherry pick? if not I will do that tomorrow ;-) Thank you!

DominikaK:
Added in https://github.com/ezsystems/developer-documentation/commit/7060398fb337cce42454947f4f1b8bd7ebc0ea1e

glye:
@DominikaK Nitpick: The version uses 3 or 4 numbers, the branch uses two normally. `(i.e: v1.11.0 and not 1.11)`

PR #41, opened 2017-08-29
Remove explicit mention of latest version

DominikaK:
Removes explicit mention of the latest version and replaces it with link to ezplatform.com

This way there is not problem with linking to release notes from a different doc version, and the pages will not have to be updated with every release.

(Idea this comment: https://github.com/ezsystems/developer-documentation/pull/38)

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -2,7 +2,7 @@
 
 ## Welcome to the eZ Platform Developer documentation!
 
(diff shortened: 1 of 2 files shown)


mere-teresa:
You can merge it, if @SylvainGuittard is not against this!

SylvainGuittard:
@mere-teresa why should I be against this change? :D

PR #45, opened 2017-08-30
resource does not provide any value

larseirik:
Deleted file to clean up the cookbook/ structure.


yannickroger:
You should change the commit message to something like `Deleted useless cookbook section`

PR #48, opened 2017-08-31
Add cluster migration section

glye:
Adds doc about cluster migration, which was added in https://jira.ez.no/browse/EZP-25946

Modifies http://doc.ezplatform.com/en/master/guide/clustering/

### DIFF
--- a/docs/guide/clustering.md
+++ b/docs/guide/clustering.md
@@ -146,7 +146,7 @@ rewrite "^/var/([^/]+/)?storage/images(-versioned)?/(.*)" "/app.php" break;
 
 ## Migration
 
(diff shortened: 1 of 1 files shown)


docs/guide/clustering.md
```diff
@@ -143,3 +143,33 @@ RewriteRule ^/var/([^/]+/)?storage/images(-versioned)?/.* /app.php [L]
 ```
 rewrite "^/var/([^/]+/)?storage/images(-versioned)?/(.*)" "/app.php" break;
 ```
+
+## Migration
+
+If you are migrating an existing single server site to a cluster, and not setting up a cluster from scratch, you need to migrate your files. Once you have configured your binarydata and metadata handlers, you can run the `ezplatform:io:migrate-files` command. You can also use it when you are migrating from one data handler to another, e.g. from NFS to Amazon S3.
```
DominikaK: "If you are migrating an existing single-server site to a cluster..."
(hyphen in single-server)

DominikaK:
Thanks, @glye, other than the tiniest of nitpicks, a perfect addition!

glye:
😊 Thanks, nit picked.

PR #55, opened 2017-09-06
Add link to eZ Publish (Legacy, from Version 3.x to 5.x) documentation

dfritschy:
Given the official support that the Legacy Kernel and Legacy Bridge have received, it feels right that this documentation should be mentioned on the index page.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -20,7 +20,9 @@ This is the center for eZ Platform documentation for developers and system admin
 
 **Need Release Notes & older versions? See [Releases](releases/releases.md)**
 
-**Looking for eZ Publish (Legacy, from version 3.x to 5.x.) documentation? See [Legacy documentation](https://doc.ez.no/eZ-Publish)**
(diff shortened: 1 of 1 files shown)


docs/index.md
```diff
@@ -20,6 +20,8 @@ This is the center for eZ Platform documentation for developers and system admin
 
 **Need Release Notes & older versions? See [Releases](releases/releases.md)**
 
+**Looking for eZ Publish (Legacy, from version 3.x to 5.x.) documentation? See [Legacy documentation](https://doc.ez.no/eZ-Publish)**
+
```
SylvainGuittard: This URL works for eZ Publish 3.x to 4.x. 
eZ Publish Platform (5.x) documentation is here: https://doc.ez.no/display/EZP/eZ+Publish+5.x+Developer+Documentation

dfritschy: You are right, we probably need two links, one for eZ Publish Legacy and one for eZ Publish 5.

PR #57, opened 2017-09-15
Audit - multisite p1

DominikaK:
https://jira.ez.no/browse/EZP-27852

Covers:
- Added full example of configuration
- Examples consistent throughout the page
- Changed heading levels and names for better browseability
- General text polishing

### DIFF
--- a/docs/guide/multisite.md
+++ b/docs/guide/multisite.md
@@ -53,15 +53,14 @@ In the example above, the Media and Images folders will be accessible using thei
 ## Configuration example
 
 To see how multisite can be used, let's look at an example of two sites using the same eZ Platform instance: a general company site and a site for a specific event.
-Separate SiteAccess groups are set up for the two sites:
+Separate SiteAccesses are set up for the two sites:
 
 ``` yaml
 ezpublish:
     siteaccess:
         list: [site, event]
         groups:
-            site_group: [site]
-            event_group: [event]
+            site_group: [site, event]
         default_siteaccess: site
         match:
             URIElement: 1
@@ -82,12 +81,12 @@ This is your content structure:
         └── Logos
 ```
 
-You can now set the root level for `event_group` to only access the "Event" Location and its sub-items:
+You can now set the root level for `event` to only access the "Event" Location and its sub-items:
 
 ``` yaml
 ezpublish:
     system:
-        event_group:
+        event:
             content:
                 tree_root:
                     # LocationId of "Event"
@@ -100,7 +99,7 @@ In the next step, you need to reuse some content between sites, for example "Log
 You can allow the event site to access them, even though they are in a different part of the tree, via `excluded_uri_prefixes`:
 
 ``` yaml
-        event_group:
+        event:
             content:
                 tree_root:
                     location_id: 57
@@ -120,7 +119,7 @@ You can configure the Index Page separately for each SiteAccess. Place the param
 
 ezpublish:
     system:
-        event_group:
+        event:
             # The page to show when accessing IndexPage (/)
             index_page: /EventFrontPage
 ```


docs/guide/multisite.md
```diff
@@ -2,109 +2,189 @@
...
 ```
 
-##### `content.tree_root.location_id`
+### Configuration parameters
 
-This parameter sets the Location ID of the content root. The API is jailed within this root, and nothing above this level will be accessible by default. This parameter can be filtered using the `excluded_uri_prefixes` parameter described below.
+#### `location_id`
 
-##### `content.tree_root.excluded_uri_prefixes`
+`content.tree_root.location_id` sets the Location ID of the content root. The API is limited to ("jailed" within) this root, and nothing above this level will be accessible by default. This parameter can be filtered using the `excluded_uri_prefixes` parameter.
```
andrerom: This is incorrect actually.

"API" knows close to nothing about this, so there is no "jail". "Site handling" is done above in the architecture, so in controllers and view logic. In kernel often referred to as MVC layer, which actually does not deal with M:Model which is done by API, but it does deal with V:View and C:Controller. 

But that is to kernel specific here, we need a higher level way to describe it I think, it's been a topic of discussion before.

In this specific case you can say it like this perhaps: ``content.tree_root.location_id` sets the Location ID of the content root of your "site", it can be filtered using the `excluded_uri_prefixes` parameter. These settings are taken into account several places for you, for instance: Url Alias lookups for incomming web requests, and url alias path generation in twig templates. In addition you'll need to take these into account when generating site paths in your layout, or other places rendering site/tree structure.`


/cc @bdunogier

docs/guide/multisite.md
```diff
@@ -2,109 +2,189 @@
...
-            #The page to show when accessing IndexPage (/)
-            index_page: /yourURIPage
+        kids_group:
+            # The page to show when accessing IndexPage (/)
+            index_page: /KidsFrontPage
 ```
 
 If not specified, the `index_page` is the configured content root.
+
+### Multisite with different repositories
```
andrerom: A bit unsure about the title.


A couple of things makes me unsure, as this page is about multi site in single database, and the limitation we are talking abut is in regards to multi repository setup. Do we have doc for the latter? Only thing I found was: https://doc.ezplatform.com/en/1.11/guide/repository/#multi-repository-setup


Cause if there where I would maybe clarify the relationship in the intro at the top of this page, and then here have a title ala `### Known limitations when using with Multi Repository setup`

docs/guide/multisite.md
```diff
@@ -2,109 +2,189 @@
...
+
+!!! caution
+
+    Only one repository (database) can be used per domain.
+    This does not prohibit using subdomains, but they must all use the same repository, for example:
```
andrerom: `This does not prohibit using different repositories on different subdomains, but when using URI matching for multi site setup all site access sharing domain also need to share repository, for example:`

Additonal suggestion for putting something like this at end of this section: 
> Invalid configuration will cause problems for different parts of the system such as: backend UI, REST interface and other non site access aware symfony routes such as `/_fos_user_context_hash` used <link>http cache</link>.

docs/guide/multisite.md
```diff
@@ -2,109 +2,189 @@
...
+
+    - `ez.no` domain can use `ez.repo`
```
andrerom: Now that I see this, instead of `.` in the repo names here which can be confused for levels in Symfony config tree, lets rather use `_`. So here `ez_repo` and on `doc_repo` instead of `doc.repo`.

docs/guide/multisite.md
DominikaK: You are right, we lack a section on "multisite with multiple repositories". I'll add it to our to-do list, but I think this page would be the most logical place for it (then we can include this subsection into it).

andrerom: One lower hanging fruit for now could be to:
1. name this `### Limitations when using with Multi site URI matching with Multi Repository setup`
2. on first mention of multi repository use the link above even if it is super minimal on the concept :)

docs/guide/multisite.md
```diff
@@ -2,109 +2,201 @@
...
 ```
 
-Now you can use this view configuration (stored e.g. in a `views.yml` file):
+This is your content structure:
+
+```
+├── General
+│   └── Articles
+│       └── Mammals
+├── Children
```
andrerom: Hmm, the visualization here seems to not include location 2, often called "Content", it would typically be the parent of the site roots, meaning they will both pop up in  content section of backend UI.

docs/guide/multisite.md
```diff
@@ -2,109 +2,202 @@
 
 ## Introduction
 
-With eZ Platform it is possible to serve multiple, completely different sites using one eZ Platform instance and database.
+You can serve multiple different sites using one eZ Platform instance and database.
 
-Each site will have its own content root, at a lower level than the default one (Location ID 2). One backoffice can be used for each site, but it is also possible to use a global one.
+Each site will have its own content root, at a lower level than the default one (Location ID 2).
+You can use one global back office for all sites, or a separate back office for each site.
```
mere-teresa: Why do we use a two-words  `backoffice` now?

docs/guide/multisite.md
```diff
@@ -2,109 +2,202 @@
...
                     # Prefixes are not case sensitive.
                     excluded_uri_prefixes: [ /media, /images ]
 ```
 
-##### `content.tree_root.location_id`
+### Configuration parameters
 
-This parameter sets the Location ID of the content root. The API is jailed within this root, and nothing above this level will be accessible by default. This parameter can be filtered using the `excluded_uri_prefixes` parameter described below.
+The two configuration parameters, `location_id` and `excluded_uri_prefixes` are taken into account in several places,
+for example in URL alias lookups for incoming web requests, and URL alias path generation in Twig templates.
```
mere-teresa: I am surprised to found no First Uppercased word here because there are many domain-specific words in this sentence.
In addition to that, I think that it could be interesting to add a link to the URL alias path generation documentation from Twig.

docs/guide/multisite.md
```diff
@@ -2,109 +2,202 @@
...
+In addition you need to consider them when generating site paths in your layout, or other places rendering site/tree structure.
 
-##### `content.tree_root.excluded_uri_prefixes`
+#### `location_id`
```
mere-teresa: `location_id` is treated on the previous paragraph. What about changing the order of these paragraphs?

docs/guide/multisite.md
```diff
@@ -2,109 +2,202 @@
...
+!!! note
 
-### Multisite Design
+    Leading slashes (`/`) are automatically trimmed internally, so they can be ignored.
 
-eZ Platform does not apply a [Legacy template fallback](https://doc.ez.no/display/EZP/Legacy+template+fallback) like eZ Publish did. You can, however, have different designs in your multisite installation if you organize the view configuration with the use of siteaccesses.
+## Configuration example
 
-You can see how this is done on the example of a multisite that contains two blogs with different designs, but sharing some templates. Let's assume you have a siteaccess called `second_site`. You can organize your templates in the following folder structure:
+To see how multisite can be used, let's look at an example of two sites using the same eZ Platform instance: a general site and a children's site.
```
mere-teresa: Multisite if it's a feature could have its own First Uppercased word.

mere-teresa: A children's site is for kids? If so, maybe "grownup's site and children's site"
More seriously I propose: "the main website and second website" or "the main website and a sub-topic website"

docs/guide/multisite.md
```diff
@@ -2,109 +2,202 @@
...
-             |- blog_post.html.twig
+ezpublish:
+    siteaccess:
+        list: [site, kids]
+        groups:
+            site_group: [site]
+            kids_group: [kids]
+        default_siteaccess: site
+        match:
+            URIElement: 1
```
mere-teresa: Please, could you add the filename where to put this configuration?

docs/guide/multisite.md
```diff
@@ -2,109 +2,202 @@
...
+│   │       └── Mammals
+│   └── Children
+│       └── Articles
+│           └── Lions
+└── Media
+    └── Images
+        └── Animal-photos
+```
+
+You can now set the root level for `kids_group` to only access the "Children" Location and its sub-items:
```
mere-teresa: Is it **really** for kids?
It's a very confusing idea because we are talking about content and parent content to manage the Tree Content understanding.
Maybe a "main site and an events website" for a music band is more understandable.
Or you can use the Dog's World magazine example with a special website for an event, a dog's type, something like that.

docs/guide/multisite.md
```diff
@@ -2,109 +2,202 @@
...
-                full:
-                    blog_post:
-                        template: "second_site/full/blog_post.html.twig"
-                        match:
-                            Identifier\ContentType: [blog_post]
+        kids_group:
+            content:
+                tree_root:
+                    # LocationId of "Children"
+                    location_id: 57
```
mere-teresa: Root + children could lead to confusion (see my previous comment about children's site)

docs/guide/multisite.md
DominikaK: Because [there](https://en.wiktionary.org/wiki/back_office) [is](http://www.dictionary.com/browse/back-office?s=t) [no reason](https://en.wikipedia.org/wiki/Front_and_back_office_application) to write it as one word ;)

docs/guide/multisite.md
DominikaK: I'm not sure which terms you mean by domain-specific... plus, I'm not a fan of capitalizing every domain term unless it's a proper name or an eZ-only name (like SiteAccess)
Re: Twig doc, that would be a good idea, but it seems there is no page like this in their doc.

docs/guide/multisite.md
DominikaK: The previous paragraph is an overview of the two available parameters, while this and the next section cover the parameters in detail.

docs/guide/multisite.md
DominikaK: AFAIK there is no single fixed place where config like this goes. It depends on each user's setup, so enforcing one specific file could be misleading.

docs/guide/multisite.md
DominikaK: Again, I'm don't fancy overusing upper case, especially for terms that are not proper names.

Re: "kids", you are right, I didn't think it could be misleading when mixed with the use of "parent" throughout the page. I changes the examples to refer to a general and "Event" site.

docs/guide/multisite.md
mere-teresa: https://doc.ezplatform.com/en/1.7/guide/content_rendering/#adding-links

or 

https://symfony.com/doc/current/templating.html#linking-to-pages

bdunogier: Domain specific names are those who have a particular meeting in our domain (web CMS).

Content and Location have a special meaning for us. Think of them as the citizens of our domain :)

DominikaK: Yes, I do try to capitalize those (at the risk of sounding like a bad fantasy novel), but I don't see any specific domain terms in this section.

docs/guide/multisite.md
```diff
@@ -2,109 +2,202 @@
...
 
-### Multisite Design
+    Leading slashes (`/`) are automatically trimmed internally, so they can be ignored.
 
-eZ Platform does not apply a [Legacy template fallback](https://doc.ez.no/display/EZP/Legacy+template+fallback) like eZ Publish did. You can, however, have different designs in your multisite installation if you organize the view configuration with the use of siteaccesses.
+## Configuration example
 
-You can see how this is done on the example of a multisite that contains two blogs with different designs, but sharing some templates. Let's assume you have a siteaccess called `second_site`. You can organize your templates in the following folder structure:
+To see how multisite can be used, let's look at an example of two sites using the same eZ Platform instance: a general company site and a site for a specific event.
+Separate SiteAccess groups are set up for the two sites:
```
andrerom: Any reason why we opted for separate groups here? It1s valid, but if the two sites share config it's more common if they share group as well. On the other side if you plan to translate the times then groups are indeed good to represent the sites, while access will represent language.

So in sum: Maybe somehow explain why we opt for using separate group here for those that might wonder.

DominikaK: No particular reason for using groups except me being too copy-paste-happy.

docs/guide/multisite.md
mere-teresa: The only question remaining is here: using children as a Category name is it ok?

DominikaK: This has been changed to "Events" already, take a look at the last commit :)

PR #60, opened 2017-09-22
Add global reference page

DominikaK:
Adds a global page with links to all references that are distributed in other pages.

/cc @janit, this is based on your suggestion.

### DIFF
--- a/docs/references.md
+++ b/docs/references.md
@@ -1,0 +1,19 @@
+# References
+
+- API:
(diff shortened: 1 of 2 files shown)


janit:
Good stuff :)

andrerom:
Nice! As for more suggestions I guess we have some reference / glossary pages around in confluence and 5.x/4.x doc that might serve as basis for further topics to list up.

PR #63, opened 2017-09-28
"Knows limitations" section in /Guide/HTTP Cache/Usage/Context-aware HTTP cache/User hash generation

adamwojs:
> JIRA: https://jira.ez.no/browse/EZP-27791

# Description

This PR adds "Knows limitations" section in /Guide/HTTP Cache/Usage/Context-aware HTTP cache/User hash generation with paragraph about https://github.com/ezsystems/ezpublish-kernel/pull/2096

### DIFF
--- a/docs/guide/http_cache.md
+++ b/docs/guide/http_cache.md
@@ -687,6 +687,12 @@ set req.http.X-User-Hash = "b1731d46b0e7a375a5b024e950fdb8d49dd25af85a5c7dd5116a
 
 **5.** Restart the Varnish server and everything should work fine.
 
+##### Known limitations of the user hash generation
+
+A few limitations are known:
+
+* If you are using uri-based SiteAccesses matching, the default SiteAccess on the domain needs to point to the same repository, because `/_fos_user_context_hash` is not SiteAccess-aware by default (see `ezpublish.default_router.non_siteaccess_aware_routes` parameter). Varnish does not have knowledge about SiteAccesses, so it won't be able to get user content hash if the default siteaccess relies on URI.  
+
 ##### Default options for FOSHttpCacheBundle defined in eZ
 
 The following configuration is defined in eZ by default for FOSHttpCacheBundle. You may override these settings.


docs/guide/http_cache.md
```diff
@@ -687,6 +687,12 @@ set req.http.X-User-Hash = "b1731d46b0e7a375a5b024e950fdb8d49dd25af85a5c7dd5116a
 
 **5.** Restart the Varnish server and everything should work fine.
 
+##### Known limitations of the user hash generation
+
+A few limitations are known:
+
+* If you are using uri based site accesses matching, repository must be shared across one domain, because `/_fos_user_context_hash` is non siteaccess aware by default (see `ezpublish.default_router.non_siteaccess_aware_routes` parameter).
```
andrerom: Almost, but rule can be made more correct with:
`If you are using uri based site accesses matching, default siteaccess on the domain needs to point to same repository, because `

andrerom: Can maybe also add info in some form that Varnish does not have knowledge about siteaccesses, or that it might be part of uri, so this is the reason it has to be this way for this feature.

andrerom:
@DominikaK / @mere-teresa To expand on comment above this, There should be a general rule somewhere that this can link to, for instance in multi repo doc that one domain should always **only** use one repository. Because of native symfony routes like this, backend UI (`/ez`), and because of REST API which is also not aware of siteaccess in URI. 

To expand on this as we need to make sure people don't think they can't use sub domains, valid:
- ez.no => ez.repo
- doc.ez.no => doc.repo

Not valid:
- ez.no => ez.repo
- ez.no/doc => doc.repo

adamwojs:
PR updated with @andrerom suggestions

DominikaK:
@andrerom, I've added a mention of this in the (conveniently) open PR about multisite: https://github.com/ezsystems/developer-documentation/pull/57/commits/161b791f3cf3ab17fac2d7d9e4e52225db5b21f5

We can add a link to it once that PR is merged.

andrerom:
> @andrerom, I've added a mention of this in the (conveniently) open PR about multisite: 161b791

ok, for self reference that would be: https://github.com/ezsystems/developer-documentation/pull/57

PR #64, opened 2017-09-29
Removed duplicated Float Field Type content in the middle of Image Field Type reference

bdunogier:
It seems that some of the content from the Float Field Type ended up in the middle of the Image one. This PR removes it.

P.S. The whitespace changes are most likely due to PhpStorm autofixing them.


andrerom:
+1, seems 1.7 branch is also affected btw.

DominikaK:
Thanks for noticing this, @bdunogier, something must have got mixed up when moving the content from Confluence.

PR #73, opened 2017-10-13
Add 1.12 release notes

DominikaK:
(possibly still need to change the  release date)

### DIFF
--- a/docs/releases/ez_platform_v1.12.0.md
+++ b/docs/releases/ez_platform_v1.12.0.md
@@ -26,6 +26,21 @@ You can now remove translations from Content item Versions through the PHP API.
 
 See the [section on deleting translations](../api/public_php_api.md#delete-translations) for more information.
 
+You also have a new endpoint available for deleting a single Version, see [EZP-27864](https://jira.ez.no/browse/EZP-27864) for more information.
+
+#### Improved Security for password storage
+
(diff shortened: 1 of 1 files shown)


andrerom:
There are two things more from the [release note highlights](https://github.com/ezsystems/ezplatform/releases/tag/v1.12.0-beta2) that can be covered too:
```
#### Improved Security for password storage

1.12 introduces and enables more secure user passwords hashing using bcrypt now by default, 
 and is future proofed for new hashing formats being added to PHP, like Argon2i coming with PHP 7.2.

This feature is added both in eZ Platform as well as the accompanying eZ Publish legacy 2017.10 release for projects looking to migrate to a newer version of Platform and take advantage of the new features.

#### Improved Varnish performance

This release switches default HTTPCache usage to use ezplatform-http-cache package, which uses Varnsih xkey allowing: soft purge, better cache clearing logic and longer ttl.

For Varnihs users be aware thuis change implies new VCL and requriment for varnish-moduels packaage, see [github release for more info](https://github.com/ezsystems/ezplatform/releases/tag/v1.12.0-beta2). 
```

Todo:
- [ ] Update link to stable once tagged

DominikaK:
Thanks, @andrerom, I copied the BC break notice to this page, so that it is searchable with the rest of the doc.

PR #75, opened 2017-10-16
[EZP-25922] Add doc for how cache ttl for error pages can be configured

andrerom:
> Issue: https://jira.ez.no/browse/EZP-25922


andrerom:
Nice, thanks @DominikaK, all good from my side :)

PR #77, opened 2017-10-17
[PAPI] Fixed Content\Query ValueObject type hints

alongosz:
This PR includes:
- [x] Fix for erroneus type hinting of `Content\Query` ValueObject
- [x] Improve `Search` doc so user can actually find something.


alongosz:
Review ping @DominikaK

docs/api/public_php_api.md
```diff
@@ -838,15 +838,15 @@ You can also use the [`ContentTypeIdentifier`](http://apidoc.ez.no/sami/trunk/NS
...
+use eZ\Publish\API\Repository\Values\Content\Query;
 
 // [...]
 
 $query = new Query();
 $query->filter = new Criterion\LogicalAnd(
     array(
-        new Criterion\ParentLocationId( 2 ),
-        new Criterion\ContentTypeIdentifier( array( 'article', 'folder' ) )
+        new Criterion\ParentLocationId( 1 ),
```
andrerom: why did you change this? 2 is content root.

alongosz: On clean install it doesn't find anything as `LocationId = 2` refers to  `eZ Platform` folder which doesn't have any sub items. 
On the other hand, with this change we're finding
```
[1] eZ Platform
[41] Media
[45] Setup
```
(content ids in brackets)

but if the idea was here to find anything under `eZ Platform`, then the alternative improvement would be to hint to the user in the doc that he needs to add some sub items. 

Background reasoning: When you're new to API and trying to do something, it's best that you see some results IMHO, otherwise you think that you did something wrong and it doesn't work.

andrerom: I assume it is meant to look for content as it includes `article` in content type criteria, but you are right that this should be changed to work on any kind of demo install.

alongosz: Ok, it even makes more sense to hint that to user at there beginning of `Search` chapter here. So user either should add some content (folder, articles) or use demo install to find anything.
I'll adjust PR.

alongosz:
There is a small flaw in my initial plan to hint about demo, @andrerom :)
Demo actually doesn't have Article content type :laughing: 

So I've just added in 7acd52d a note on a need to create content.

I've also restrained myself this time from changing long array syntax as it exists everywhere and is out of scope of PR ;)

PR #78, opened 2017-10-17
Increased password hash length

glye:
Doc-ing DB update requirement in 1.12, ref https://jira.ez.no/browse/EZP-24744

### DIFF
--- a/docs/releases/updating_ez_platform.md
+++ b/docs/releases/updating_ez_platform.md
@@ -263,8 +263,6 @@ If you are upgrading to v1.8 you need to create this folder and assign it to a n
 
 **Increased password hash length**
 
-V1.12
-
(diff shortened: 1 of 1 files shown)


docs/releases/updating_ez_platform.md
```diff
@@ -261,6 +261,22 @@ If you are upgrading to v1.8 you need to create this folder and assign it to a n
     form_builder.upload_folder.section_identifier: <section identifier>
 ```
 
+**Increased password hash length**
+
+V1.12
```
DominikaK: With current doc versioning, we don't need lines like this anymore (they will be removed from the rest of the file soon(TM))

glye: Expunged.

PR #80, opened 2017-10-23
EZP-28083: Detect capabilities of configured search engine

DominikaK:
https://jira.ez.no/browse/EZP-28083

### DIFF
--- a/docs/api/public_php_api.md
+++ b/docs/api/public_php_api.md
@@ -704,7 +704,7 @@ Query object contains two properties you can set criteria on, `filter` and `quer
 
 !!! tip "Checking feature support per search engine"
 
-    To find out if a given search engine supports advanced full text capabilities, use the [`supports($capabilityFlag)`](https://github.com/ezsystems/ezpublish-kernel/blob/master/eZ/Publish/API/Repository/SearchService.php#L187-L197) method.
+    To find out if a given search engine supports advanced full text capabilities, use the [`$searchService->supports($capabilityFlag)`](https://github.com/ezsystems/ezpublish-kernel/blob/master/eZ/Publish/API/Repository/SearchService.php#L187-L197) method.
 
 Full code
 


docs/api/public_php_api.md
```diff
@@ -702,6 +702,10 @@ Query object contains two properties you can set criteria on, `filter` and `quer
 
 #### Performing a simple full text search
 
+!!! tip "Checking feature support per search engine"
+
+    To find out if a given search engine supports advanced full text capabilities, use the [`supports($capabilityFlag)`](https://github.com/ezsystems/ezpublish-kernel/blob/master/eZ/Publish/SPI/Search/Capable.php#L25) method.
```
andrerom: link is wrong, it should point to API not SPI for the intended audience here.

andrerom: https://github.com/ezsystems/ezpublish-kernel/blob/master/eZ/Publish/API/Repository/SearchService.php#L187-L197

SIDE: Linking to master is a bit risky, better to swap master for 6.12 once we have a branch for it.

docs/api/public_php_api.md
```diff
@@ -702,6 +702,10 @@ Query object contains two properties you can set criteria on, `filter` and `quer
 
 #### Performing a simple full text search
 
+!!! tip "Checking feature support per search engine"
+
+    To find out if a given search engine supports advanced full text capabilities, use the [`supports($capabilityFlag)`](https://github.com/ezsystems/ezpublish-kernel/blob/master/eZ/Publish/API/Repository/SearchService.php#L187-L197) method.
```
andrerom: example could be expanded to `$searchService->supports($capabilityFlag)`

andrerom: Side: @lserwatka you and team should probably decide on a API doc feature, links below point to Sami, others to doxygen.

PR #83, opened 2017-10-27
EZP-28042: Document concept of http cache tagging

DominikaK:
https://jira.ez.no/browse/EZP-28042

Mostly copied from https://github.com/ezsystems/ezplatform-http-cache/blob/master/docs/response_taggers.md

### DIFF
--- a/docs/guide/http_cache.md
+++ b/docs/guide/http_cache.md
@@ -763,6 +763,31 @@ This feature is based on [Context aware HTTP caching post](http://asm89.github.
 ezplatform-http-cache enables HTTP cache tagging.
 This allows you to add tags to cached content, which simplifies selective cache invalidation.
 
+#### Using tags
+
+Understanding tags is the key to making the most of `ezplatform-http-cache`.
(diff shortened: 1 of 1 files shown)


DominikaK:
ping @andrerom for a pro forma review, as this is really just moving existing doc.

andrerom:
yeah, we need more here, give me a bit adn I'll try coming up with something

docs/guide/http_cache.md
```diff
@@ -757,3 +757,72 @@ fos_http_cache:
 #### Credits
 
 This feature is based on [Context aware HTTP caching post](http://asm89.github.io/2012/09/26/context-aware-http-caching.html) by [asm89](https://github.com/asm89).
```
andrerom: Can be updated to something like this as our approach has changed a bit away from the concept there.

` This feature was originally inspired by on [Context aware HTTP caching blog post](http://asm89.github.io/2012/09/26/context-aware-http-caching.html) by [asm89](https://github.com/asm89) and has since evolved in FOSHttpCache and eZ to Varnish restart instead of curl. But the overall approach is the same, and its a good read for better understanding it.`

PR #92, opened 2017-11-08
 EZP-28185: Doc for "Database update SQL missing from Kernel 6.12" 

mere-teresa:
Documentation for the PR https://github.com/ezsystems/ezpublish-kernel/pull/2143

Should be merged in 1.12 and master

### DIFF
--- a/docs/releases/updating_ez_platform.md
+++ b/docs/releases/updating_ez_platform.md
@@ -133,6 +133,9 @@ Because from release 16.02 onwards eZ Platform is compatible only with PHP 5.5 a
     in `app/AppKernel.php`.
 
 
+!!! caution "Common errors"
+
+    If you experienced issues during the update, please check [Common errors](../getting_started/about_composer/#cloning-failed-using-an-ssh-key) section on the Composer about page.
 
 ## 4. Update database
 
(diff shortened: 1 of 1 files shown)


docs/releases/updating_ez_platform.md
```diff
@@ -212,29 +212,36 @@ index 49a17a9..80c4cd7 100644
...
-    ```
+    ​```
 
 **Increased password hash length**
 
 !!! caution "V1.12"
 
     v1.12.0 improves password security by introducing support for PHP's `PASSWORD_BCRYPT` and `PASSWORD_DEFAULT` hashing algorithms. By default `PASSWORD_DEFAULT` is used. This currently uses bcrypt, but this may change in the future as PHP adds support for new and stronger algorithms.
+    Apply the following database update script to change the schema and enable the storage of longer passwords:
+    Note that the script is available for PostGreSQL as well.
```
DominikaK: `PostgreSQL` with small "g"
Also, right now this sounds like the same script also works for Postgres, so maybe:
`Note that there is a script available for PostgreSQL as well.`

mere-teresa:
Conflicts and non-visible chars to fix.

andrerom:
Seems it now contains all commits from master not in 1.12.

To fix:
```
git rebase -i 1.12
# delete all lines not relevant for this PR, in vim it is by pressing `d` when marker is on those lines and end with `:wq`
git push origin EZP-28185-database-update-sql-missing -f

mere-teresa:
Okay, I made more mistakes with it, now.
Say, I am learning many things! (thank you @andrerom )

mere-teresa:
See #93

PR #99, opened 2017-11-24
EZP-28185: Fixing formatting, main titles were in Warning notes

mere-teresa:
Adding bold markers instead of a # that turns text into a Main Title (instead of SQL comments)


mnocon:
Shouldn't it go to the 1.12 branch too? (Same as the original one: https://github.com/ezsystems/developer-documentation/pull/93)

PR #102, opened 2017-11-29
Fix CachedValue typo and extending the input parser section

arfaram:
This PR aims to Fix the `CachedValue` typo error and also extend the `Input parser` section by adding a new action in the previous Controller to handle payload body of a HTTP Request.

### DIFF
--- a/docs/api/extending_the_rest_api.md
+++ b/docs/api/extending_the_rest_api.md
@@ -56,13 +56,22 @@ Next, you need to create the REST route. We need to define the route's [controll
 
 ``` yaml
 myRestBundle_hello_world:
-    pattern: /my_rest_bundle/hello/{name}
+    path: /my_rest_bundle/hello/{name}
     defaults:
         _controller: myRestBundle.controller.default:sayHello
     methods: [GET]
 ```
 
 Due to [![](https://jira.ez.no/images/icons/issuetypes/bug.png)EZP-23016](https://jira.ez.no/browse/EZP-23016?src=confmacro) - Custom REST API routes (v2) are not accessible from the legacy backend Closed , custom REST routes must be prefixed with `ezpublish_rest_`, or they won't be detected correctly.
+
+**My/Bundle/RestBundle/Resources/config/services.yml**
+
+``` yaml
+services:
+    myRestBundle.controller.default:
+        class: My\Bundle\RestBundle\Rest\Controller\Default
+        parent: ezpublish_rest.controller.base
+```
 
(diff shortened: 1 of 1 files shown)


docs/api/extending_the_rest_api.md
```diff
@@ -155,9 +164,28 @@ The easiest way to handle cache is to re-use the `CachedValue` Value Object. It
...
+    {
+        return new CachedValue(
+          new HelloValue( $name ),
+          ['locationId'=> 42]
+        );
+
+    }
+
+```
+Below you find the correspondent response header when using varnish as reverse proxy up 1.12:
```
DominikaK: `Below you can find the corresponding response header when using Varnish as reverse proxy:`

I suggest removing mention of 1.12, because we version the doc through branches. We can simply replace this with info proper for 1.12 when we merge this PR to 1.12 and later branches.

docs/api/extending_the_rest_api.md
```diff
@@ -223,20 +251,45 @@ class Greetings extends BaseParser
...
+      );
+
+      $name =  $createStruct->name ;
+
+      //...
+
+    }
 
+```
+The `inputDispatcher` is responsible to match the `Content-Type` sent in the header with the Greetings InputParser class.
```
DominikaK: `is responsible for matching`

docs/api/extending_the_rest_api.md
```diff
@@ -223,20 +251,45 @@ class Greetings extends BaseParser
...
-    myRestBundle.controller.default:
-        class: My\Bundle\RestBundle\Rest\Controller\Default
-        parent: ezpublish_rest.controller.base
+myRestBundle_hello_world_using_post:
+    path: /my_rest_bundle/hello/
+    defaults:
+        _controller: myRestBundle.controller.default:sayHelloUsingPost
+    methods: [POST]
 ```
+<div class="admonition note">
```
DominikaK: We have an easier to read way of placing note boxes directly in markdown. In this case it would be:

```
!!! note

    Post Requests are not able to access the repository without performing a user authentication. More information [REST API Authentication ](https://github.com/ezsystems/ezpublish-kernel/blob/master/doc/specifications/rest/REST-API-V2.rst#authentication).
```

(note the indent in the first line of the actual text)

arfaram:
@DominikaK  changes are applied. 
> Also, to make sure: this change applies to all version (including 1.7), but since 1.12 the quoted response will have to be modified, am I correct?

yes

PR #103, opened 2017-11-30
EZP-27752: Remove translation from all Versions

DominikaK:
https://jira.ez.no/browse/EZP-27752

Based on https://github.com/ezsystems/ezpublish-kernel/blob/master/doc/specifications/translations/removal.md#removing-the-specific-language-translation-from-all-the-content-object-versions

### DIFF
--- a/docs/api/public_php_api.md
+++ b/docs/api/public_php_api.md
@@ -1125,6 +1125,18 @@ try {
 }
 ```
 
+##### delete translation from all versions
+
+To delete a translation from all Content item versions, use the `deleteTranslation` method on `ContentService`.
+
+```
+public function deleteTranslation(ContentInfo $contentInfo, $languageCode);
+```
+
+This operation is permanent, so the user should be warned about that.
+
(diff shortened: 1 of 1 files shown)


DominikaK:
ping @alongosz

docs/api/public_php_api.md
```diff
@@ -1125,6 +1125,20 @@ try {
...
+
+To delete a translation from all Content item versions, use the `deleteTranslation` method on `ContentService`.
+
+```
+public function deleteTranslation(ContentInfo $contentInfo, $languageCode);
+```
+
+This operation is permanent, so the user should be warned about that.
+Since it is designed for maintenance tasks and might be a long-running operation,
+it should be used by a console command, not in the back office.
```
alongosz: Hmm, since I wrote this many months ago we've changed approach. So while to be safe I kept that in internal doc, we don't need to put it here as deleteTranslation is designed to have an UI (both in v1 and v2)

PR #106, opened 2017-11-30
EZEE-1781: Add Fastly integration

DominikaK:
https://jira.ez.no/browse/EZEE-1781

(v1.13)

Adapts internal docs from https://github.com/ezsystems/ezplatform-http-cache-fastly/pull/2/files#diff-04c6e90faac2675aa89e2176d2eec7d8


DominikaK:
ping @vidarl

docs/guide/http_cache.md
```diff
@@ -337,6 +337,79 @@ ezpublish:
                 purge_servers: [http://my.varnish.server:8081]
 ```
 
+#### Serving Varnish through Fastly
+
+[Fastly](https://www.fastly.com/) delivers Varnish as a service. See [Fastly documentation](https://docs.fastly.com/guides/basic-concepts/how-fastlys-cdn-service-works) to read how it works.
+
+#### Configuring Fastly
+
```
vidarl: It should be mentioned here or somewhere else that this is a EE feature and requires 1.13.0 or later
The fastly support is implemented in the EzSystemsPlatformFastlyCacheBundle bundle which will be shipped in platform-ee 1.13

DominikaK: Wrapped the text in a note box (it looks weird on GitHub, but will render as a nice box like e.g. here: https://doc.ezplatform.com/en/1.12/guide/extending_ez_platform_ui/#block-templates)

The version from which it is valid will be indicated by the doc version -> this will only go to the 1.13 and master branches, so it won't be visible in doc for earlier versions

PR #107, opened 2017-12-04
EZEE-1794: Publishing content in the future does not work

DominikaK:
https://jira.ez.no/browse/EZEE-1794

Adds a note about setting up date-based publisher during installation (in all install guides).

(1.7 and onwards)

### DIFF
--- a/docs/getting_started/install_manually.md
+++ b/docs/getting_started/install_manually.md
@@ -240,11 +240,9 @@ You will be able to see your page under <http://ez1.lh> (or the address you chos
 
 !!! enterprise
 
-    j. Enable Date-based Publisher
+    ###### Enable Date-based Publisher
 
-    To enable delayed publishing of Content using the Date-based publisher, run the following line:
-
-    `php app/console ezstudio:scheduled:publish --quiet --env=prod`
+    To enable delayed publishing of Content using the Date-based publisher, see [below](#enable-date-based-publisher_1)
 
 #### 10. Optional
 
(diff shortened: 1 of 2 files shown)


DominikaK:
ping @vidarl

docs/getting_started/install_manually.md
```diff
@@ -385,6 +393,14 @@ If Composer asks you for your token, you must log in to your GitHub account and
 
 Please note that a clean install of eZ Platform doesn’t include the DemoBundle anymore.
 
+!!! enterprise
+
+    j. Enable Date-based Publisher
+
+    To enable delayed publishing of Content using the Date-based publisher, run the following line:
+
+    `php app/console ezstudio:scheduled:publish --quiet --env=prod`
```
vidarl: It is not as strigt forward as this.

The command needs to be run on a regular basis. If you need minute resolution ( ie possibility to publish something at 10:32, and next article at 10:33) you need to run this once per minute. If a 5 minute resolution is fine, then run it every 5 minute..

The usual way running commands regulary on linux/unix is using a crontab (I don't know about Mac)
The second problem is to decide as which user the crontab should run ( user stuff is covered a bit later in this doc ).
It can run either as the web server user, or as the CLI user.

Assuming the web server user data is www-data, the commands to add this script in a cron is (run the commands as root):

```
    # for one-minute resolution
    echo '* * * * * cd [path-to-ezplatform]; php app/console ezstudio:scheduled:publish --quiet --env=prod' > ezp_cron.txt
    # for 5-minutes resolution, do instead
    echo '*/5 * * * * cd [path-to-ezplatform]; php app/console ezstudio:scheduled:publish --quiet --env=prod' > ezp_cron.txt

    # Next, append the new cron to user's crontab without destroying existing crons
    crontab -u www-data -l|cat - new_cron.txt | crontab -u www-data -

    # remove tmp file
    rm new_cron.txt
```

DominikaK: Thanks for the explanation, extended the description according to your comment

docs/getting_started/install_manually.md
```diff
@@ -385,6 +391,29 @@ If Composer asks you for your token, you must log in to your GitHub account and
 
 Please note that a clean install of eZ Platform doesn’t include the DemoBundle anymore.
 
+!!! enterprise
+
+    ###### Enable Date-based Publisher
+
+    To enable delayed publishing of Content using the Date-based publisher, you need to set up cron to run commands periodically.
```
vidarl: There is only one command that needs to run periodically ( but there multiple commands needed in order to set it up)
I would change it into:

(...) set up cron to run the command `app/console ezstudio:scheduled:publish` periodically.

docs/getting_started/install_manually.md
andrerom: Note: I think the intention from studio team is that it would be run via:
https://github.com/ezsystems/ezstudio-cron

At least the command here is registered for it:
https://github.com/ezsystems/date-based-publisher/search?utf8=%E2%9C%93&q=ezpublish.cron.job&type=

PR #112, opened 2017-12-07
Add tip for the user-context-hash request

kmadejski:
This PR contains small tip for the end-user about the destination of a request which has to be sent for the new user context hash.


kmadejski:
@andrerom take a glance, please.

docs/guide/http_cache.md
```diff
@@ -706,6 +706,8 @@ The easiest way to find the new hash is:
 
 `curl -I -H "Accept: application/vnd.fos.user-context-hash" http://<your-domain.com>/_fos_user_context_hash`
 
+Please, remember that you have to send this request to the backend, not to the Varnish.
```
andrerom: -the

So `not to Varnish.`

PR #118, opened 2017-12-12
EZP-28409: ezobjectrelationlist fieldtype with selection limit

DominikaK:
https://jira.ez.no/browse/EZP-28409

(2.0)

Also adds a note that Relation (single) FT is deprecated.


DominikaK:
ping @webhdx

PR #124, opened 2017-12-18
Extending the Dashboard

DominikaK:


docs/guide/extending_ez_platform_ui.md
```diff
@@ -165,6 +166,43 @@ $menu->reorderChildren(
 );
 ```
 
+## Dashboard
+
+To extend the Dashboard, make use of an event subscriber.
+
+In the following example, the `DasboardEventSubscriber.php` reverses the order of sections of the Dashboard
```
mere-teresa: Is `Sections` a domain-related thing? If so, could we first upper-case this word?

DominikaK: No, it's not, it's unrelated to "Sections" as in Admin Panel

PR #125, opened 2017-12-19
Add anonymous_user_id setting description. 

kmadejski:
This PR contains short description about `anonymous_user_id` setting. 

These changes cover approach valid for eZ Platform 1.7 and lower. 
It requires additional changes for 1.13 and up.

### DIFF
--- a/docs/guide/http_cache.md
+++ b/docs/guide/http_cache.md
@@ -609,6 +609,10 @@ sub ez_user_hash {
 }
 ```
 
+!!! tip "Upgrade your installation"
++
++    Consider upgrade to version 1.13 or newer because it doesn't require that much VCL changes to be able to benefit from `anonymous_user_id` setting while using Varnish.
+
 ##### Known limitations of the user hash generation
 
 If you are using URI-based SiteAccesses matching, the default SiteAccess on the domain needs to point to the same repository, because `/_fos_user_context_hash` is not SiteAccess-aware by default (see `ezpublish.default_router.non_siteaccess_aware_routes` parameter). Varnish does not have knowledge about SiteAccesses, so it won't be able to get user content hash if the default SiteAccess relies on URI.  


kmadejski:
@andrerom could you take a look, please?

andrerom:
Looks ok for anonymous_user_id, but didn't we have a need to also document how this can be used with Varnish in regards to fos hash lookups?

kmadejski:
Thanks. For now, I've added the only small tip on Varnish page, but of course, I can improve it if you think that possibility should be wider documented.

andrerom:
Well, wouldn't user need to adapt VCL to be able to use it? Doc changes here seems to imply it will work out of the box, was that the conclusion from recent issue on this?

kmadejski:
@andrerom yes, you are right. It should be covered also. I'm going to add this use-case.

kmadejski:
@andrerom I've added a section about using anonymous_user_id with Varnish, could you review once again, please?

docs/guide/best_practices.md
```diff
@@ -151,3 +151,18 @@ ezpublish:
 This setting **does not change Symfony behavior** regarding redirection after login. If set, it will only substitute the value set for `default_target_path`. It is therefore still possible to specify a custom target path using a dedicated form parameter.
 
 **Order of precedence is not modified.**
+
+### `anonymous_user_id`
+When a user that is not logged in is accessing an eZ Platform installation, the request will be handled with the permissions of the anonymous user. You can select any user on your system to represent the anonymous user.
```
DominikaK: `When a user who is...` and `select any user in your system`

docs/guide/best_practices.md
```diff
@@ -151,3 +151,18 @@ ezpublish:
...
+For instance, you can create also separated anonymous user group per siteaccess. Then, you can create the new user, assign him to the new group and make all permissions depend on siteaccess.
```
DominikaK: `For instance, you can also create a separate anonymous user group per siteaccess. Then, you can create the new user, assign them to the new group and make all permissions depend on siteaccess.`

docs/guide/best_practices.md
```diff
@@ -151,3 +151,18 @@ ezpublish:
...
+Take care to give the correct permissions to the anonymous user since these permissions are available for all users on your site. 
+
+You can set `anonymous_user_id` in the way like:
```
DominikaK: `in the way like` -> `the following way` or `like this`

docs/guide/http_cache.md
```diff
@@ -730,11 +737,167 @@ set req.http.X-User-Hash = "b1731d46b0e7a375a5b024e950fdb8d49dd25af85a5c7dd5116a
...
+                session:
+                    name: eZSESSID_eng
+            nor:
+                anonymous_user_id: 16
+                session:
+                    name: eZSESSID_nor
+        
+```
+
+You need to get the new X-User-Hash for every new anonymous user / anonymous users group. It is done in exactly same way as described above ([User hash generation](http_cache.md#user-hash-generation)), but you have to remember to use siteaccess URI element and additional header which tells eZ which URI should take into account. For instance for `eng` siteaccess:
```
DominikaK: `It is done in exactly the same way`
`remember to use the siteaccess URI element and an additional header`
`which URI it should take into account`

docs/guide/http_cache.md
```diff
@@ -730,11 +737,167 @@ set req.http.X-User-Hash = "b1731d46b0e7a375a5b024e950fdb8d49dd25af85a5c7dd5116a
...
+
+  `curl -I -H "Accept: application/vnd.fos.user-context-hash" -H "x-fos-original-url: /eng/" http://<your-domain.com>/eng/_fos_user_context_hash`
+  
+  and for `nor` siteaccess:
+  
+  `curl -I -H "Accept: application/vnd.fos.user-context-hash" -H "x-fos-original-url: /nor/" http://<your-domain.com>/nor/_fos_user_context_hash`
+
+Lets assume, that the new X-User-Hashes are:
```
DominikaK: `Let's assume that`

docs/guide/http_cache.md
```diff
@@ -730,11 +737,167 @@ set req.http.X-User-Hash = "b1731d46b0e7a375a5b024e950fdb8d49dd25af85a5c7dd5116a
...
+        // Request to nor siteaccess, so we should use other X-User-Hash for it
+        if (req.url ~ "^/nor") {
+            // Anonymous user w/o session => Use hardcoded anonymous hash to avoid backend lookup for hash
+            if (req.http.Cookie !~ "eZSESSID_nor" && !req.http.authorization) {
+                // Note: You should then update it every time anonymous user rights change.
+                set req.http.X-User-Hash = "a33ba7050ec3b848b266ef187623417b88b9df4b90483b7ef6582aa54ee72ee7";
+            }
+            // Pre-authenticate request to get shared cache, even when authenticated
+            else {
+                set req.http.x-fos-original-url    = req.url;
```
andrerom: Afaik we don't need to duplicate this section, the VCL session cookie pattern should be able to handle `_whatever` on session name.

Also lets move this syntax to be for ezplatform-http-cache VCL syntax, so it's the same across 1.7(.7) / 1.13 / 2.0 versions. In this xkey VCL flavor session cookie logic is already handled in `vcl_recv()` so less complicated logic in `ez_user_context_hash ()` to explain how to modify.

Optional: Move the VCL part of the doc to 1.13 to avoid having to explain which VCL type example is for, most users are moving to 1.13 / 2.0 anyway.

docs/guide/http_cache.md
```diff
@@ -337,6 +337,13 @@ ezpublish:
                 purge_servers: [http://my.varnish.server:8081]
 ```
 
+Moreover, we need to set/change `trusted_proxies` parameter in `app/config/config.yml` accordingly:
+
+```yaml
+framework:
+    trusted_proxies: [127.0.0.1,193.22.44.22]
+```
+
```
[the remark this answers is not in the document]
kmadejski: Yes, you are right. It is not necessary even in 2.8.x if _SYMFONY_TRUSTED_PROXIES_ is set properly. I'm going to remove this part.

docs/guide/http_cache.md
```diff
@@ -303,7 +303,7 @@ Somehow we need to tell php process that we are behind a Varnish proxy and not t
  
     # Configure IP of your Varnish server to be trusted proxy
     # Replace fake IP address below by your Varnish IP address
-    SetEnv SYMFONY_TRUSTED_PROXIES "193.22.44.22"
+    SetEnv SYMFONY_TRUSTED_PROXIES "127.0.0.1,193.22.44.22"
```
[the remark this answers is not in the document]
kmadejski: Yesterday we got ticket related to this setting. Mainly, it is required when HTTPS is in use and `{{render_esi(controller(...))}}` or `hinclude.js` is involved.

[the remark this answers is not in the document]
andrerom: Thats not an explanation ;) In order to explain why this is needed we also need to understand why it is like that. Is the example regarding cases where one of the proxies run on same machine, or is it caused by web server acting on the forwarding headers also so we need to tell symfony to trust it?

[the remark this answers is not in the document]
kmadejski: I've spent some time trying to find out the use cases when it is necessary, and honestly, I'm lost a bit. I can't find the correlation between this setting and some of the issues which our customers experienced in the past, even if setting this value solved problems. I will dig into this matter and will be back to this part of documentation as soon as possible with full and clear understanding.

kmadejski:
@DominikaK @MagdalenaZuba just for your information, I'm still waiting for the customer's feedback about this solution. We decided to include this in our documentation because of his use-case, so basically I need final confirmation that this approach works as expected. I'll finish this PR as soon as possible.

docs/guide/http_cache.md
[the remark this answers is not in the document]
andrerom: @kmadejski I guess you can drop this change now given Symfony will release your fix in next release ;)

kmadejski:
@andrerom I've simplified the example code and clean-up the whole PR according to the `TRUSTED_PROXIES` change in Symfony.
@DominikaK I've rebased this PR, so there are no conflicts. Please review again.

docs/guide/http_cache.md
```diff
@@ -501,6 +501,118 @@ set req.http.X-User-Hash = "b1731d46b0e7a375a5b024e950fdb8d49dd25af85a5c7dd5116a
 
 5\. Restart the Varnish server.
 
+##### New anonymous X-User-Hash based on `anonymous_user_id` setting
+
+Since you can configure anonymous user per SiteAccess, you need to change the default `.vcl` template to be able to benefit from this functionality. This example assumes that your SiteAccesses are matched using `URLElement` matcher. It works also with other SiteAccess matchers. As Varnish is not SiteAccess-aware itself, it is recommended to set also different session cookie names per SiteAccess.
```
andrerom: Maybe make it clear it's in cases where site access is matched by url we need to do this. In case of matching by host or port I think Varnish will keep the cache items separated.

andrerom: Also, why do we suggest using different session cookie name btw?

kmadejski: Updated the part about matchers.

> Also, why do we suggest using different session cookie name btw?
I kept it from the customer's solution, but TBH I can see now, that this is not necessary. Therefore, I've removed it also.

PR #129, opened 2017-12-21
Tab extensibility in V2

DominikaK:


DominikaK:
cc/ @nattfarinn

Nattfarinn:
cc/ @webhdx

docs/guide/extending_ez_platform_ui.md
```diff
@@ -165,6 +167,250 @@ $menu->reorderChildren(
 );
 ```
 
+## Tabs
+
+Many elements of the back-office interface, such as System Info or Location View, are built using tabs.
+
```
mere-teresa: Will we have a screenshot here, in the future?

docs/guide/extending_ez_platform_ui.md
```diff
@@ -165,6 +167,250 @@ $menu->reorderChildren(
...
+You can extend existing tab groups with new tabs, or create your own tab groups.
+
+### Adding a new tab group
+
+To create a custom tab group with additional logic you need to create it at the level of compiling the Symfony container, using a CompilerPass:
```
mere-teresa: This example needs some explanations IMHO.

mere-teresa: Could we add a link to the Sf documentation on CompilerPass?
https://symfony.com/doc/current/service_container/compiler_passes.html

docs/guide/extending_ez_platform_ui.md
```diff
@@ -165,6 +167,250 @@ $menu->reorderChildren(
...
+```
+
+A shorter way that you can use when no custom logic is required, is to add your own tab with the new group.
+If a tab's group is not defined, it will be created automatically.
+
+A new tab group can be rendered using the `ez_platform_tabs` Twig helper:
+
+``` jinja
+<div class="my-tabs">
+    {{ ez_platform_tabs('custom-tab-group') }}
```
mere-teresa: Does it have other parameters?
(see  https://github.com/ezsystems/ezplatform-admin-ui/blob/a7f9c83fe818775aeee90b960cc65f5ebe60e94c/src/bundle/Resources/views/dashboard/block/all.html.twig )

mere-teresa: Do we update the Twig Helper page list at the same time?

docs/guide/extending_ez_platform_ui.md
```diff
@@ -165,6 +167,250 @@ $menu->reorderChildren(
...
+</div>
+```
+
+This will render the group and all tabs assigned to it.
+
+### Adding a new tab
+
+Before you add a tab to a group you must create the tab's PHP class and define it as a Symfony service with the `ezplatform.tab` tag:
+
```
mere-teresa: Could we add a note: "See our doc about using Services"

docs/guide/extending_ez_platform_ui.md
```diff
@@ -165,6 +167,250 @@ $menu->reorderChildren(
...
+        return 100;
+    }
+}
+```
+
+The tabs will be displayed according to this value in ascending order.
+
+!!! tip
+
+    It is good practice to reserve some distance between these values, for example by multiplying them by 10.
```
mere-teresa: > by multiplying them by 10.

I suggest to use a formulation with the _step_ notion, it's from 10 to 10 more, like scale bars, not a multiplication. (I don't know if I am clear, here.)

docs/guide/extending_ez_platform_ui.md
DominikaK: We're talking about how the UI is built internally. A screenshot would be pointless, because it won't show anything.

docs/guide/extending_ez_platform_ui.md
DominikaK: Oh, and where do we have our doc on services?

docs/guide/extending_ez_platform_ui.md
DominikaK: Sorry, I totally don't understand what you mean here

docs/guide/extending_ez_platform_ui.md
mere-teresa: Tabs?

docs/guide/extending_ez_platform_ui.md
mere-teresa: We want to space the values from 10, it's a step.
item10, item 20, item 30, etc.
So later, we can put an item 21 after item 20 if needed.

docs/guide/extending_ez_platform_ui.md
DominikaK: Ok, makes sense, added a screenshot

PR #133, opened 2017-12-27
 Fixed OptionsBasedLatestContentQueryType example under "Guide">"Content rendering"

adamwojs:
> JIRA: -

## Description

Fixed multiple issues in [`OptionsBasedLatestContentQueryType`](https://doc.ezplatform.com/en/latest/guide/content_rendering/#the-optionsresolverbasedquerytype-abstract-class) example: 

1) `criterion` property was removed long time ago from `Query` class (https://github.com/ezsystems/ezpublish-kernel/commit/e2c171f31eabc93d26eccfc0a8a480ebf6393472) 
2) Value assigned to `limit` property should be `$parameters['limit']` instead of `$parameters`
3) `\Symfony\Component\OptionsResolver\OptionsResolver::{setAllowedValues,setAllowedTypes}` can be used only on defined options 
4) I guess that what the author had in mind is `$resolver->setAllowedTypes('limit', 'int')` instead of `$resolver->setAllowedValues('limit', 'int')` (because otherwise it doesn't make sense)
5) Missing `use` statements and code style in general

### DIFF
--- a/docs/guide/content_rendering.md
+++ b/docs/guide/content_rendering.md
@@ -1044,31 +1044,44 @@ The LatestContentQueryType from the [example above](#querytype-example-latest-co
 
 ``` php
 <?php
+
 namespace AppBundle\QueryType;
+
 use eZ\Publish\API\Repository\Values\Content\Query;
+use eZ\Publish\Core\QueryType\OptionsResolverBasedQueryType;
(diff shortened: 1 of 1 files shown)


mere-teresa:
Merged into 
1.7
 1.13
master

mere-teresa:
Merged into 2.0

PR #135, opened 2018-01-03
Fix cron setup for DBP

DominikaK:
Follow-up to comment https://jira.ez.no/browse/EZEE-1794?focusedCommentId=220711&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-220711

Fixes file name and applies the same command as in updating_ez_platform

Note to self: make sure to keep bin/console instead of app/console for 2.0


DominikaK:
ping @nattfarinn

Nattfarinn:
raf 750x1000 075 t 322e3f-696a94a5d4

PR #139, opened 2018-01-09
Fix folder permissions in installation

DominikaK:
Fixes folders to give permissions to in manual installation.

(2.0)

Note: Once this is approved, the same change will have to be applied to INSTALL.md in the meta repositories.

### DIFF
--- a/docs/getting_started/install_manually.md
+++ b/docs/getting_started/install_manually.md
@@ -536,8 +536,8 @@ Some systems don't support ACL at all. You will need to set your web server's us
 
 ``` bash
 sudo chown -R www-data:www-data var web/var
-sudo find {var,web/var} -type d | xargs sudo chmod -R 775
-sudo find {var,web/var} -type f | xargs sudo chmod -R 664
+sudo find web/var var -type d | xargs sudo chmod -R 775
+sudo find web/var var -type f | xargs sudo chmod -R 664
 ```
 
 ##### D. Using chmod on a *Linux/BSD/OS X* system where you can't change owner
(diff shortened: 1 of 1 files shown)


DominikaK:
ping @sunpietro

docs/getting_started/install_manually.md
```diff
@@ -503,7 +503,7 @@ First, change `www-data` to your web server user.
 ##### Clean the cache/ and logs/ directories
 
 ``` bash
-$ rm -rf var/cache/* var/logs/*
+rm -rf var/cache/* var/logs/*
```
sunpietro: `rm -rf var/cache/* var/logs/* var/sessions/*`

andrerom: if so, then do it on all examples (adding sessions)

andrerom:
@DominikaK where did this one come from? (the suggested changes here)

The INSTALL file was updated around the release, so unless it's wrong, this should perhaps be rather updated to reflect it's content: https://github.com/ezsystems/ezplatform/blob/master/INSTALL.md#user-content-install-2-folder-rights

DominikaK:
@andrerom, this PR's based on info from Piotr, who had trouble installing Enterprise using the existing doc and had to include more folders. @sunpietro, care for an explanation? :)

sunpietro:
In my case, the information about `var/sessions` folder was missing and session info could not be created. In the end, I was not able to run the eZ Platform EE v2 after cloning it from Github. I suggested this change to avoid any issues like that in the future.

andrerom:
Context clear then :)

Same issue was also fixed in the INSTALL file by giving access to full `var/` folder. This is taken from official [Symfony 3.4 doc](http://symfony.com/doc/3.4/setup/file_permissions.html), so I would suggest we align with that all over.

docs/getting_started/install_manually.md
```diff
@@ -535,9 +535,9 @@ Some systems don't support ACL at all. You will need to set your web server's us
 **Using chown on Linux/BSD/OS X systems that don't support ACL**
 
 ``` bash
-sudo chown -R www-data:www-data var/cache var/logs var/sessions web
-sudo find {var/{cache,logs,sessions},web} -type d | xargs sudo chmod -R 775
-sudo find {app/{cache,logs,sessions},web} -type f | xargs sudo chmod -R 664
+sudo chown -R www-data:www-data var web/var
+sudo find {var,web/var} -type d | xargs sudo chmod -R 775
```
andrerom: Now that options are flat and far less of them, here we can just as well drop the bash expansion syntax in favor of what is supported in plain shell as well:
`sudo find var web/var -type d | xargs sudo chmod -R 775`

Same on 4 more cases below. Do you agree @vidarl? If so we can update INSTALL.md as well in https://github.com/ezsystems/ezplatform/pull/249

andrerom: @DominikaK Update these to `web/var var` now: https://github.com/ezsystems/ezplatform/pull/249

PR #141, opened 2018-01-09
Corrects name of ElasticSearch index command

nmeirik:
The command in the current documentation does not seem to exist anymore. I found the alternative command here: https://fossies.org/linux/ezplatform/vendor/ezsystems/ezpublish-kernel/eZ/Bundle/EzPublishElasticsearchSearchEngineBundle/Command/ElasticsearchCreateIndexCommand.php


DominikaK:
Thank you for the contribution, @nmeirik!

PR #146, opened 2018-01-17
Improving the admin_group siteaccess description when adding multisites with multirepositories configuration

arfaram:


docs/guide/siteaccess.md
```diff
@@ -32,7 +32,7 @@ Typical uses of a SiteAccess are:
 The back-office UI of eZ Platform is housed in a predefined `admin` SiteAccess in `admin_group`.
 
 If you have a multisite setup with a separate back-office interface for each site,
-you need to create your own admin SiteAccesses for each site individually and add them to this group.
+you need to create your own admin SiteAccesses and add them to this group. In cases where the sites are on separate databases they'll need own repository config _(incl own storage and search connection)_, var dir, cache pool, as well as separate varnish/fastly cache in case that is in use for each site individually.
```
andrerom: `for each site individually` can be somewhat misunderstood, afaik host header is taken into account for http cache so it only needs to be separated per repo, or where you thinking of some issue here?

mere-teresa: I'm afraid that we will need an example of this configuration.

arfaram: actually var dir and varnish/fastly could be optional and configured in system.default for all siteaccesses (or groups). But in this context mandatory setting are repository config, DBs and cache pools. I could add an optionally settings note for storage and varinsh/fastly.

arfaram: Well, examples are available in the repository doc. what I can add here are some links to the related configuration.

docs/guide/siteaccess.md
```diff
@@ -32,7 +32,7 @@ Typical uses of a SiteAccess are:
 The back-office UI of eZ Platform is housed in a predefined `admin` SiteAccess in `admin_group`.
 
 If you have a multisite setup with a separate back-office interface for each site,
-you need to create your own admin SiteAccesses for each site individually and add them to this group.
+you need to create your own admin SiteAccesses and add them to this group. In cases where the sites are on separate databases they'll need own [repository](repository/#configuration-examples) _(incl own storage and search connection)_, and [cache pool](repository/#persistence-cache-configuration) config for each site individually. Optionally, you could define _var dir_ storage as well as separate varnish/fastly cache in case that is in use.
```
andrerom: @arfaram I might be missing something here, but do we have have a reference on the fact that seperation of var dir is optional? As in it might technically be possible to share*, but when using different databases I would think it would be clear recommendations to also separate binary files to make it easier to handle the different repositories.

* _NOTE: I can't at current time before ☕️  remember us having any requirement about making sure that would work, so I would perhaps be a bit cautious in regards to sharing var dir like that._


These are some of the reasons why I think this topic is better described on own page ;)

mere-teresa: Could you take a look at the comment, please @arfaram ?

arfaram: @andrerom well I think we should separate the var storage when using a different repository to avoid any confusion and mostly to easier manage binaries from the siteaccess management side. 
A scenario that can reflect and produce an error could be : 
- You upload a binary in siteaccess1 using repo1
The path of the e.g. image is:  _var/site/storage/images/8/6/2/0/268-1-eng-US/image.jpg_ and this path is saved in  _repo1.ezimage_  table. 
Now if it happens that you upload the same image to repo2 using siteaccess2 and the path generated let's admin that it will be the same, then you will override it with the same image in that case. But if the image is different and use the same name, then you will have another image in site1. 
The probability that this scenario could happen is very low, but still possible depend on the path generated and the binary name.

I agree with you that we should move this section to multidb set up. May be latter a link from here.

PR can closed

andrerom: 👍 But I don't think we need to close this, just adjust it a littel, @arfaram would you be ok with the following text change:
```
and add them to this group. In cases where the sites are on separate databases they'll need own [repository](repository/#configuration-examples) _(incl own storage and search connection)_, var dir, [cache pool](repository/#persistence-cache-configuration), and ideally also separate Varnish/Fastly config for each site individually.
```

PR #159, opened 2018-01-30
Update Bike tutorial to v2

DominikaK:
Updated the Platform beginner tutorial to v2.0.0.

Changes:
- updated screenshots
- change `app/console` to `bin/console`
- remade config examples to use `site_group` instead of `default`, as overriding `default` in v2 prohibits access to the Back Office (see https://jira.ez.no/browse/EZP-28506)
- added warning about user registration step only working from versions later than 2.0.1 (see https://jira.ez.no/browse/EZP-28711)
- modified one registration template to avoid an error
- added new necessary permissions

Note: if this is merged after 2.0.2 (at least rc) is out, we can change the warning note to refer to specific version, if it's confirmed to resolve the problem


MagdalenaZuba:
Step 2 - Create your content model
Under the hood, concepts and organization - link to 1.13 version
Admin Panel instead of Admin tab


Step 3 - Customizing the general layout
File structure - missing graphic in all versions

Step 1 - Display content of a Ride - should read 'Step 4'
Empty Note under 'Add a new parameter to your override rule' - all versions
Admin Panel instead of Admin tab

Step 6 - Enabling account registration
Picture not changed for Admin Roles
Admin Panel instead of Admin tab

DominikaK:
Changed Admin Panel -> Admin and broken image. The remaining fixes should be done to v1 as well, so I'll add them to https://github.com/ezsystems/developer-documentation/pull/185

PR #160, opened 2018-01-30
Update php package requirements in 1.13

DominikaK:
Removes the unnecessary `php-json` and `php-readline` from required php extensions

(1.13, 2.0 and master)

Note: two more changes are upcoming for 2.0


andrerom:
you can also drop:
- `php-memcached *(recommended, via [pecl](https://pecl.php.net/package/memcached))` part on redhat, which duplicates the info on row below for cluster extensions.
- `json` and `readline` from supported section below

PR #161, opened 2018-01-30
Update php package requirements in 2.0

DominikaK:
Removes the unnecessary `php-mcrypt` required php extensions.

Specifies the 7.1 version in `php7.1-xsl`.

(2.0 and master)


andrerom:
> Specifies the 7.1 version in php7.1-xsl.

I think we might be able to drop that one here, it seems like it is provided by php-xml package:
https://packages.debian.org/search?suite=all&section=all&arch=any&searchon=names&keywords=php7.1-xsl

DominikaK:
So I guess we should replace it with `php-xml`? At the moment xml is listed for Ubuntu but not for Debian.

andrerom:
yes indeed: https://packages.debian.org/stretch/php-xml

andrerom:
If you want as a follow up we could also start adding these links on the package names, will make it easier for us to keep track of updating them on new distro releases to verify they still have same name.

andrerom:
Could you btw for this PR change `17.10` to `17.10 "Artful"`?

PR #168, opened 2018-02-01
Fix SiteAccess Map\URI examples

DominikaK:
Fixes examples for Map\URI SiteAccess matching:
- key must equal value
- all SiteAccesses must be matched


docs/guide/siteaccess.md
```diff
@@ -323,8 +325,8 @@ ezpublish:
```
mnocon: ` **Example.** URI: `/something/my/content``
I think we should change `something` to `ezdemo_site` (otherwise this example does not make sense).

PR #169, opened 2018-02-01
Add permissions to log in to admin in v2

DominikaK:
Add to permissions use cases the new Policies that are necessary for logging in to the back office since v2.


DominikaK:
ping @webhdx

PR #172, opened 2018-02-06
Add warning about outdated Extending tutorial

DominikaK:


docs/tutorials/extending_platformui/extending_platformui_with_new_navigation.md
```diff
@@ -1,5 +1,10 @@
 # Extending PlatformUI with new navigation
 
+!!! caution
+
+    For the time being this tutorial only covers v1 of eZ Platform.
```
mere-teresa: I'll be more direct and let "for the time being" go.

DominikaK: Hm, how about "currently"? I'm wondering if simply saying "This tutorial only covers" may sound like we're not planning to update it in future.

mere-teresa: We don't want to expose our intentions, we want to expose the facts. So I am comfortable with saying "this tutorial only covers".

DominikaK: Makes sense, but should I remove the next line then?

PR #177, opened 2018-02-06
Explain connection between boosting, scoring and API use

andrerom:
review ping @pspanja


andrerom:
@DominikaK Forgot REST yesterday, added note on that, and it's valid on 1.7 and up _(was added in  a patch release of 1.7.x around a year ago)_

andrerom:
@DominikaK Seems like you got your review ;)

PR #178, opened 2018-02-07
TweetTutorial v2 docs update

ViniTou:
Jira ticket : https://jira.ez.no/browse/EZP-28690

### DIFF
--- a/docs/tutorials/field_type/8_add_a_validation.md
+++ b/docs/tutorials/field_type/8_add_a_validation.md
@@ -155,6 +155,7 @@ We would like to offer a way to the user to input a list of authors (upon which 
 This is a minimal example of our FormMapper:
 
 ``` php
+<?php
 // eZ/Publish/FieldType/Tweet/FormMapper.php
 
 namespace EzSystems\TweetFieldTypeBundle\eZ\Publish\FieldType\Tweet;
@@ -188,6 +189,7 @@ class FormMapper implements FieldDefinitionFormMapperInterface, FieldValueFormMa
 In our case, the TweetValueValidator expects authorList to be an array. On the other hand, our input field has TextType, so it will return a string. To solve this, we will transform data from an array to a comma-separated string and in the other way using a DataTransformer:
 
 ``` php
+<?php
 // Form/StringToArrayTransformer.php
 
 namespace EzSystems\TweetFieldTypeBundle\Form;
@@ -285,6 +287,7 @@ Also, we will register the new template in the configuration by editing the `ez_
 
 The last thing to do is to make sure that validation data is properly saved into and retrieved from the database. To achieve this, we will implement these two functions in LegacyConverter file:
 ``` php
+<?php
 // eZ/Publish/FieldType/Tweet/LegacyConverter.php
 
 public function toStorageFieldDefinition(FieldDefinition $fieldDef, StorageFieldDefinition $storageDef)
@@ -312,4 +315,4 @@ public function toFieldDefinition(StorageFieldDefinition $storageDef, FieldDefin
 You should now be able to configure a list of authors when editing a Content Type with the Tweet Field Type.
 You should also have your tweets validated in accordance to this list when you create or edit Content Item with this Field Type. 
 
-⬅ Previous: [Allow to add and edit Field in Back Office](7_allow_to_add_and_edit_field_in_back_office.md)
+⬅ Previous: [Allow adding and editing the Field in Back Office](7_allow_adding_and_editing_the_field_in_back_office.md)
--- a/docs/tutorials/field_type/3_implement_the_tweet_type_class.md
+++ b/docs/tutorials/field_type/3_implement_the_tweet_type_class.md
@@ -271,6 +271,7 @@ The interface has one method: `getEmbed( $statusUrl )` that, given a tweet's URL
 Your Field Type doesn't have a constructor yet. You will create one, with an instance of `Twitter\TwitterClientInterface` as the argument, and store it in a new protected property:
 
 ``` php
+<?php
 // eZ/Publish/FieldType/Tweet/Type.php:
 
 use EzSystems\TweetFieldTypeBundle\Twitter\TwitterClientInterface;
--- a/docs/tutorials/field_type/5_implement_the_legacy_storage_engine_converter.md
+++ b/docs/tutorials/field_type/5_implement_the_legacy_storage_engine_converter.md
@@ -34,9 +34,8 @@ The Converter will be placed alongside the Type and Value definitions (the kerne
 A Legacy Converter must implement the `eZ\Publish\Core\Persistence\Legacy\Content\FieldValue\Converter` interface:
 
 ``` php
-// TweetFieldTypeBundle/eZ/Publish/FieldType/Tweet/LegacyConverter.php
-
 <?php
+// TweetFieldTypeBundle/eZ/Publish/FieldType/Tweet/LegacyConverter.php
 
 namespace EzSystems\TweetFieldTypeBundle\eZ\Publish\FieldType\Tweet;
 
--- a/docs/tutorials/field_type/6_introduce_a_template.md
+++ b/docs/tutorials/field_type/6_introduce_a_template.md
@@ -73,6 +73,7 @@ However, this is far from ideal. You want this to be part of our bundle, so that
 To do so, you are going to make your Bundle's dependency injection extension (`DependencyInjection/EzSystemsTweetFieldTypeExtension.php`) implement `Symfony\Component\DependencyInjection\Extension\PrependExtensionInterface`. This interface will let you prepend bundle configuration:
 
 ``` php
+<?php
 // TweetFieldTypeBundle/DependencyInjection/EzSystemsTweetFieldTypeExtension.php
 
 use Symfony\Component\DependencyInjection\Extension\PrependExtensionInterface;
@@ -133,4 +134,4 @@ You should now be able to display a Content item with this Field Type from the f
 
 ⬅ Previous: [Implement the Legacy Storage Engine Converter](5_implement_the_legacy_storage_engine_converter.md)
 
-Next: [Allow to add and edit Field in Back Office](7_allow_to_add_and_edit_field_in_back_office.md) ➡
+Next: [Allow adding and editing the Field in Back Office](7_allow_adding_and_editing_the_field_in_back_office.md) ➡
--- a/docs/tutorials/field_type/7_allow_adding_and_editing_the_field_in_back_office.md
+++ b/docs/tutorials/field_type/7_allow_adding_and_editing_the_field_in_back_office.md
@@ -1,10 +1,11 @@
-# Allow to add and edit Field in Back Office
+# Allow adding and editing the Field in Back Office
 
-To able to add and edit Content Item with our Field Type using Back Office, we will implement a FormMapper with FieldValueFormMapperInterface. 
+To be able to add and edit Content Item with our Field Type using Back Office, we will implement a FormMapper with FieldValueFormMapperInterface. 
 DataTransformer is also needed in order to correctly transform our Value Object into single input field.
 
 # FormMapper
 ``` php
+<?php
 // eZ/Publish/FieldType/Tweet/FormMapper.php
 
 namespace EzSystems\TweetFieldTypeBundle\eZ\Publish\FieldType\Tweet;
@@ -63,6 +64,7 @@ Next thing is to register the FormMapper as a service, so the system would know 
 As mentioned earlier, we also need to tell our FormMapper how to transform ValueObject into a single string which contains URL to given tweet. We do this by creating an implementation of DataTransformerInterface.
 
 ``` php
+<?php
 // Form/TweetValueTransformer.php
 namespace EzSystems\TweetFieldTypeBundle\Form;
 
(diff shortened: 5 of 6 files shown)


ViniTou:
@DominikaK 
I can't assign you to this for some reasons.

docs/tutorials/field_type/8_add_a_validation.md
```diff
@@ -247,20 +254,22 @@ class FormMapper implements FieldDefinitionFormMapperInterface
 }
 ```
 
-Next thing is to register the FormMapper as a service, so the system would know to use it to automatically add the input field to the Content Type edit form. You can read more about services and [service container in the documentation](../../guide/service_container/). To register the FormMapper as a service, let's add the following lines to `fieldtypes.yml`:
-```yml
+Next thing is to tell our system that our FormMapper right now works also as FieldDefinitionFormMapper. In order to do that edit it definition in `fieldtypes.yml` with addition of extra tag:
```
DominikaK: In order to do that, add an extra tag in its definition in `fieldtypes.yml`:

docs/tutorials/field_type/7_add_content_and_edit_views.md
```diff
@@ -1,275 +1,116 @@
 # Add content and edit views
 
-To able to add and edit Content Item with our Field Type using PlatformUI, we need to create an edit view.
-The content view will also be needed to display correct information when viewing the Content Item in the "Content structure" tab in PlatformUI.
+To able to add and edit Content Item with our Field Type using PlatformUI, we will implement a FormMapper with FieldValueFormMapperInterface. 
+Also DataTransformer is needed in order to correctly transform our Value Object into single inputbox.
```
DominikaK: `DataTransformer is also needed` ... `single input box`, or even better `input field` because this is what you use later on.

docs/tutorials/field_type/7_add_content_and_edit_views.md
```diff
@@ -1,275 +1,116 @@
...
+        tags:
+            - {name: ez.fieldFormMapper.value, fieldType: eztweet}
+        arguments: ['@ezpublish.api.service.field_type']
 ```
 
-### Javascript view file
+# DataTransformer
 
-Next, we need to implement Javascript file that will be responsible for handling the template:
+As mentioned earlier we also somehow need to tell our FormMapper how to transform ValueObject into single string which contains URL to given tweet. We do this by creating implementation of DataTransformerInterface.
```
DominikaK: `As mentioned earlier, we also need to tell` ... `into a single string which contains the URL to a given tweet` ... `creating an implementation`

ViniTou:
@DominikaK 
Updated accordingly to comments :)

DominikaK:
Looks good, will be merged when the corresponding code is ready in the tutorial repo.

docs/tutorials/field_type/7_add_content_and_edit_views.md
```diff
@@ -1,275 +1,116 @@
...
-            res = {url: value.url};
+    public function __construct(FieldType $fieldType)
+    {
+        $this->fieldType = $fieldType;
+    }
 
-            return res;
+    public function transform($value)
+    {
+        if (!$value instanceof Value) {
```
mere-teresa: In a tutorial (aka to learn) could we use a strict comparison, please?

docs/tutorials/field_type/7_add_content_and_edit_views.md
```diff
@@ -1,275 +1,116 @@
...
+    }
+}
 ```
 
-After these steps you should see correct information when viewing Content Item in "Content structure" tab in PlatformUI.
+Next point will be using above transformer in our FormMapper. This is doable with addModelTransformer() method. Remember to pass eZ\Publish\API\Repository\FieldType object as constructor.
+
+``` php
+// eZ/Publish/FieldType/Tweet/FormMapper.php
+(...)
```
mere-teresa: If someone makes a copy-paste, what the (...) will become in PHP code?
Maybe a comment with `//beginning of the class` or put these parentheses in a comment could be better here.

docs/tutorials/field_type/8_add_a_validation.md
```diff
@@ -149,38 +149,45 @@ private function isAuthorApproved($author, $validatorConfiguration)
...
                 'required' => false,
                 'property_path' => 'validatorConfiguration[TweetValueValidator][authorList]',
+                'translation_domain' => 'eztweet_fieldtype',
                 'label' => 'field_definition.eztweet.authorList'
             ]);
     }
+    
+    public function mapFieldValueForm(FormInterface $fieldForm, FieldData $data)
+    {
+        (...)
```
mere-teresa: Same comment: could we comment this parenthesis, because copy-paste is frequent in tutos?

docs/tutorials/field_type/7_add_content_and_edit_views.md
DominikaK: Agreed, but this is something that crops up in other parts of the tutorial (e.g. https://doc.ezplatform.com/en/2.0/tutorials/field_type/5_implement_the_legacy_storage_engine_converter/#implementing-field-value-converters-tofieldvalue-and-tostoragevalue), so we'll have to fix that anyway in other pages when doing an audit.

docs/tutorials/field_type/7_add_content_and_edit_views.md
ViniTou: To be honest, I don't have a single clue what do you mean by strict comparison in case of 'instanceof' operator. Could you explain it to me?

docs/tutorials/field_type/7_add_content_and_edit_views.md
```diff
@@ -1,275 +1,116 @@
 # Add content and edit views
```
natanael89: This whole step in v2 is no longer about adding content and edit views, but it is about allowing add/edit form to be handled properly. Is that correct @ViniTou ? 
If so, then we should consider changing the name of the whole step.

ViniTou: Yup, that would be more specific indeed.

docs/tutorials/field_type/7_add_content_and_edit_views.md
mere-teresa: The example you gave is parenthesis (or brackets) but in a PHP comment, so it's copy-paste safe.

docs/tutorials/field_type/7_add_content_and_edit_views.md
mere-teresa: Strict comparison is `===` instead of `==`
Here it could have been
`if ( false === $value instanceof Value)`
Maybe it has no impact here.

ViniTou: To be honest, I am totally against something like this. instanceof is not a function but an operator. Also afair it is against all styling guides.

natanael89: I agree with @ViniTou, the current version is correct

docs/tutorials/field_type/7_add_content_and_edit_views.md
natanael89: @DominikaK @mere-teresa we need a decision about this. My suggestion would be to go for something like "Allow to add and edit Field in AdminUI", but perhaps better phrased :)

docs/tutorials/field_type/7_add_content_and_edit_views.md
```diff
@@ -1,275 +1,116 @@
...
 
-To able to add and edit Content Item with our Field Type using PlatformUI, we need to create an edit view.
-The content view will also be needed to display correct information when viewing the Content Item in the "Content structure" tab in PlatformUI.
+To able to add and edit Content Item with our Field Type using PlatformUI, we will implement a FormMapper with FieldValueFormMapperInterface.
```
natanael89: Are we still calling it PlatformUI? @DominikaK

natanael89:
This PR also needs code changes which emerged from the review of the related PR https://github.com/ezsystems/TweetFieldTypeBundle/pull/39.

docs/tutorials/field_type/7_add_content_and_edit_views.md
DominikaK: Good catch, it should be changed to `the Back Office`

docs/tutorials/field_type/7_add_content_and_edit_views.md
DominikaK: `Enable editing the Field in AdminUI` or `in the Back Office` would be my suggestion

ViniTou:
@DominikaK 
@mere-teresa 
@natanael89 
@MagdalenaZuba 

updated

docs/tutorials/field_type/7_allow_to_add_and_edit_field_in_back_office.md
```diff
@@ -0,0 +1,115 @@
+# Allow to add and edit Field in Back Office
```
DominikaK: `Allow adding and editing the Field in Back Office`
(ideally also in the file name)

docs/tutorials/field_type/7_allow_to_add_and_edit_field_in_back_office.md
```diff
@@ -0,0 +1,115 @@
...
+
+To able to add and edit Content Item with our Field Type using Back Office, we will implement a FormMapper with FieldValueFormMapperInterface.
```
DominikaK: `To be able to...`

docs/tutorials/field_type/8_add_a_validation.md
```diff
@@ -309,6 +310,6 @@ public function toFieldDefinition(StorageFieldDefinition $storageDef, FieldDefin
 `toStorageFieldDefinition()` converts a field definition to a legacy one using the `dataText1` field (which in this example means converting it to json), and `toFieldDefinition()` converts a stored legacy field definition to an API field definition (which means converting it back to an array according to validation schema).
 
 You should now be able to configure a list of authors when editing a Content Type with the Tweet Field Type.
-You should also have your tweets validated in accordance to this list when you create or edit Content Item with this Field Type.
+You should also have your tweets validated in accordance to this list when you create or edit Content Item with this Field Type. 
 
-⬅ Previous: [Add content and edit views](7_add_content_and_edit_views.md)
+⬅ Previous: [Allow to add and edit Field in Back Office](7_allow_to_add_and_edit_field_in_back_office.md)
```
DominikaK: Also change title and file name here

docs/tutorials/field_type/7_allow_to_add_and_edit_field_in_back_office.md
```diff
@@ -0,0 +1,115 @@
+# Allow to add and edit Field in Back Office
+
+To able to add and edit Content Item with our Field Type using Back Office, we will implement a FormMapper with FieldValueFormMapperInterface. 
+DataTransformer is also needed in order to correctly transform our Value Object into single input field.
+
+# FormMapper
+``` php
+// eZ/Publish/FieldType/Tweet/FormMapper.php
+
```
barbaragr: missing: `<?php`

docs/tutorials/field_type/7_allow_to_add_and_edit_field_in_back_office.md
```diff
@@ -0,0 +1,115 @@
...
+            - {name: ez.fieldFormMapper.value, fieldType: eztweet}
+        arguments: ['@ezpublish.api.service.field_type']
+```
+
+# DataTransformer
+
+As mentioned earlier, we also need to tell our FormMapper how to transform ValueObject into a single string which contains URL to given tweet. We do this by creating an implementation of DataTransformerInterface.
+
+``` php
+// Form/TweetValueTransformer.php
```
barbaragr: missing: `<?php`

docs/tutorials/field_type/8_add_a_validation.md
```diff
@@ -149,38 +149,45 @@ private function isAuthorApproved($author, $validatorConfiguration)
...
-## Implement FormMapper
+## Implement FieldDefinitionFormMapperInterface in FormMapper
 
-We would like to offer a way to the user to input a list of authors (upon which the data will be validated). To achieve this, we will implement a FormMapper that allows us to define the necessary input field.
+We would like to offer a way to the user to input a list of authors (upon which the data will be validated). To achieve this, we will add additional functionality in FormMapper which allows us to define the necessary input field.
 This is a minimal example of our FormMapper:
 
-```php
+``` php
 // eZ/Publish/FieldType/Tweet/FormMapper.php
```
barbaragr: missing `<?php`

docs/tutorials/field_type/8_add_a_validation.md
```diff
@@ -149,38 +149,45 @@ private function isAuthorApproved($author, $validatorConfiguration)
...
+        // (...)
+    }
 }
 ```
 
 In our case, the TweetValueValidator expects authorList to be an array. On the other hand, our input field has TextType, so it will return a string. To solve this, we will transform data from an array to a comma-separated string and in the other way using a DataTransformer:
 
-```php
+``` php
 // Form/StringToArrayTransformer.php
```
barbaragr: missing `<?php`

ViniTou:
@DominikaK 
updated!

PR #179, opened 2018-02-08
API'page title

mere-teresa:
The purpose is to have a page that is not a category in the site, and which title is different from the category in which the page is.

### DIFF
--- a/docs/api/api.md
+++ b/docs/api/api.md
@@ -2,7 +2,7 @@
 
 An Application Programming Interface (API) allows you to connect your code to eZ Platform.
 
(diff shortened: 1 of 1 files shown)


DominikaK:
Agree on the idea, but why "Presentation of APIs" in one place, and "API Introduction" in another?

Also, why `2.0` branch specifically?

mere-teresa:
On branch `2.0` because I thought we were applying on the last version and then on previous versions.

DominikaK:
No, if you look at any previously closed PR, it's almost always the earliest version -> latest version + master (or optionally starting with master and then backwards). That shouldn't make a difference in this case, just asking if you had any specific reason for it.

PR #186, opened 2018-02-13
Added caution: 127.0.0.1 required in Trusted Proxies when on Varnish 

kmadejski:
This PR contains important information about the `SYMFONY_TRUSTED_PROXIES` environment variable configuration which has to be enriched with the `127.0.0.1` address whenever application works behind Varnish and SSL offloader.

### DIFF
--- a/docs/guide/http_cache.md
+++ b/docs/guide/http_cache.md
@@ -318,6 +318,10 @@ fastcgi_param SYMFONY_HTTP_CACHE 0;
 fastcgi_param SYMFONY_TRUSTED_PROXIES "193.22.44.22";
 ```
 
+!!! caution "Trusted proxies when using SSL offloader / loadbalancer in combination with Varnish"
+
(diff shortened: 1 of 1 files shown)


docs/guide/http_cache.md
```diff
@@ -318,6 +318,10 @@ fastcgi_param SYMFONY_HTTP_CACHE 0;
 fastcgi_param SYMFONY_TRUSTED_PROXIES "193.22.44.22";
 ```
 
+!!! caution "Trusted proxies when using SSL offloader / loadbalancer"
+
+    If your installation works behind Varnish and SSL offloader (like haproxy) then you need   to add `127.0.0.1` address to the `SYMFONY_TRUSTED_PROXIES`. Otherwise, you might notice incorrect schema (`http` instead of `https`) in the URLs for the images or other binary files.
```
andrerom: `need   to ` => `need to `

suggestion: `other binary files when they are rendered inline Symfony _(as used by file based field templates)_, as opposed to via ESI.`

docs/guide/http_cache.md
```diff
@@ -318,6 +318,10 @@ fastcgi_param SYMFONY_HTTP_CACHE 0;
 fastcgi_param SYMFONY_TRUSTED_PROXIES "193.22.44.22";
 ```
 
+!!! caution "Trusted proxies when using SSL offloader / loadbalancer"
```
andrerom: `Tusted proxies when using SSL proxy / loadbalancer in combination with Varnish`

kmadejski:
Thanks for the review @andrerom. All suggestions were applied.

PR #187, opened 2018-02-13
Update "Custom controllers" part of the "Content Rendering" documentation

jacek-foremski:
JIRA issues: [EZP-28313](https://jira.ez.no/browse/EZP-28313) [EZP-28086](https://jira.ez.no/browse/EZP-28086)

This started as a simple fix for the code examples but includes some more drastic changes. The reasoning:
https://doc.ezplatform.com/en/1.7/guide/content_rendering/#custom-controllers currently describes three ways to extend the ViewController. The first one ("Enriching ViewController with a custom controller") is valid, I just updated the code example there. The only difference is that in my opinion, it shouldn't serve as a way to modify the Response. It's better (and I think the most common case) to return the View and not to deal with Response object at all.
As an alternative to this, I have described using Listeners to apply some custom logic, since it is also a valid way to do that.
The second way ("Using only your custom controller") is still possible but in my opinion less readable than the first way and/or Listeners. I left it in, but I have marked it as "not recommended".
The third way ("Overriding the built-in ViewController") in my opinion shouldn't be suggested at all because it comes down to overriding class from Core. I guess a lot of Core classes can be overridden in this way, but it should be considered last resort (and it shouldn't even be needed). Therefore I have removed it from this part of the documentation.

### DIFF
--- a/docs/guide/content_rendering.md
+++ b/docs/guide/content_rendering.md
@@ -1026,6 +1026,11 @@ class DefaultController extends Controller
         // If you wish, you can also easily access Location and Content objects
         // $location = $view->getLocation();
         // $content = $view->getContent();
+        
+        // Set custom header for the Response
+        $response = new Response();
+        $response->headers->add(['X-Hello' => 'World']);
+        $view->setResponse($response);
 
         return $view;
     }


docs/guide/content_rendering.md
```diff
@@ -1047,6 +1045,125 @@ These parameters can then be used in templates, for example:
...
+        $view->addParameters(['anotherCustomVariable' => "Another message!"]);
+
+        // If you wish, you can also easily access Location and Content objects
+        $location = $view->getLocation();
+        $content = $view->getContent();
+    }
+}
+```
+
+Similarly, you can use listener to modify the response generated after the view is rendered.
```
DominikaK: `you can use a listener`

docs/guide/content_rendering.md
```diff
@@ -1047,6 +1045,125 @@ These parameters can then be used in templates, for example:
...
+The configuration will look alike the previous one, for example:
```
DominikaK: `The configuration will look like the one earlier, for example:`

natanael89:
@DominikaK thanks for the review. Fixed according to your suggestions.

docs/guide/content_rendering.md
```diff
@@ -979,11 +979,11 @@ Typical use cases include access to:
 - Main Location and alternative Locations for the current Content item
 - etc.
 
-There are three ways in which you can apply a custom controller:
+There are three ways in which you can apply a custom logic:
 
-- Configure a custom controller alongside regular matcher rules to use **both** your custom controller and the `ViewController` (recommended).
-- [**Override**](#overriding-the-built-in-viewcontroller) the built-in `ViewController` with the custom controller in a specific situation.
-- **Replace** the `ViewController` with the custom controller for the whole bundle.
+- Configure a custom controller alongside regular matcher rules to use **both** your custom controller and the `ViewController`.
```
andrerom: Naming wise, it would be a `custom view controller` or? /cc @bdunogier

docs/guide/content_rendering.md
```diff
@@ -979,11 +979,11 @@ Typical use cases include access to:
...
+- [Add a listener to add your custom logic](#adding-a-listener) before or after the view is rendered.
+- [**Override**](#overriding-the-built-in-viewcontroller) the built-in `ViewController` with the custom controller in a specific situation (not recommended).
```
andrerom: This is "custom Symfony controller" right?

docs/guide/content_rendering.md
```diff
@@ -1073,31 +1190,31 @@ In this example, as the `ViewController` is not applied, the custom controller t
...
                 'content' => $content,
                 'foo' => 'Hey world!!!',
-                'osTypes' => array( 'osx', 'linux', 'windows' )
-            ) + $params
+                'osTypes' => ['osx', 'linux', 'windows']
+            ]
         );
+    
+        // Set caching for 1h and make the cache vary on user hash
+        $response->setSharedMaxAge(3600);
```
andrerom: What do we want to show here?

Casue by default CotnentView will be cached, by user hash, unless you opt out.

So we can show:
- addParameters(): To add parameters like the one above
- setResponse(): For setting custom repsponse headers
- setCacheEnabled(false): For disabling caching (but of course not recommended)

So example here could be written in a way that it still returns view, depending on what we want to show here.

docs/guide/content_rendering.md
```diff
@@ -1047,6 +1045,125 @@ These parameters can then be used in templates, for example:
...
+        $view->addParameters(['anotherCustomVariable' => "Another message!"]);
+
+        // If you wish, you can also easily access Location and Content objects
+        $location = $view->getLocation();
+        $content = $view->getContent();
+    }
+}
+```
+
+Similarly, you can use a listener to modify the response generated after the view is rendered.
```
andrerom: Why do we need this example? It overlaps with what we do in the product already and will easily lead to confusion.
ref: https://github.com/ezsystems/ezplatform-http-cache/tree/master/src/EventSubscriber

bdunogier:
I'd like to clarify what the intention was here, actually.

The main idea is that the view controller does *nothing*, and can be freely replaced without any dependency whatsoever.

The way to customize what a content view does is actually the second way (the least readable one ;)): use your own controller. You get a `ContentView` as the input, and return the `ContentView`. You may use `addParameters()`, and `setResponse` to customize the Response.
Under the hood, fetching the location, content, and configuring the template happens, transparently, _before_ the controller is invoked, by a controller listener. Rendering is also done transparently, by a View listener.

 That's why `eZ\Publish\Core\MVC\Symfony\Controller\Content\ViewController` has been *deprecated* since v1.0 of eZ Platform (kernel 6.0.0). The only exception is the `viewAction()` method, since it is the default one used by the system, but as you can see in the code, it is empty.

andrerom:
To add some context here, great work here @natanael89 on getting rid of the deprecated viewLocation and viewContent and rebase the examples to use contentView. However the other changes and decisions in this change needs some more work to take into account the new review notes here.

docs/guide/content_rendering.md
bdunogier: Yeah, it's a custom view controller.
But it doesn't run alongside with the `ViewController`. It replaces it. But since `ViewController::viewAction()` doesn't _do_ anything, it is perfectly regular.

docs/guide/content_rendering.md
bdunogier: I'm actually wondering if I'd mention it at all. Yes, it can be done, but why ? If you want to do something for every content view, the event listener can be used for that. I'd rather not mention it at all if it doesn't add any value.

Overriding services is something you can always do, it is not limited to this part of the system.

natanael89: Actually, this part isn't about doing something for every content view. The part you most probably have in mind is called "Overriding the built-in ViewController" and I removed it because I had the same thoughts as you.
This part describes applying the custom controller in a given match situation by indicating the controller, but no template. 
@bdunogier if I understand correctly, you have described this as "The way to customize what a content view does". However, as there is no template provided, you either need to provide one inside the controller or return a {{Response}} instead of the {{ContentView}}. I'd argue if it wouldn't be a better idea to use the first way by actually providing the template inside the configuration (if it is known). This way it's easy to just return {{ContentView}} object and let the Listener do the rendering (and this is the first way after my changes).
I'd probably also argue that if I'd like to modify the rendered {{Response}} object, then I'd also let the Listener do the rendering and then modify the {{Response}} in my own Listener, as I have described in the "Adding a listener" part (and this is the second way after my changes). I think that would be preferable to doing the rendering myself in the controller. 
However, I'm not so experienced with this topic, so I would like your opinion on the matter :)

docs/guide/content_rendering.md
natanael89: I have added this for the sake of completion - it describes how to modify the Response after the view is rendered. I have described in my previous comment why I think it is a good idea, at least as an option.
It doesn't have to concern cache (and you are right, I guess it shouldn't). I was just following the previous examples. I will change it to show how (for example) to add a custom Response header.

docs/guide/content_rendering.md
natanael89: I have changed the name accordingly.

docs/guide/content_rendering.md
natanael89: I changed this example to setting custom response header instead of dealing with cache. I wanted to show how to modify the `Response` object, so it was a good and safe choice.

docs/guide/content_rendering.md
natanael89: I changed it to also add the custom Response header instead of dealing with cache. Also simplified the example a bit.

natanael89:
Thanks @bdunogier for your clarification, it was very helpful. I have some doubts which I expressed in one of the comments, could you take a look?
Also thanks @andrerom for your suggestions. Hopefully, the examples will be more clear now. I have decided against splitting this PR into two (fixing examples/more advanced changes), because I figured that the updated examples wouldn't make much sense on their own, especially for the (previously) third way, which was removed entirely.

docs/guide/content_rendering.md
andrerom: AS I think @bdunogier mentioned you can drop this, in it's current form it is not clear when the user would need to do this.

So maybe rather change it, so it so the folioing in list at the top:
`- [Add a Symfony Response listener](#adding-a-listener) to add custom logic to all responses`

And section here can be cut down to just point to Symfony documentation for creating response listeners.

docs/guide/content_rendering.md
```diff
@@ -1073,31 +1174,30 @@ In this example, as the `ViewController` is not applied, the custom controller t
...
 namespace Acme\TestBundle\Controller;
 
-use Symfony\Component\HttpFoundation\Response;
 use eZ\Bundle\EzPublishCoreBundle\Controller;
+use eZ\Publish\Core\MVC\Symfony\View\ContentView;
 
 class DefaultController extends Controller
 {
-    public function viewFolderAction( $locationId, $layout = false, $params = array() )
+    public function viewFolderAction(ContentView $view)
```
andrerom: @bdunogier Is this example correct if we use ContentView?
If so some of the default logic is used so it might be a bit waste or?

If not, I guess we should expand intro here to make it clear that the difference here is that code takes care about rendering and the response itself.

docs/guide/content_rendering.md
```diff
@@ -1013,27 +1013,25 @@ With this configuration, the following controller will forward the request to th
...
-        // Forward the request to the original ViewController
-        // And get the response. Possibly alter it (here we change the smax-age for cache).
-        $response = $this->get( 'ez_content' )->viewLocation( $locationId, $viewType, $layout, $params );
-        $response->setSharedMaxAge( 600 );
+        $view->addParameters(['myCustomVariable' => "Hey, I'm a custom message!"]);
 
-        return $response;
+        // If you wish, you can also easily access Location and Content objects
+        $location = $view->getLocation();
+        $content = $view->getContent();
```
andrerom: comment out these two lines

Then add section below with something like so it is clear it is examples:
```
// Disable Content View cache
// Preferred alternative if possible: Vary by cookie, with default VCL it implies session cookie only
// $view->setCacheEnabled(false);

// Set as embed (normally this is done for you)
// $view->setIsEmbed(true);

// Set custom Response variables (for instance custom headers)
//$view->setResponse((new Response())->setExpires(new DateTime('@946684800')));
```

Please test last option tough, unsure how it works, of content and status code will be set correctly, and if some of our listeners/VCL might override things like expiry.

docs/guide/content_rendering.md
```diff
@@ -1119,80 +1219,6 @@ Here again custom parameters can be used in the template, e.g.:
...
- * @return \Symfony\Component\HttpFoundation\Response
- */
-public function viewContent( $contentId, $viewType, $layout = false, array $params = array() )
-```
-
-!!! caution "Caching"
-
-    When you use your own controller, **it is your responsibility to define cache rules**, like with every custom controller!
-
-    So don't forget to set cache rules and the appropriate `X-Location-Id` header in the returned `Response` object.
```
andrerom: This is still valid, for any kind of plain symfony controller.

docs/guide/content_rendering.md
```diff
@@ -979,11 +979,11 @@ Typical use cases include access to:
 - Main Location and alternative Locations for the current Content item
 - etc.
 
-There are three ways in which you can apply a custom controller:
+There are three ways in which you can apply a custom logic:
 
-- Configure a custom controller alongside regular matcher rules to use **both** your custom controller and the `ViewController` (recommended).
-- [**Override**](#overriding-the-built-in-viewcontroller) the built-in `ViewController` with the custom controller in a specific situation.
-- **Replace** the `ViewController` with the custom controller for the whole bundle.
+- Configure a custom view controller alongside regular matcher rules.
```
andrerom: + **(recommended)**

docs/guide/content_rendering.md
andrerom: Ignore this for now, would only make sense to mention for symfomy controllers, and as that is not done here anymore it can be removed.

Todo as followup:
- update `docs/guide/http_cache.md` for 1.13 and up with latests updates in: https://github.com/ezsystems/ezplatform-http-cache/blob/master/docs/using_tags.md#how-response-tagging-is-done

mere-teresa: Is this "todo as a followup" a story we can plan for a next sprint?

docs/guide/content_rendering.md
natanael89: Done.

docs/guide/content_rendering.md
natanael89: Done.

docs/guide/content_rendering.md
natanael89: I have commented out these two lines. I haven't added the section below yet because we need to make sure that it makes sense as an example here.

docs/guide/content_rendering.md
bdunogier: Yes, this example is correct with `ContentView`

docs/guide/content_rendering.md
andrerom: if doc team wants to take it, feel free, alternative was if @natanael89 wanted to update it.

docs/guide/content_rendering.md
```diff
@@ -1047,6 +1045,10 @@ These parameters can then be used in templates, for example:
 {% endblock %}
 ```
 
+### Adding a listener
+
+One way to add custom logic to all responses is to use your own listener. Please refer to the [Symfony documentation](https://symfony.com/doc/2.8/event_dispatcher/before_after_filters.html#after-filters-with-the-kernel-response-event) for the details on how to achieve this.
```
andrerom: NOTE: Remember to update version here on  merge to 2.0 and up

docs/guide/content_rendering.md
bdunogier: > I'd probably also argue that if I'd like to modify the rendered Response object, then I'd also let the Listener do the rendering and then modify the Response in my own Listener

I'd actually pre-configure the `Response` using `setResponse()`.

natanael89: @bdunogier You are right, I haven't thought about this option. I added an example of how to do this. Thanks.

PR #193, opened 2018-02-16
[Varnish] Add tip about X-User-Hash variation with Varnish

Plopix:
When using Varnish, Vary on X-User-Hash is added all the time when the header is present in the request:
https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/blob/1.3/EventListener/UserContextSubscriber.php#L180

```php
        if ($request->headers->has($this->hashHeader)) {
         ...
            if (!in_array($this->hashHeader, $vary)) {
                $vary[] = $this->hashHeader;
            }
```

which is the case all the time:
https://github.com/ezsystems/ezplatform/blob/master/doc/varnish/vcl/varnish4_xkey.vcl#L222

```vcl
sub vcl_deliver {
    // On receiving the hash response, copy the hash header to the original
    // request and restart.
    if (req.restarts == 0
        && resp.http.content-type ~ "application/vnd.fos.user-context-hash"
    ) {
        set req.http.x-user-hash = resp.http.x-user-hash;

        return (restart);
    }
```

> all requests pass here

My point is that the documentation is confusing as when we read it we expect to decide when the vary will be set.
**But with Varnish, we are not.**

In the eZ Kernel, (and with Varnish)
https://github.com/ezsystems/ezpublish-kernel/blob/master/eZ/Publish/Core/MVC/Symfony/Controller/Content/ViewController.php#L113
```php
         if ($request->headers->has('X-User-Hash')) {
                $response->setVary('X-User-Hash');
            }
```
That is useless in this case. FosHTTP Cache is doing it.

> That is not for that reason we should remove it, as we could have scenarios that need it.

Although it is fine to me, to Vary all the time, it is not a bug, I don't see an issue with that BUT that is confusing *not to know*.

I have entered into a use case where I wanted to share an ESI between everybody (every user hash) and that is not possible easily because of that automatic vary.

I struggled because to me, you had to `->setVary` yourself, but again, you don't ;)

In FOSHttpCacheBundle they have added an option `add_vary_on_hash` to allow that.

++

### DIFF
--- a/docs/guide/http_cache.md
+++ b/docs/guide/http_cache.md
@@ -178,6 +178,10 @@ $response = new Response();
 $response->setVary('X-User-Hash');
 ```
 
+!!! tip "Using Varnish"
+
+    When using Varnish and the proposed VCL, each Request will have the X-User-Hash header and then, every Response will vary on the X-User-Hash whether you explicitly set it up or not.
+
 ## Configuration
 
 ### Cache Purge
@@ -427,6 +431,10 @@ $response->setVary('X-User-Hash');
 
 This solution is implemented in Symfony reverse proxy (aka *HttpCache*) and is also accessible to dedicated reverse proxies like Varnish.
 
+!!! tip "Using Varnish"
+
+    When using Varnish and the proposed VCL, each Request will have the X-User-Hash header and then, every Response will vary on the X-User-Hash whether you explicitly set it up or not.
+
 !!! note
 
     Note that sharing ESIs across SiteAccesses is not possible by design (see [![](https://jira.ez.no/images/icons/issuetypes/bug.png)EZP-22535](https://jira.ez.no/browse/EZP-22535?src=confmacro) - Cached ESI can not be shared across pages/siteaccesses due to "pathinfo" property Closed for technical details)
@@ -478,6 +486,10 @@ $response->setVary('X-User-Hash');
 ```
 
 This solution is [implemented in Symfony reverse proxy (aka *HttpCache*)](http://foshttpcachebundle.readthedocs.org/en/latest/features/symfony-http-cache.html) and is also accessible to [dedicated reverse proxies like Varnish](http://foshttpcache.readthedocs.org/en/latest/varnish-configuration.html).
+
+!!! tip "Using Varnish"
+
+    When using Varnish and the proposed VCL, each Request will have the X-User-Hash header and then, every Response will vary on the X-User-Hash whether you explicitly set it up or not.
 
 #### Workflow
 


andrerom:
I would kind of assume our + fos symfony proxy enhancements also add this.  Also, we might threat this as bug/improvement, it needs to be possible to opt out somehow.

Plopix:
it does not hurt to warn people in the doc though :P

andrerom:
Should we cover all bugs in doc, or should we aim to solve the bugs? :)

MagdalenaZuba:
It depends on when we are planning to fix them.
The question is, what is an acceptable time frame of not including valid warning in the documentation. For me if the fix will be available within next 2 sprints we can leave this warning out but if it takes longer I'm for adding it to the documentation. After fix is available we can remove it.

andrerom:
If this is a bug it could be reported to support if this is customer instal.

> When using Varnish, Vary on X-User-Hash is added all the time when the header is present in the request:

Seems this is the case also when using Symfony Proxy:
https://github.com/FriendsOfSymfony/FOSHttpCache/blob/1.4/src/SymfonyCache/UserContextSubscriber.php#L108-L112

Did you test it using symfony proxy or is the varnish part here perhaps an assumption given it's hidden from you there?

Plopix:
hey @andrerom yes it might be the case with the Symfony Cache too. I don't how to debug that as the `request` must include the Header to trigger the automatic assignation on the `response`.

Again to me, it is not necessarily a bug, it is not bad to Vary on the X-User-Hash all the time, it manages almost 98% of the cases.

And a simple workaround is to remove the header in the Request depending on the URL in the VCL for instance. You can probably do that as well in a Listener `kernel.request` that is why I went for a PR on the doc ;)

Then it is really up to you, I did this PR to share this knowledge.

silversolutions:
hi @Plopix @andrerom 

we have added a Listener with a prio of -1000 in order to remove a Vary on User-Hash if we are using our own controller. Right now we are checking the X-Location-Id for it. 
It would be better to send the X-User-Hash only if a eZ view controller is involved.

andrerom:
@silversolutions Yeah I also think this should be threat as a bug / improvement, so please feel free to report issue on this.

Plopix:
I don't get why we don't warn the current users waiting for the resolution... or did I miss something?

andrerom:
Please report the issue so we can solve it, we can't fill the doc with warning of all issues in JIRA and on top of that on unreported issues. On top of that this was inaccurate afaik.

Plopix:
I don't know what is the more "inaccurate" ;-) the doc that says something which is not the case or a tip that warn people on a "known behavior".
But I will comply as usual, here is the JIRA issue: https://jira.ez.no/browse/EZP-28916

silversolutions:
Thanks @Plopix. I will create a ticket in the Enterprise contract were the issue happened as well and link it to the Jira issue. 
I think it has to be solved since in our case the vary on user hash would create in worst case a lot of duplicate http cache elements in varnish.

PR #198, opened 2018-02-22
Field Type Audit II

MagdalenaZuba:
Field Types reference - Audit
https://jira.ez.no/browse/EZP-28808 
https://doc.ezplatform.com/en/1.13/guide/field_type_reference

### DIFF
--- a/docs/guide/field_type_reference.md
+++ b/docs/guide/field_type_reference.md
@@ -1,6 +1,6 @@
 # Field Types reference
 
-A Field Type is the underlying building block of a content model. It consists of two entities: Field value and Field definition. Field value is determined by values entered into the Content Field. Field definition is provided by the Content Type, and holds any user defined rules used by Field Type to determine how a Field Value is validated, stored, retrieved, formatted and so on.
+A Field Type is the underlying building block of the content model. It consists of two entities: Field value and Field definition. Field value is determined by values entered into the Content Field. Field definition is provided by the Content Type, and holds any user defined rules used by Field Type to determine how a Field Value is validated, stored, retrieved, formatted and so on.
 
 eZ Platform comes with a collection of Field Types that can be used to build powerful and complex content structures. In addition, it is possible to extend the system by creating custom types for special needs. 
 


docs/guide/field_type_reference.md
```diff
@@ -219,7 +298,7 @@ array(
 );
 ```
 
-When you set an array directly on a Content field you don't need to provide all this information, the Field Type will assume it is a hash and in this case will accept a simplified structure described below under [To / From Hash format](#to-from-hash-format).
+When you set an array directly on a Content field you don't need to provide all this information, the Field Type will assume it is a hash and in this case will accept a simplified structure described below under [Hash format](#to-from-hash-format).
```
DominikaK: You also need to change the anchor link `(#hash-format)`

docs/guide/field_type_reference.md
```diff
@@ -393,11 +473,11 @@ $settings = array(
 
 ### Template rendering
 
-The template called by [the `ez_render_field()` Twig function](content_rendering.md/#ez_render_field) while rendering a Date Field has access to the following parameters:
+The template called by [the `ez_render_field()` Twig function](../guide/content_rendering.md/#ez_render_field) while rendering a Date Field has access to the following parameters:
```
DominikaK: This time th change is not necessary, because we are inside /guide/ already...
But the link is broken, I think, because it should be `(content_rendering.md#ez_render_field)`, without the backslash

docs/guide/field_type_reference.md
```diff
@@ -509,11 +589,11 @@ $settings = array(
 
 ### Template rendering
 
-The template called by [the `ez_render_field()` Twig function](content_rendering.md/#ez_render_field) while rendering a Date Field has access to the following parameters:
+The template called by the [`ez_render_field()` Twig function](../guide/content_rendering.md/#ez_render_field) while rendering a Date Field has access to the following parameters:
```
DominikaK: Like above

docs/guide/field_type_reference.md
```diff
@@ -746,14 +833,14 @@ When displayed using `ez_render_field`, an Image Field will output this type of
 <img src="var/ezdemo_site/storage/images/0/8/4/1/1480-1-eng-GB/image_medium.png" width="844" height="430" alt="Alternative text" />
 ```
 
-The template called by [the `ez_render_field()` Twig function](content_rendering.md/#ez_render_field) while rendering a Image Field accepts the following parameters:
+The template called by the [`ez_render_field()` Twig function](../guide/content_rendering.md/#ez_render_field) while rendering a Image Field accepts the following parameters:
```
DominikaK: Like above

docs/guide/field_type_reference.md
```diff
@@ -1282,7 +1372,7 @@ $MapLocationValue = new MapLocation\Value(
 
 ### Template rendering
 
-The template called by [the `ez_render_field()` Twig function](content_rendering.md/#ez_render_field) while rendering a Map Location Field accepts the following parameters:
+The template called by [the `ez_render_field()` Twig function](../guide/content_rendering.md/#ez_render_field) while rendering a Map Location Field accepts the following parameters:
```
DominikaK: The same with link

docs/guide/field_type_reference.md
```diff
@@ -1450,7 +1540,7 @@ This Field Type is used as fallback and for testing purposes.
 
 ### Description
 
-The Null Field Type serves as an aid when migrating from eZ Publish Platform and earlier versions. It is a dummy for legacy Field Types that are not implemented in eZ Platform.
+The Null Field Type serves as an aid when migrating from eZ Publish and earlier versions. It is a dummy for legacy Field Types that are not implemented in eZ Platform.
```
DominikaK: Actually "Publish Platform" can stay here, it's another name for 5.x (4.x was "Publish" and 5.x was "Publish Platform")

docs/guide/field_type_reference.md
```diff
@@ -1265,7 +1355,7 @@ The Value class of this Field Type contains the following properties:
 
 ###### Constructor
 
-The `MapLocation\Value` constructor will initialize a new Value object with the values provided. Two floats and a string are expected.
+The `MapLocation\Value` constructor will initialize a new Value object with three values provided: two floats and a string.
```
adamwojs: This sentence may suggest that constructor accepts three argument with is not true. More precise version IMHO: 

> The `MapLocation\Value` constructor will initialize a new Value object with values provided as hash. Accepted keys are `latitude` (`float`), `longitude` (`float`), `address` (`string`).

adamwojs:
Review ping @alongosz 😉

docs/guide/field_type_reference.md
```diff
@@ -116,6 +180,21 @@ The hash format mostly matches the value object. It has the following keys:
 - `uri`
 - `downloadCount`
 
+Example:
```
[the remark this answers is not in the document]
MagdalenaZuba: done

docs/guide/field_type_reference.md
```diff
@@ -70,12 +73,69 @@ It helps you get started by creating a skeleton for a Field Type, including temp
...
+   ]),
+]);
+```
+
+#### Hash format
+
+The hash format mostly matches the value object. It has the following key `authors`.
+
+
+Example
```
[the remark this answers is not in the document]
MagdalenaZuba: done

docs/guide/field_type_reference.md
```diff
@@ -1,12 +1,15 @@
 # Field Types reference
 
+A Field Type is the smallest entity of storage in eZ Platform. It determines how a specific type of information is validated, stored, retrieved, formatted and so on.
```
andrerom: I know this is form text that was here before, but this is not entirely correct for a few reasons:
- Field Type consists of two entities: Field Value and Field Definition, I guess you where referring to Field Value
- super nitpic: It's not the smallest, a hard coded integer column would be smaller


So I guess we should reword, what would we like to explain here as an abstract? That it's the underlying building block of content model/modeling or something like that? Text below kind of says that so guess we can make it even shorter.

docs/guide/field_type_reference.md
```diff
@@ -691,11 +769,20 @@ The `value` property of an Image Field will return an `\eZ\Publish\Core\FieldTy
...
 |`fileSize`|int|`37931`|The original image's size, in bytes|
 |`uri`|string|`var/ezdemo_site/storage/images/0/8/4/1/1480-1-eng-GB/image.png`|The original image's URI|
 |`imageId`|string|`240-1480`|A special image ID, used by REST|
+|`inputUri`|string|`var/storage/images/test/199-2-eng-GB/image.png`|Input image file URI.|
+|`width`|int|`null`|Original image width in pixels. For more details see Caution note below.|
+|`height`|int|`null`|Original image height in pixels. For more details see Caution note below.|
+
+!!! caution
+
+    Properties marked with an asterisk are currently unsupported. They are available but their value is always `null`.
```
andrerom: there is no asterisk

docs/guide/field_type_reference.md
```diff
@@ -1,12 +1,15 @@
 # Field Types reference
 
+A Field Type is the underlying building block of a content model. It consists of two entities: Field value and Field definition. Field value is determined by values entered into the Content Field. Field definition is provided by the Content Type, and holds any user defined rules used by Field Type to determine how a Field Value is validated, stored, retrieved, formatted and so on.
```
andrerom: Maybe wrong so more for @DominikaK to review on, but should it be "the content model" as we refer to our one and only amazing content model ;)

DominikaK: Agreed :)

PR #199, opened 2018-02-22
Field Type Audit Graphs

MagdalenaZuba:
New graphs for Create Content Sequence and Publish Content Sequence
Jira: https://jira.ez.no/browse/EZP-28808
Website: https://doc.ezplatform.com/en/2.0/api/field_type_api_and_best_practices/


MagdalenaZuba:
@alongosz please take a look

docs/api/field_type_api_and_best_practices.md
```diff
@@ -18,10 +18,22 @@ Below that, the Field Type must support the **Public API** implementation (aka B
...
+
+#### Create Content Sequence
 
 ![](img/create_content_sequence.png)
 
+#### Publish Content Sequence
+ 
+ !!! note "indexLocation()"
+ 
+    Below diagram shows implementation for **ElasticSearch**.
```
alongosz: Not exactly. Rather: "`indexLocation()` is implemented for **ElasticSearch** only."

(The Diagram is valid for all implementations as it reflects calls to methods. `indexLocation()` is just empty in case of Solr and SQL search.)

docs/api/field_type_api_and_best_practices.md
```diff
@@ -18,10 +18,22 @@ Below that, the Field Type must support the **Public API** implementation (aka B
...
+#### Create Content Sequence
 
 ![](img/create_content_sequence.png)
 
+#### Publish Content Sequence
+ 
+ !!! note "indexLocation()"
+ 
+    Below diagram shows implementation for **ElasticSearch**. 
+    For **Solr** `indexLocation()` is indexed during Content indexing.
```
alongosz: More precisely "For **Solr** Locations are indexed during Content indexing".

docs/api/field_type_api_and_best_practices.md
```diff
@@ -18,10 +18,22 @@ Below that, the Field Type must support the **Public API** implementation (aka B
...
 
 ![](img/create_content_sequence.png)
 
+#### Publish Content Sequence
+ 
+ !!! note "indexLocation()"
+ 
+    Below diagram shows implementation for **ElasticSearch**. 
+    For **Solr** `indexLocation()` is indexed during Content indexing. 
+    For **Legacy/SQL** indexing is not required as Location data is always available.
```
alongosz: Hmm, on second thought this wording I suggested might be misleading. 
Maybe let's change it to "... as Location data already exists in a database".

alongosz:
@andrerom if you could take a look at new diagrams it would be great. It seems to me that the diagram [we have now](https://doc.ezplatform.com/en/2.0/api/img/create_content_sequence.png) does not reflect any version (even 1.7). So maybe it was some sort of Legacy workflow where indexing occured on create and was dispatched from persistence layer?

Anyway we've worked with @MagdalenaZuba on changing them according to current implementation, taking into account indexing being dispatched via Signal Slot. We also separated create from publish as indexing occurs on the latter one. I hope it's fine, but please take a look :)

docs/api/field_type_api_and_best_practices.md
DominikaK: `The below diagram`

docs/api/field_type_api_and_best_practices.md
```diff
@@ -18,10 +18,22 @@ Below that, the Field Type must support the **Public API** implementation (aka B
...
 
+#### Publish Content Sequence
+ 
+ !!! note "indexLocation()"
+ 
+    Below diagram shows implementation for **ElasticSearch**. 
+    For **Solr** `indexLocation()` is indexed during Content indexing. 
+    For **Legacy/SQL** indexing is not required as Location data is always available. 
+
+![](img/publish_content_sequence.png)
```
DominikaK: Maybe add alt text? (Also in the image above, I can't comment there)

alongosz:
Good catch @adamwojs, there should be `getValidatorConfiguration()` as in the original one.

andrerom:
@alongosz  For the diagrams:
- create_content_sequence.png: Change looks correct, it reflects the change done back around v5.4.x to move search out of persistence when SolrBundle was introduced, so we can also remove "Content/SearchHandler" box now I think.  Should we explain field converters? It's specific to legacy storage engine and we might not need them in improved storage engine, but ..
- publish_content_sequence.png:  Looks ok, I was thinking the load operations might be overkill in this context, but can just as well explain it is how this works.

alongosz:
> create_content_sequence.png: (...) So we can also remove "Content/SearchHandler" box now I think.

That's right, good catch.

> Should we explain field converters? It's specific to legacy storage engine and we might not need them in improved storage engine, but ..

Hmm, I've just looked into what would needed to be done and it seems to me it's too much information. If we want to describe it, it should be a separate diagram.

PR #200, opened 2018-02-26
detailed info on BASIC Auth, Troubleshooting and diagrams for content exchange

johaho:


johaho:
@DominikaK please add resources in the img folder and reread the new content. Thx

PR #201, opened 2018-02-26
Add info about TTL for Landing Page blocks

DominikaK:


DominikaK:
ping @webhdx for review

docs/guide/http_cache.md
```diff
@@ -341,6 +341,27 @@ ezpublish:
...
+            block_type:
+                ttl: 600
+                views:
+                    (...)
+    ```
+
+    `block_type` should be replace with the actual block name i.e. `embed`, `collection`, `schedule` and so on.
+    In the example above `block_type` will be cached for 10 minutes.
+
+    By default blocks are not cached (TTL = 0).
```
andrerom: Can maybe add `.. (TTL = 0), for backwards compatibility reasons. This will change in v2.2`

PR #205, opened 2018-03-02
Fix environment option

DominikaK:
Fixes incorrect way of providing the environment in commands.

(all versions)


DominikaK:
ping @barbaragr @alongosz

PR #208, opened 2018-03-07
Update HTTP cache tagging

DominikaK:
Copies updates to HTTP cache tagging doc done in https://github.com/ezsystems/ezplatform-http-cache/commit/5aa25dc735dbbad0ccbba162e6ebfc1c6f7dd688 and https://github.com/ezsystems/ezplatform-http-cache/commit/1ed4ae87b5fbcd25f8a9f82b26f3d53a614161cc

### DIFF
--- a/docs/guide/http_cache.md
+++ b/docs/guide/http_cache.md
@@ -875,6 +875,9 @@ when someone requests them.
 Note that the system does not add this tag to responses itself, just purges if present.
 Response tagging using this tag is currently meant to be done inline in the template logic / views
 based on your decision.
+- `ez-all` - An internal tag used for being able to clear all cache.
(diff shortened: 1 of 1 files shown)


DominikaK:
ping @andrerom, follow-up to your https://github.com/ezsystems/developer-documentation/pull/187#discussion_r171854323

andrerom:
Note that the tag above will need some updates for next patch release. something like:
```
 - `relation-<content-id>` & `relation-location-<location-id>`- For use when updates affect their all reverse relations.
 Note that the system does not add this tags to responses itself, just purges them if present.
 Response tagging using this tag is currently meant to be done inline in the template logic / views
 based on your decision.
```

Change/addition is `relation-location-<location-id>` and changing tag*s* to plural.

PR #218, opened 2018-03-14
REST: get location that matches URL alias

MagdalenaZuba:
Added new feature based on https://github.com/ezsystems/ezpublish-kernel/pull/2255


MagdalenaZuba:
ping @ViniTou for review

ViniTou:
I think this doesn't belong at all in general REST usage.

PR #230, opened 2018-03-24
Add documentation on HTTPCACHE_PURGE_TYPE

Plopix:
HTTPCACHE_PURGE_TYPE is not mentioned anywhere.

Since the removal of this code in eZ Platform, purge_type is not set automatically.

```php
if ($purgeServer = getenv('HTTPCACHE_PURGE_SERVER')) {
    $container->setParameter('purge_type', 'http');
    // BC : In earlier versions, purge_type was set automatically if purge_server was set
    if ($purgeType === false) {
        $container->setParameter('purge_type', 'http');
    }
     $container->setParameter('purge_server', $purgeServer);
 }
```

Breaking the purge when using Varnish

### DIFF
--- a/docs/guide/http_cache.md
+++ b/docs/guide/http_cache.md
@@ -219,6 +219,11 @@ ezpublish:
                 purge_servers: ["http://varnish.server1", "http://varnish.server2", "http://varnish.server3"]
 ```
 
+!!! tip "Environment Variables"
+
+    eZ Platform uses environment variables by default, so you can define those values in the environment.
+    see [Update your Virtual Host](#update-your-virtual-host)
+
 For further information on setting up Varnish, see [Using Varnish](#using-varnish).
 
 #### Purging
@@ -300,7 +305,9 @@ Somehow we need to tell php process that we are behind a Varnish proxy and not t
  
     # Force front controller NOT to use built-in reverse proxy.
     SetEnv SYMFONY_HTTP_CACHE 0
- 
+    SetEnv HTTPCACHE_PURGE_TYPE http
+    SetEnv HTTPCACHE_PURGE_SERVER "http://varnish:80"
+
     # Configure IP of your Varnish server to be trusted proxy
     # Replace fake IP address below by your Varnish IP address
     SetEnv SYMFONY_TRUSTED_PROXIES "193.22.44.22"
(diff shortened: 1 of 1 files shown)


SylvainGuittard:
Thanks @Plopix

andrerom:
This 1.13 and up afaik.

Change was done in beginning of December here: https://github.com/ezsystems/ezplatform/pull/231

docs/guide/http_cache.md
```diff
@@ -300,7 +305,9 @@ Somehow we need to tell php process that we are behind a Varnish proxy and not t
  
     # Force front controller NOT to use built-in reverse proxy.
     SetEnv SYMFONY_HTTP_CACHE 0
- 
+    SetEnv SYMFONY_HTTP_CACHE http
```
andrerom: this will work, but as of 2.1 with only has http cache via ezplatform.http-cache, it also has alias as `varnish` on top of the `fastly` one provided by fastly bundle.

So for a master PR I guess it should rather set value as varnish.

ref: https://github.com/ezsystems/ezplatform-http-cache/blob/master/src/Resources/config/services.yml#L13-L18

docs/guide/http_cache.md
```diff
@@ -326,6 +336,8 @@ fastcgi_param SYMFONY_TRUSTED_PROXIES "193.22.44.22";
 
 Secondly we need to tell eZ Platform to change to use http based purge client *(specifically FosHttpCache Varnish purge client is used)*, and specify url Varnish can be reached on:
 
+Since eZ Platform 2.x, this following configuration is not required as eZ will read the Environment Variables set above.
```
andrerom: Can maybe mark this section as optional. More relevant as example if you need to set several purge servers for instance _(when you have for instance two varnish servers, typically one active and one passive)_.

Plopix:
I fixed a typo in my previous commit (in the Apache conf)
Let me know if you take over here, or if you expect me to do it ;)

docs/guide/http_cache.md
```diff
@@ -219,6 +219,11 @@ ezpublish:
                 purge_servers: ["http://varnish.server1", "http://varnish.server2", "http://varnish.server3"]
 ```
 
+!!! tip "Environment Variables"
+
+    eZ Platform is also using Environment Variables by default, then you can define those values in the Environment.
```
DominikaK: Maybe formulate it a bit more simply:
`eZ Platform uses environment variables by default, so you can define those values in the environment.`

docs/guide/http_cache.md
DominikaK: As this will only go to 2.0+ version of the doc, let's drop `Since eZ Platform 2.x,` and just do `The following configuration is not required as eZ Platform will read the environment variables set above.`

(we do not use "eZ" by itself in the doc to refer to eZ Platform)

docs/guide/http_cache.md
```diff
@@ -341,6 +353,10 @@ ezpublish:
                 purge_servers: [http://my.varnish.server:8081]
 ```
 
+!!! note "Multiple Purge Servers"
+
+    If you need to set multiple purge servers, then you need to configure it in the YAML file.
```
DominikaK: `you need to configure them`

PR #235, opened 2018-03-28
EZP-28985 1.7 Field Type Restructure

MagdalenaZuba:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | https://jira.ez.no/browse/EZP-28985
| Versions      | 1.7

Restructure the Field Type section of the developer doc to be more organized and more easily browseable.

### DIFF
--- a/docs/guide/service_container.md
+++ b/docs/guide/service_container.md
@@ -35,7 +35,7 @@ You will find all service tags exposed by Symfony in [its reference documentatio
 |`twig.loader`|Registers a template loader for twig|
 |`ezpublish.content_view_provider`|Registers a ContentViewProvider for template selection depending on content/Location being viewed|
 |`ezpublish.storageEngine`|Registers a storage engine in the Repository factory|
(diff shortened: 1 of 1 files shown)


mkdocs.yml
```diff
@@ -66,13 +66,22 @@ pages:
             - 'REST API reference': 'api/rest_api_reference.md'
         - Public PHP API:
             - 'Public PHP API': 'api/public_php_api.md'
-        - 'Field Type API and best practices': 'api/field_type_api_and_best_practices.md'
+        - Field Type:
+            - 'Reference': 'api/field_type_reference.md'
+            - 'Field Type API': 'api/field_type_api_and_best_practices.md'
```
DominikaK: Actually, maybe get rid of "best practices" here and in the page itself? I don't think we have any.

DominikaK: Also, maybe put this page in first place under "Field Type", even before "Reference"?

MagdalenaZuba: done

PR #241, opened 2018-04-04
EZP-28950: Update UTF8 to UTF8MB4

DominikaK:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | [EZP-28950](https://jira.ez.no/browse/EZP-28950)
| Versions      | 2.2, master

Changes utf8 to utf8mb4 where mentioned, plus adds upgrade note based on https://github.com/ezsystems/ezpublish-kernel/pull/2277/files#diff-079518e72cd553a190335b8aeaaf92de

Also a warning will need to be added in 2.2 relase notes

Note: merge after https://github.com/ezsystems/ezpublish-kernel/pull/2296

### DIFF
--- a/docs/releases/updating_ez_platform.md
+++ b/docs/releases/updating_ez_platform.md
@@ -256,6 +256,8 @@ These steps are only relevant for some releases:
     This is because the indexes have been shortened, so duplicates may occur.
     If that happens, you must remove the duplicates manually, and then repeat the statements that failed.
 
+    After successfully running those statements, change the character set and collation for each table, as described in https://github.com/ezsystems/ezpublish-kernel/blob/master/doc/upgrade/7.2.md.
+
     You should also change the character set that is specified in the application config:
 
     In `app/config/config.yml`, set the following:


docs/releases/updating_ez_platform.md
```diff
@@ -242,6 +242,33 @@ These steps are only relevant for some releases:
...
+    To apply this change, use the following database update script:
+
+    ``` bash
+    mysql -u <username> -p <password> <database_name> < vendor/ezsystems/ezpublish-kernel/data/update/mysql/dbupdate-7.1.0-to-7.2.0.sql
+    ```
+
+    Be aware that these upgrade statements may fail due to index collisions.
+    This is because the indexes have been shortened, so duplicates may occur.
+    If that happens, you must remove the duplicates manually, and then repeat the statements that failed.
+
```
glye: After this upgrade script is run, one must change the character set (and collation) of database tables. This is described in https://github.com/ezsystems/developer-documentation/pull/241/files
_After successfully running those statements, change the character set of each table..._

glye: (I don't much like all the steps that the db admins will have to do manually, but I think it is the best/safest way.)

docs/releases/updating_ez_platform.md
```diff
@@ -242,6 +242,35 @@ These steps are only relevant for some releases:
...
+    After successfully running those statements, change the character set and collation for each table.
```
glye: Sorry I was unclear. This needs a little more info. But I think we can get away with something like _...as described in `doc/upgrade/7.2.md`_, do you agree? Ref. https://github.com/ezsystems/ezpublish-kernel/pull/2296/files#diff-079518e72cd553a190335b8aeaaf92de

NB: That file won't contain that info until *after* https://github.com/ezsystems/ezpublish-kernel/pull/2296 has been merged. So don't merge your PR until that PR is merged. Sorry :)

DominikaK: You're right, done.

glye: I don't see it yet 🔭

DominikaK: Way to go, Dominika, forgetting to push :)

glye: They wrote a song about this back in the old days, before git was invented even: https://www.youtube.com/watch?v=Pmd3UiNfNkA

glye:
👍 https://github.com/ezsystems/ezpublish-kernel/pull/2296 is merged, so AFAIC this can be merged now. 👍

PR #255, opened 2018-04-16
Change PHP code blocks to short array syntax

DominikaK:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | n/a
| Versions      | all

Changes all PHP code blocks to use short array syntax (with the exception of those that copy code from the CookbookBundle in the API section).


DominikaK:
This was a special request by @alongosz (I bet you didn't think I'd actually do it ;) )

PR #265, opened 2018-04-25
Getting started - removing eZ Publish content

MagdalenaZuba:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | -
| Versions      | master

Getting started - removing outdated eZ Publish content

### DIFF
--- a/docs/getting_started/about_composer.md
+++ b/docs/getting_started/about_composer.md
@@ -169,7 +169,7 @@ Best practice for Bundles is described in Symfony documentation under [Best Pra
     
 !!! note "Package Dependencies"
 
-    Regardless of type you set, always remember to specify package dependencies in Composer requirements so the dependencies are clear. For example if you reply on `ezsystems/landing-page-fieldtype-bundle` make sure to add that with the major versions you support of it.
+    Regardless of the type you set, always remember to specify package dependencies in Composer requirements so the dependencies are clear. For example if you reply on `ezsystems/landing-page-fieldtype-bundle` make sure to add that with the major versions you support of it.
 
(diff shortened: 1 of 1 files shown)


docs/getting_started/about_composer.md
```diff
@@ -165,12 +165,11 @@ Best practice for Bundles is described in Symfony documentation under [Best Pra
...
-    -   `ezstudio-bundle` | Symfony bundles that uses eZ Platform Enterprise Edition features
-        -   Deprecated: Please use ezplatform-bundle and add dependencies on the ee packages you depend on instead.
     -   `symfony-bundle` | Standard symfony bundles as described in Symfony doc.
--   For eZ Publish (legacy) and eZ Publish Platform there where also:
-    -   `ezpublish-legacy-extension` | For standalone 4.x (legacy) extensions, to be used with [ezpublish-legacy-installer](https://github.com/ezsystems/ezpublish-legacy-installer)
-    -   `ezpublish-bundle` | For eZ Publish Platform 5.x bundles, may optionally be a "legacy bundle".
+    
+!!! note "Package Dependencies"
+
+    Remeber to specify package dependencies in Composer requirements so the dependencies are clear.
```
DominikaK: `RemeMber` :)

andrerom: > Remember to specify package dependencies in Composer requirements so the dependencies are clear, for instance if you reply on `ezsystems/landing-page-fieldtype-bundle` make sure to add that with the major versions you support of it.


Example is a bit dated as it will not be relevant for 2.2 page builder, but can maybe pick some ee package which is valid across all 2.x versions at least.

PR #281, opened 2018-05-17
User/Group identifier configuration

DominikaK:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | n/a
| Versions      | master

Covers new settings introduced in https://github.com/ezsystems/ezplatform-admin-ui/blob/master/src/bundle/DependencyInjection/Configuration/Parser/UserIdentifier.php#L16 and https://github.com/ezsystems/ezplatform-admin-ui/blob/master/src/bundle/DependencyInjection/Configuration/Parser/UserGroupIdentifier.php#L16


DominikaK:
ping @mikadamczyk for review

PR #284, opened 2018-05-21
Revamp (simplify?) the update procedure and thus install without --keep-vcs

Plopix:
This PR aims to _simplify?_ (from my point of view) the update procedure and hence to remove the `--keep-vcs` in the install procedure.

These points made me do this PR:

- `--keep-vcs` usage does not seem very common.
- As an eZ user I don't want to get the `git history` of the solution I am installing.
- Current update procedure deals with conflicts, remote, hard merging and sounds complex for a newcomers or for people who doesn't master `git`.
- But most importantly, the **procedure does not take into account** projects that would put eZ in a "subdirectory" (like with eZ Launchpad does, like we could do in a multi-app on Platform.sh or like we could do for any reasons)

Then I propose this PR and open the discussion using `rsync` and `git diff` to manage the update.

The idea behind it is that a developer who updates knows about the project specifics and can re-apply specifics really easily thanks to the `git diff` rather to mess with conflicts, remotes etc.

Can't wait to see eZ Platform install through Symfony Flex to simplify even more ;-)

EDIT: What would you think of a Smart script in `bin/updateplatform.bash` that would do the `rsync` with more intelligence.
Example:
- read the `.gitignore` to do not remove files present in it
- some other "procedure" we could automate.

This script would be in the new `cloned` eZ version and hence ready to run.

### DIFF
--- a/docs/releases/updating_ez_platform.md
+++ b/docs/releases/updating_ez_platform.md
@@ -13,7 +13,7 @@ If you intend to skip a version (for example, update directly from v1.3 to v1.5 
 
 ## Update procedure
 
-## 1. Check out a tagged version
+## 1. Prepare a branch to update your project
 
 **1.1.** From the project's root, create a new branch from the project's master, or from the branch you're updating on:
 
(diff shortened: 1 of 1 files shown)


DominikaK:
I think we need to summon @andrerom here as an expert

andrerom:
> I think we need to summon @andrerom here as an expert

I'll pong that to @bdunogier, he was the one setting up the original update/upgrade procedure which fits hand in hand with the usage of `--keep-vcs`

ilukac:
ping @emodric

emodric:
The idea itself is good, but I'm not sure I like the idea rsync wiping out the entire repo (apart from `src/`), which we need to bring back from git. What happens to files which are not tracked by git? Obvious ones being `app/config/parameters.yml`, which can grow quite big as the project matures, and `web/var/site/storage` or `.idea`, but really all other files and folders which might exist that are not part of the git repo.

Plopix:
@emodric you are right I was kind of lazy to just put that. I have updated.

`app/config/parameters.yml` should not be a problem, it should `recreated` with composer.

But to me, the current process is not strong/simple enough.

@andrerom @bdunogier @lserwatka What would you think of a Smart script in `bin/updateplatform.bash` that would do the `rsync` with more intelligence.
Example:
- read the `.gitignore` to do not remove files present in it
- some other "procedure" we could automate.

This script would be in the new `cloned` eZ version and hence ready to run.

emodric:
> `app/config/parameters.yml` should not be a problem, it should recreated with composer.

It is a problem when you have data in there that differs from `.dist` versions :)

Plopix:
@emodric that could be part of the proposed `bin/updateplatform.bash` 
What do you think about that?

Plopix:
and @emodric it is less and less true, the trend is to use env variables, which could report the problem to the `.env` you could say. But somehow we should be able to provide the vaule from the real `env` (opposed to DotEnv mechanism) (using vhost, docker, php-fpm or whatever)

bdunogier:
> It is a problem when you have data in there that differs from .dist versions :)

@emodric you mean the values, not the keys, don't you ? Are those stored anywhere (as in `.env`, as @Plopix said) in your case ?

emodric:
> you mean the values, not the keys, don't you ? Are those stored anywhere (as in .env, as @Plopix said) in your case ?

Yes, ofcourse, I mean the values :) That really depends on the project, various location IDs, API keys and so on, are not part of the repo and are only located in `app/config/parameters.yml`, and I can imagine the same situation with other developers too. `.env` has the same problem, since it is also not stored in the repo.

My point is, upgrade procedure should *never ever* delete anything on its own or assume that the developer will configure the `rsync` command correctly. It's a clear path to disaster, sooner or later. You don't know what the repo has or doesn't have, nor how it is structured.

docs/releases/updating_ez_platform.md
```diff
@@ -25,84 +25,66 @@ git checkout -b <branch_name>
...
-Conflicts in `composer.json` need to be fixed manually. If you're not familiar with the diff output, you may checkout the tag's version and inspect the changes. It should be readable for most:
+    The list of excludes is really depending on your project specifics and should be adapted. Ex: if you have eZ in a subfolder it would be
+    --exclude 'ezplatform/src/' --exclude 'ezplatform/web/var' etc. etc.
 
-**From your new update branch**
 
-``` bash
-git checkout --theirs composer.json && git diff HEAD composer.json
-```
+## 2. Check your branch before to update to the new version
```
bdunogier: "before updating"

docs/releases/updating_ez_platform.md
```diff
@@ -25,84 +25,66 @@ git checkout -b <branch_name>
...
-You should see what was changed, as compared to your own version, in the diff output. The update changes the requirements for all of the `ezsystems/` packages. Those changes should be left untouched. All of the other changes will be removals of what you added for your own project. Use `git checkout -p` to selectively cancel those changes:
-
-``` bash
-git checkout -p composer.json
-```
+At this stage your project will be updated with the last changes coming from the new eZ Platform version. This procedure
+will remove the custom configuration you have for your project specific.
 
-Answer `no` (do not discard) to the requirement changes of `ezsystems` dependencies. Answer `yes` (discard) to removals of your changes.
+No reason to worry, `git` will help, depending on you organization and your specifics you may want to check the working branch.
```
bdunogier: "git will help. Depending..."

docs/releases/updating_ez_platform.md
```diff
@@ -25,84 +25,66 @@ git checkout -b <branch_name>
...
-Once you are done, inspect the file, either using an editor or by running `git diff composer.json`. You may also test the file's sanity with `composer validate`, and test the dependencies by running `composer update --dry-run`. (will output what it would do to dependencies, without applying the changes.
+ ``` bash
+ git status
+ ```
 
-Once finished, run `git add composer.json` and commit`.`
+ !!! tip
 
-#### Fixing other conflicts (if any)
+     Or using PHPStorm you could do Cmd+K to see the `diff`
```
bdunogier: Cmd+K = Mac OS only. You could use the command's name, and provide the Mac OS specific shortcut.

docs/releases/updating_ez_platform.md
```diff
@@ -25,84 +25,66 @@ git checkout -b <branch_name>
...
 
-Depending on the local changes you have done, you may get other conflicts on configuration files, kernel, etc.
 
-There shouldn't be many, and you should be able to figure out which value is the right one for all of them:
+ It is time for you to check what the `rsync` did and (re)adapt accordingly. You should check and re-introduce your specifics:
```
bdunogier: This part is the one that doesn't convince me. But maybe I am biased and don't have much hands-on experience on doing updates. If all of you think that it is "simple enough" for most, then I'm fine with it. The question may also be if it is "simpler" than the current procedure.

lserwatka:
@Plopix we can certainly consider such a script and even more we can help with upgrade testing by involving our QA team.

alongosz:
Hi. This seems more relevant place to post this, so here's my POV:
- I agree we should drop `--keep-vcs` on install. If we wanted meta to be treated like a monolithic project, we shouldn't use composer at all. Which doesn't make sense, but just sayin'... :) `composer create-project` is meant for one time use to generate initial skeleton.
- **For some time now we upgrade meta repository when we want to include something for new installs only.** I'm afraid that if we officially introduce new script for syncing changes, the effect will be the same - people will update their files without thinking or even reading what they're changing.
- Because of the above I just think that release notes should include all updates which are required for meta-package, at least:
  - composer.json changes,
  - AppKernel bundles changes.
   
The upgrade procedure should involve only those steps mentioned above and executing 
```bash
composer update --with-dependencies
```

That said, @Plopix I think we could indeed use your idea with `rsync`, but rather as optional thing, maybe in a separate doc file linked in the upgrade instruction, but not obligatory. Basic upgrade should be much much simpler.

Of course that's just my POV. It requires approval by the important people: @lserwatka @bdunogier @SylvainGuittard @andrerom

andrerom:
TBH I don't have strong opinions on this. But I think Flex/3.x will solve this, so if this would take a lot of time to do and add complexity to updates I'm unsure if it's worth it.

DominikaK:
Closing this, because until the engineering and support teams reach a decision to change the recommended installation and update process, we can't merge it on our side.

(Anyway, the PR got outdated with the changes to the Getting started section that were introduced in the meantime.)

Plopix:
hey @DominikaK @SylvainGuittard I think we can reopen that one ;) that is still very true even more true with eZ 2->3

PR #288, opened 2018-05-23
[EZP-29234] As a developer, I want to read about customizing REST API response

konradoboza:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | [EZP-29234](https://jira.ez.no/browse/EZP-29234)
| Versions      | 1.7, 1.13, 2.1

This PR contains cookbook page which describes how to create custom REST API response. Example of such customization was introduced in https://github.com/ezsystems/PlatformUIBundle/pull/964.

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -122,10 +122,10 @@ pages:
         - 'Service container': 'guide/service_container.md'
         - 'Personalization': 'guide/personalization.md'
     - Cookbook:
+        - 'Accept header-based REST API response': 'cookbook/creating_custom_rest_api_response.md'
         - 'API search result pagination': 'cookbook/paginating_api_search_results.md'
         - 'AWS S3 clustering': 'cookbook/setting_up_amazon_aws_s3_clustering.md'
         - 'Binary attachments via REST API': 'cookbook/creating_content_with_binary_attachments_via_rest_api.md'
-        - 'Creating custom REST API response based on Accept header': 'cookbook/creating_custom_rest_api_response.md'
         - 'Cross-origin HTTP requests': 'cookbook/making_cross_origin_http_requests.md'
         - 'Displaying Content children': 'cookbook/displaying_children_of_a_content_item.md'
         - 'Executing long commands': 'cookbook/executing_long_running_console_commands.md'


docs/cookbook/creating_custom_rest_api_response.md
```diff
@@ -0,0 +1,268 @@
...
+    }
+
+    protected function visitVersionInfoAttributes(Visitor $visitor, Generator $generator, APIVersionInfo $versionInfo)
+    {
+        parent::visitVersionInfoAttributes($visitor, $generator, $versionInfo);
+
+        $this->visitTestValue($generator);
+    }
+
+    protected function visitTestValue(Generator $generator)
```
andrerom: `visitCommentValue `

docs/cookbook/creating_custom_rest_api_response.md
```diff
@@ -0,0 +1,268 @@
+#Creating custom REST API response based on Accept header
+
+There are several situations when customizing REST API response could be useful. One of the examples is extending it by adding comments count to
```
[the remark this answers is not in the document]
konradoboza: It's much better indeed, I was struggling a bit with the beginning, to be honest.

docs/cookbook/creating_custom_rest_api_response.md
```diff
@@ -0,0 +1,268 @@
...
+class ValueObjectVisitorDispatcher extends BaseValueObjectVisitorDispatcher
+{
+    private $parentDispatcher;
+
+    public function __construct(BaseValueObjectVisitorDispatcher $parentDispatcher)
+    {
+        $this->parentDispatcher = $parentDispatcher;
+    }
+
+    public function setOutputVisitor(Visitor $outputVisitor)
```
bdunogier: Not a big fan of these, but I guess that you couldn't find a simpler way to get this to work ?

docs/cookbook/creating_custom_rest_api_response.md
```diff
@@ -0,0 +1,268 @@
...
+
+namespace Acme\ExampleBundle\Rest\Generator;
+
+use eZ\Publish\Core\REST\Common\Output\Generator\Json as BaseJson;
+
+class Json extends BaseJson
+{
+    protected function generateMediaType($name, $type)
+    {
+        return "application/vnd.ez.example.{$name}+{$type}";
```
bdunogier: I don't understand that part. First, it feels odd that such an override is required. Then how does it work, as it uses the same code than the extended class: 

```
    public function getMediaType($name)
    {
        return $this->generateMediaType($name, 'json');
    }

    protected function generateMediaType($name, $type)
    {
        return "application/vnd.ez.api.{$name}+{$type}";
    }
```

Please help me :)

docs/cookbook/creating_custom_rest_api_response.md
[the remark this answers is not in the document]
konradoboza: >Customizing REST API response can be useful in many situations. Both for headless and more traditional setups, REST responses can be enriched in a clean way and limit client to server round trips.

>To do this we can take advantage of eZ Platform's [HATEOAS](https://en.wikipedia.org/wiki/HATEOAS) based REST API and extend it with custom content types for our own needs. Specifically in this cookbook we'll add comments count to eZ\Publish\API\Repository\Values\Content\VersionInfo responses.

Are you ok with this version?

docs/cookbook/creating_custom_rest_api_response.md
konradoboza: It allows us to identify request by our custom `Accept` header, e.g. 'application/vnd.ez.example.VersionList+json'. 

The default one doesn't allow it by using *api* part:

```
protected function generateMediaType($name, $type)
{
    return "application/vnd.ez.api.{$name}+{$type}";
}
```

In this example I was able to replace it with *example*.

andrerom: Maybe improve the naming a bit more, besides what Bertrand says maybe “my.api” as others should not use eZ as vendor name

docs/cookbook/creating_custom_rest_api_response.md
andrerom: Improvment can be done as a followup, given the doc here builds upon what has been released, it's imho good to go for now.

Maybe @adamwojs and/or @bdunogier can create a feature request to handle the improvment once there is priority for it.

andrerom:
Status update: This can be merged, and @konradoboza plans to do a PR on master to simplify a bit on the steps which would then result in doc for master once that is being accepted.

konradoboza:
@DominikaK do you want me to rebase it to get rid of the conflicts or do you prefer to do it differently?

DominikaK:
@konradoboza yes, please :)

PR #289, opened 2018-05-23
EZP-28868: Standard design

DominikaK:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | https://jira.ez.no/browse/EZP-28868
| Versions      | master

### DIFF
--- a/docs/guide/design_engine.md
+++ b/docs/guide/design_engine.md
@@ -29,8 +29,16 @@ By default two designs are included:
 `admin` covers templates for the Back Office. It contains the `admin` theme.
 
 `standard` covers the default content rendering templates. It contains the `standard` theme.
-This theme is automatically mapped to the directory in kernel containing the templates.
-The following templates are covered:
+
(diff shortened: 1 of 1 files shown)


docs/guide/design_engine.md
```diff
@@ -22,6 +22,42 @@ For assets, the system uses the Symfony Asset component with asset packages.
...
   A design is identified with a name. One design can be used per SiteAccess.
 
+### Default designs
+
+By default two designs are included:
+
+`admin` covers templates for the Back Office. It contains the `admin` theme.
+
+`standard` covers the default content rendering templates. It contains the `standard` theme.
+This theme is automatically mapped to the directory in kernel containing the templates.
```
alongosz: I'd actually reorder and rephrase something here. This line is true only if the setting mentioned below is set to `true`.
So here I'd move `Overriding templates` section and place this line somewhere there, with the list of the templates.

PR #290, opened 2018-05-23
Two clarifications for design engine

DominikaK:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | n/a
| Versions      | 1.13, 2.1, master

Adds two warnings about the design engine.

### DIFF
--- a/docs/guide/design_engine.md
+++ b/docs/guide/design_engine.md
@@ -19,8 +19,8 @@ For assets, the system uses the Symfony Asset component with asset packages.
 
 !!! caution
 
-    After creating a new folder in `Resources/views/themes/` you need to clear Symfony cache `bin/console cache:clear`,
-    even if you are working in the dev environment.
+    After creating a new folder either in `app/Resources/views/themes/` or your bundle's `Resources/views/themes/` directory
+    you need to clear Symfony cache (`bin/console cache:clear`), even if you are working in the dev environment.
 
 - **Design**: A collection of themes.
 


docs/guide/design_engine.md
```diff
@@ -17,11 +17,20 @@ For assets, the system uses the Symfony Asset component with asset packages.
...
+    even if you are working in the dev environment.
+
 - **Design**: A collection of themes.
 
   The order of themes within a design is important as it defines the fallback order.
   A design is identified with a name. One design can be used per SiteAccess.
 
+!!! caution
+
+    The templates contained in `app/Resources/views` will always override templates from a design theme.
```
alongosz: We've dropped that as it caused a headache and released eZ Design Engine 2.0.
Note: I'm considering making this directory fallback instead of global/override, but we'll see...

docs/guide/design_engine.md
```diff
@@ -17,6 +17,11 @@ For assets, the system uses the Symfony Asset component with asset packages.
   Typically a directory containing templates. For example, templates located under `app/Resources/views/themes/my_theme`
   or `src/AppBundle/Resources/views/themes/my_theme` are part of `my_theme` theme.
 
+!!! caution
+
+    After creating a new folder in `Resources/views/themes/` you need to clear Symfony cache `bin/console cache:clear`,
```
alongosz: On second thought, this might be not clear to everyone, so maybe something like:
```
After creating a new folder either in `app/Resources/views/themes/` or your bundle `Resources/views/themes/` directory you need to clear Symfony cache (`bin/console cache:clear`),
```

PR #313, opened 2018-06-25
EZP-25487: Warning about clearing persistence cache when changing image variations

DominikaK:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | https://jira.ez.no/browse/EZP-25487
| Versions      | all

### DIFF
--- a/docs/guide/images.md
+++ b/docs/guide/images.md
@@ -48,11 +48,7 @@ The following parameters are set for each variation:
 
 !!! caution
 
-    If you change image variation properties manually, you need to clear persistence cache:
-
-    `php bin/console cache:pool:clear <pool_name>`
-
-    where `<pool_name>` is `cache.app` by default, and `cache.redis` when using Redis.
+    If you change image variation properties manually, you need to clear persistence cache.
 
 ### Built-in image variations
 


docs/guide/images.md
```diff
@@ -46,6 +46,14 @@ The following parameters are set for each variation:
 - `reference`: Name of a reference variation to base the variation on. If set to `null` (or `~`, which means `null` in YAML), the variation will take the original image for reference. It can be any available variation configured in the `ezpublish` namespace, or a `filter_set` defined in the `liip_imagine` namespace.
 - `filters`: Array of filter definitions (hashes containing `name` and `params` keys). See possible values [below](#available-filters).
 
+!!! caution
+
+    If you change image variation properties manually, you need to clear persistence cache:
+
+    `php bin/console cache:pool:clear <pool_name>`
```
alongosz: Unfortunately this applies to 2.x stack only, since this command is available in Symfony 3.4.
For 1.x we can state something like if you're using external persistence cache handler (not sure naming is consistent with what we had), like Redis, then you need to manually clear that cache (in case of Redis it is native Redis command flushall).

For 2.x this is fine.

DominikaK: True, the specific command will be added when merging to 2.x

PR #320, opened 2018-06-29
[2.2 release notes] Mention improvements to API

andrerom:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | 
| Versions      | 2.2

Adds mention of these two changes:
- https://github.com/ezsystems/ezpublish-kernel/pull/2328
- https://github.com/ezsystems/ezpublish-kernel/pull/2064

_Note: language loading is intended to work on ContentView, but QA uncovered it does not in all cases yet. I mention the intended behavior here, even if by now fix for it will not come before 2.2.1 or 2.3 depending on impact._

### DIFF
--- a/docs/api/public_php_api_browsing.md
+++ b/docs/api/public_php_api_browsing.md
@@ -101,7 +101,8 @@ If you want to take SiteAccess languages into account:
 -   Provide prioritized languages on `load()` this will be taken into account by the returned Content object when retrieving translated properties like fields, for example:
     `$contentService->loadContent( 66, $configResolver->getParameter('languages'));`
     -   `ConfigResolver` is a service, so it can be obtained from Symfony Container or injected directly: `@ezpublish.config.resolver`.
-    -   See [below](#siteaccess-aware-repository-optional) for information about SiteAccess-aware repository.
+
+See [below](#siteaccess-aware-repository-optional) for information about SiteAccess-aware repository.
 
 Otherwise if you want to use an altogether different language, you can specify a language code in the `getField()` call:
 


DominikaK:
@andrerom It would be best to keep the release notes as concise as possible with just a mention of the change and link to extended explanation in a different page.
What would you say about moving the bulk of the explanation to https://doc.ezplatform.com/en/2.2/guide/internationalization/ or https://doc.ezplatform.com/en/2.2/api/public_php_api_managing_content/#managing-content ?

andrerom:
Yes, that would be better indeed, also there are other improvements to API also, both notification and bookmarks have been made part of API which is equally huge news. So this can be adapted to list all features added to API.

andrerom:
@DominikaK So that would mean we will split up the example, as the one I have here shows both changes in effect.

DominikaK:
@andrerom Are you good with the changes done here? (more or less according to what we discussed).

docs/api/public_php_api_managing_content.md
```diff
@@ -65,6 +65,10 @@ In order to set a Location for your object, you must instantiate a [`LocationCre
 $locationCreateStruct = $locationService->newLocationCreateStruct( 2 );
 ```
 
+#### Location object with access to Content
+    
+You can directly get it by using `$location->getContent()`, it can also be very useful in a Twig via `location.content`. This functionality additionally introduces possibility to specify prioritised languages when loading a Location. Content will be loaded on-demand across result set you are loading (e.g. search and other places you can load several Locations).
```
andrerom: So, I checked the doc and actually wanted to ask you if this part is better placed in `Browsing, finding, viewing` as this page is abut creating/updating data atm.

docs/guide/internationalization.md
```diff
@@ -268,6 +268,34 @@ You can also assign a Default content availability flag to Content Types (availa
 
 Note that if a language is not provided in the list of prioritized languages and it is not the Content item's first language, the URL alias for this content in this language will not be generated.
 
+### SiteAccess-aware Repository - optional
```
andrerom: This can probably also go into above mentioned page, specifically part of, or next to:
https://doc.ezplatform.com/en/2.2/api/public_php_api_browsing/#loading-content-in-different-languages

For instance this bullet point of that section is referring to the SiteAccessAware repository change we do know:
`- Note: As of v2.0 this is planned to be done for you when you don't specify languages.`

So that note can be changes to referring to: as of 2.2, ContentView and Parameters resolvers use this by default, as of 2.3 auto-wiring of services will most likely have this by default also it seems:  https://github.com/ezsystems/ezpublish-kernel/pull/2380

And then the note can say read more below in SiteAccess Aware section, which for the most part can be what you propose here in internationalization.md.

But while on this block in Browsing, finding, viewing, the example already there:
`$content->getFieldValue( $fieldDefinition->identifier, 'fre-FR' )`

becomes the following with SiteAccess aware repo:
`$content->getFieldValue( $fieldDefinition->identifier )`

--

**For the text proposed here, maybe something like:**


The SiteAccess-aware repository is an instance of the eZ Platform Repository API which injects prioritised languages for you when you load data *(Content, Location, Content type, etc.)*, and don't specify languages to load yourself in API arguments.

Currently it is:
- Available as a private service `ezpublish.siteaccessaware.repository`, including for the other Repository services.
- Used out of the box on parameter converters for Content & Location
- Used out of the box on ContentView
- Planned to be used for autowired API services as of 2.3, and as of 3.0 will be returned on `ezpublish.api.*` service id's

Example of code before using SiteAccess aware repository:
```php
$content = $this->contentService->loadContent(
    42,
    $this->configResolver->getParameter('languages')
);

$name = $content->getVersionInfo()->getName();
$value = $content->getFieldValue('body')
```

Becomes:
```php
$content = $this->contentService->loadContent(42);

$name = $content->getVersionInfo()->getName();
$value = $content->getFieldValue('body')
```

docs/api/public_php_api_browsing.md
```diff
@@ -101,7 +101,7 @@ If you want to take SiteAccess languages into account:
 -   Provide prioritized languages on `load()` this will be taken into account by the returned Content object when retrieving translated properties like fields, for example:
     `$contentService->loadContent( 66, $configResolver->getParameter('languages'));`
     -   `ConfigResolver` is a service, so it can be obtained from Symfony Container or injected directly: `@ezpublish.config.resolver`.
-    -   *Note: As of v2.0 this is planned to be done for you when you don't specify languages.*
+    -   See [below](#siteaccess-aware-repository-optional) for information about SiteAccess-aware repository.
```
andrerom: Outdent this, as it's an alternative choice, and maybe start with `Or if you want repository to inject this for you see ..` or something

andrerom:
That was 2 outdenting, but either way is fine :)

+1

PR #330, opened 2018-07-05
Added doc about injecting SiteAccess via DI Container

alongosz:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | N/A
| Versions      | 1.7+

It looks like we're missing information about how to inject `SiteAccess` into a custom service. I took [this outdated kernel doc](https://github.com/ezsystems/ezpublish-kernel/blob/v6.7.7.1/doc/specifications/siteaccess/synchronized_siteaccess.md) and rephrased it, because `ezpublish.siteaccess` is not  `synchronized`  since ezsystems/ezpublish-kernel@2cee706.


alongosz:
review ping @andrerom @DominikaK

PR #332, opened 2018-07-06
Add warning about updating to 2.2

DominikaK:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | n/a
| Versions      | all (releas notes) / 2.2 (updating)


DominikaK:
ping @andrerom

lserwatka:
Yes, looks good for now. Let's publish it ASAP.

docs/releases/updating_ez_platform.md
```diff
@@ -281,6 +281,96 @@ These steps are only relevant for some releases:
...
+    To update to v2.2, you need to run the following script to add database tables for the Page Builder:
+
+    ??? note "Database update script"
+
+        ```
+        --
+        -- Page Builder
+        --
+
+        DROP TABLE IF EXISTS `ezpage_attributes`;
```
andrerom: as schemas tend to change over time, it might be better providing path to sql to run, but risk with that is that it breaks when engineering decided to move it like happened on certain studio package in 1.x.

DominikaK: This is a temporary measure, AFAIK the script will be provided in a file later on, we'll change it to link here then.

PR #334, opened 2018-07-09
Fix path error

stevecohenfr:
Fix the path error in the tree_root.location_id example


DominikaK:
Thanks for the contribution, @ReaperSoon, good catch!

PR #337, opened 2018-07-09
Restore v1 persistence cache content

DominikaK:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | n/a
| Versions      | 1.13 (only!)

Restores the v1 version of the persistence cache page which was inadvertently overwritten with v2 content in https://github.com/ezsystems/developer-documentation/pull/87


DominikaK:
Ping @andrerom, this is based on what we have in 1.7, I have not found any mention of differences in this section between 1.7 and 1.13, could you confirm this?

PR #338, opened 2018-07-10
EZP-28986 Document commands from the EzPublishMigrationBundle

MagdalenaZuba:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | https://jira.ez.no/browse/EZP-28986
| Versions      | all

Currently, there are four cleanup commands that are included in the EzPublishMigrationBundle. It would be useful to document what they are doing, how to execute them and what issues they solve, so the developers will be able to easily find them if they encounter such issue.

### DIFF
--- a/docs/migrating/common_issues.md
+++ b/docs/migrating/common_issues.md
@@ -11,6 +11,10 @@ issues that can occur after migration to eZ Platform.
     $bundles[] = new \eZ\Bundle\EzPublishMigrationBundle\EzPublishMigrationBundle();
     ```
 
+!!! caution
+
+    Remember about **proper backup** before running any of the commands below.
+
 ## Regenerating URL Aliases
 
 Basic `ezplatform:regenerate:` command implements the cleanup script that will check if 
@@ -42,8 +46,7 @@ it in production environment using the `--env=prod` switch.
 
 ## Unknown relation type 0
 
-"Unknown relation type 0." error occurs only when using REST API. It is caused by 
-inserting  different images into two or more XML Block Fields. The issue does not occur 
+"Unknown relation type 0." error occurs only when using REST API. The issue does not occur 
 the first time article is published (upon creation). It only happens after the article is 
 edited and published.
 
@@ -54,8 +57,8 @@ php app/console ezpublish:update:legacy_storage_clean_up_relation_type_eq_zero
 ```
 The command can be executed in two modes:
 
+- list / dry-run - prints table with all corrupted relations that will be deleted (to be executed first)
 - fix - executes clean up
-- list / dry-run - prints table with all corrupted relations that will be deleted
 
 You can read more about this issue here: [EZP-27254](https://jira.ez.no/browse/EZP-27254)
 


docs/migrating/common_errors.md
```diff
@@ -0,0 +1,82 @@
...
+```
+The command can be executed in two modes:
+
+- fix - executes clean up
+- list / dry-run - prints table with all corrupted relations that will be deleted
+
+## Regenerating URL Aliases
+
+Basic `ezplatform:regenerate:` command implements the cleanup script that will check if 
+fields of given field type have correct sort key, and update it if needed.
```
DominikaK: Fields and Field Type - capitalized, also in a few other places on the page

docs/migrating/common_errors.md
```diff
@@ -0,0 +1,82 @@
...
+It is also possible to skip custom Location and global URL aliases altogether and 
+regenerate only automatically created URL aliases for Locations (use the `autogenerate` 
+action to achieve this). During the script execution the database should not be modified. 
+Since this script can potentially run for a very long time, to avoid memory exhaustion run 
+it in production environment using the `--env=prod` switch.
+
+## Always available flag set on all fields
+
+Always available flag is set on all fields, instead of only on fields in the main 
+language. This problem occurs when a new stack is used to create Content that is both
```
DominikaK: 'the new stack'? Assuming this means eZ Platform, as opposed to legacy

docs/migrating/common_errors.md
```diff
@@ -0,0 +1,82 @@
...
+regenerate only automatically created URL aliases for Locations (use the `autogenerate` 
+action to achieve this). During the script execution the database should not be modified. 
+Since this script can potentially run for a very long time, to avoid memory exhaustion run 
+it in production environment using the `--env=prod` switch.
+
+## Always available flag set on all fields
+
+Always available flag is set on all fields, instead of only on fields in the main 
+language. This problem occurs when a new stack is used to create Content that is both 
+always available and has multiple translations. Cleanup script will set correctly always available flag for prioritized language filtering in Legacy Search Engine.
```
DominikaK: `The cleanup script will correctly set...`

docs/migrating/common_errors.md
```diff
@@ -0,0 +1,82 @@
...
+
+```
+ezpublish:update:legacy_storage_fix_fields_always_available_flag
+```
+
+Only affected fields will be processed by the cleanup command.
+
+## Trouble listing sub content
+
+It is possible that after upgrade sort_key_string is left empty this may cause problems in
```
DominikaK: `sort_key_string` in backticks
`is left empty. This may...`

docs/migrating/common_errors.md
```diff
@@ -0,0 +1,82 @@
...
+
+Only affected fields will be processed by the cleanup command.
+
+## Trouble listing sub content
+
+It is possible that after upgrade sort_key_string is left empty this may cause problems in 
+searches throughout the API. The cleanup script will check if fields of given field type 
+have correct sort key, and update it if needed.
+
+Execute following command from the installation root directory:
```
DominikaK: `the following command`

docs/migrating/common_errors.md
```diff
@@ -0,0 +1,82 @@
+# Common errors
+
+Below you will find cleanup commands from the EzPublishMigrationBundle for the most common 
+issues that can manifest itself after migration to eZ Platform.
```
DominikaK: `manifest themselves`
Or maybe more simply, `occur` or `happen`?

docs/migrating/common_errors.md
```diff
@@ -0,0 +1,82 @@
+# Common errors
```
DominikaK: Maybe a more "delicate" title like `Common issues` or `Troubleshooting`?

docs/migrating/common_errors.md
```diff
@@ -0,0 +1,82 @@
...
+    ```
+
+## Unknown relation type 0
+
+"Unknown relation type 0." error occurs only when using REST API. It is caused by 
+inserting  different images into two or more XML block fields. The issue does not occur 
+the first time article is published (upon creation). It only happens after the article is 
+edited and published.
+
+If this error occurs use below console command that will clean up redundant relations rows:
```
DominikaK: `use the console command below. It will clean...`

docs/migrating/common_errors.md
```diff
@@ -0,0 +1,82 @@
...
+
+```
+ezpublish:update:legacy_storage_clean_up_relation_type_eq_zero
+```
+The command can be executed in two modes:
+
+- fix - executes clean up
+- list / dry-run - prints table with all corrupted relations that will be deleted
+
+## Regenerating URL Aliases
```
natanael89: I would put that one on top since it's probably the most commonly used.
Also, I think it's important to add a warning below that history url aliases won't be kept after running the command.

docs/migrating/common_errors.md
```diff
@@ -0,0 +1,82 @@
...
+
+"Unknown relation type 0." error occurs only when using REST API. It is caused by 
+inserting  different images into two or more XML block fields. The issue does not occur 
+the first time article is published (upon creation). It only happens after the article is 
+edited and published.
+
+If this error occurs use below console command that will clean up redundant relations rows:
+
+```
+ezpublish:update:legacy_storage_clean_up_relation_type_eq_zero
```
natanael89: Let's write the full console command everywhere. So, in this case, it would be:
`php app/console ezpublish:update:legacy_storage_clean_up_relation_type_eq_zero`
and for 2.x:
`php bin/console ezpublish:update:legacy_storage_clean_up_relation_type_eq_zero`
This applies to all commands.

PR #341, opened 2018-07-11
Fixed the doc for the ez_field twig function (return value)

bdunogier:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | none :)
| Versions      | All of them, presumably.

Fixes the doc of the `ez_field` twig function.

It does not return a Field Value (that's what `ez_field_value` does), but a Field object. Also added a short description of what a Field object is, and how it differs from a Field Value. The idea is that people will use `ez_field_value` much more often than they'll use `ez_field`.

### TODO
- [ ] Rebase on top of the oldest supported branch.

### DIFF
--- a/docs/guide/twig_functions_reference.md
+++ b/docs/guide/twig_functions_reference.md
@@ -153,7 +153,7 @@ If the Content item does not have a translation in the current language, the mai
 
 #### Description
 
-`ez_field()` is a Twig helper which returns a Field  in the current language. The field gives you access to the field value, as well as the Field's Definition and Type identifiers.
+`ez_field()` is a Twig helper which returns a Field  in the current language. The field gives you access to the field value, as well as the Field's Definition identifier and Type identifier.
 
 !!! tip
 


docs/guide/twig_functions_reference.md
```diff
@@ -153,7 +153,7 @@ If the Content item does not have a translation in the current language, the mai
 
 #### Description
 
-`ez_field()` is a Twig helper which returns a Field value in the current language.
+`ez_field()` is a Twig helper which returns a Field  in the current language. The field gives you access to the field value, as well as the Field's Definition and Type identifiers.
```
andrerom: `Field's Definition` Maybe make it more clear we are referring Field Definition identifier, as to not give impression we give field Definition.

aka `Field's Definition identifier and Type identifier`

DominikaK: I'd say the current version is fine, but people *could* misunderstand it, and adding the second `identifier` will prevent it.

DominikaK:
The oldest branch is `1.7`, but you can leave it for `master` and we'll merge it to all the needed branches (that would be 1.7, 1.13 and 2.2)

bdunogier:
About merging that way, are your branches disconnected ? e.g. can you safely merge 2.2 onto master ? I'd assume you can't, given how you suggest the merge is done :)

DominikaK:
True, merge is not the right word :) The branches are disconnected, because there are to many changes that only apply to one version.
Anyway, we can handle *placing* ( ;) ) the changes in any branch that needs them.

PR #342, opened 2018-07-16
EZP-27678: Update Anonymous X-User-Hash section

DominikaK:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | https://jira.ez.no/browse/EZP-27678
| Versions      | 1.13, 2.2, master

### DIFF
--- a/docs/guide/http_cache.md
+++ b/docs/guide/http_cache.md
@@ -526,7 +526,7 @@ eZ Platform already interferes with the hash generation process by adding the cu
 
 The anonymous `X-User-Hash` is generated based on the `anonymous user`, `group` and `role`.
 
-If you need to find out the new anonymous `X-User-Hash`:
+If you need to find out the anonymous `X-User-Hash`:
 
 1\. Connect to your server (*shh* should be enough)
 


docs/guide/http_cache.md
```diff
@@ -524,7 +524,7 @@ eZ Platform already interferes with the hash generation process by adding the cu
 
 ##### New anonymous `X-User-Hash`
```
andrerom: For 2.x we can even remove this section afaik. @kmadejski right?

kmadejski: I talked with @DominikaK about this yesterday, personally, I think that we can keep these instructions for getting `X-User-Hash` manually, just in case if somebody needs it for some debugging. Maybe this part can be stated differently, because for now it sounds like something necessary in case of changed user / group / permission. Besides that, totally right, this is not required to configure Varnish with 1.13 and up (with `PURGE` and `xkey`).

docs/guide/http_cache.md
```diff
@@ -524,10 +524,9 @@ eZ Platform already interferes with the hash generation process by adding the cu
...
 
-The anonymous `X-User-Hash` is generated based on the `anonymous user`, `group` and `role`. The `38015b703d82206ebc01d17a39c727e5` hash that is provided in the link above will work only when these three variables are left unchanged. Once you change the default permissions and settings, the `X-User-Hash` will change and Varnish won't be able to effectively handle cache anymore.
-In that case you need to find out the new anonymous `X-User-Hash` and change the VCL accordingly, else Varnish will return a no-cache header.
+The anonymous `X-User-Hash` is generated based on the `anonymous user`, `group` and `role`.
 
-The easiest way to find the new hash is:
+If you need to find out the new anonymous `X-User-Hash`:
```
andrerom: - `new`

PR #358, opened 2018-07-26
EZP-29396: Cache size blows up when using Redis

DominikaK:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | https://jira.ez.no/browse/EZP-29396
| Versions      | 1.7, 1.13

*Note*: The same information will be added to https://doc.ez.no/display/EZP/Persistence+cache+configuration for 5.4 once approved.

### DIFF
--- a/docs/guide/persistence_cache.md
+++ b/docs/guide/persistence_cache.md
@@ -237,7 +237,7 @@ stash:
     - `ezpublish.stash_cache.igbinary` enables you to use the `igbinary` library to serialize objects stored in cache.
     - `ezpublish.stash_cache.lzf` enables you to use the `LZF` library to compress serialized objects stored in cache.
 
-    After changing these settings you need to clear cache and purge Redis content.
+    After changing these settings you need to clear cache and purge Redis content (see above).
 
 ##### Redis Cluster
 


docs/guide/persistence_cache.md
```diff
@@ -222,6 +222,21 @@ stash:
     The regular `php app/console cache:clear` command does not clear Redis persistence cache.
     To clear it, use the console command shipped with Redis: `redis-cli flushall`.
 
+!!! tip
+
+    If you use the Redis cache driver and encounter problems with memory consumption,
```
kmadejski: `problems with memory consumption` -> `problems with high memory consumption`

docs/guide/persistence_cache.md
```diff
@@ -222,6 +222,21 @@ stash:
...
+    you can use the following (non-SiteAccess-aware) settings:
+
+    ``` yaml
+    ezpublish:
+        stash_cache:
+            igbinary: true
```
kmadejski: This suggests that user should set both settings to true, maybe use `bool` or `true/false` instead of `true`.

docs/guide/persistence_cache.md
```diff
@@ -222,6 +222,21 @@ stash:
...
+            lzf: true
+    ```
+
+    - `ezpublish.stash_cache.igbinary` enables you to use the `igbinary` library to serialize objects stored in cache.
+    - `ezpublish.stash_cache.lzf` enables you to use the `LZF` library to compress serialized objects stored in cache.
+
```
kmadejski: Please, add also a note that clearing cache and purge of Redis content is necessary after changing this configuration.

andrerom: Do we have an example of that somewhere perhaps? for instance using redis cli.

docs/guide/persistence_cache.md
```diff
@@ -222,6 +222,23 @@ stash:
...
+    ezpublish:
+        stash_cache:
+            igbinary: true/false
+            lzf: true/false
+    ```
+
+    - `ezpublish.stash_cache.igbinary` enables you to use the `igbinary` library to serialize objects stored in cache.
+    - `ezpublish.stash_cache.lzf` enables you to use the `LZF` library to compress serialized objects stored in cache.
+
+    After changing these settings you need to clear cache and purge Redis content.
```
andrerom: maybe add `, as explained above.`

andrerom:
Reminder: This should not go into 2.x or master ;)

PR #367, opened 2018-08-03
Service definition conform to Symfony 3.3+

arfaram:
Changing the Slot service definition with the autowiring functionality introduced in Symfony 3.3.  Dependencies Injection are passed automatically via a class's constructor

| Question      | Answer
| ------------- | ---
| Versions      | 2.0+

### DIFF
--- a/docs/cookbook/listening_to_core_events.md
+++ b/docs/cookbook/listening_to_core_events.md
@@ -54,7 +54,7 @@ In `services.yml` (in your bundle):
 services:
     _defaults:
         autowire: true
-        autoconfigure: false
+        autoconfigure: true
         public: false
  
     Acme\ExampleBundle\Slot\OnPublishSlot:
@@ -94,12 +94,10 @@ In `services.yml` (in your bundle):
 services:
     _defaults:
         autowire: true
-        autoconfigure: false
+        autoconfigure: true
         public: false
 
-    Acme\ExampleBundle\EventListener\SignalListener:
-        tags:
-            - { name: kernel.event_subscriber }
+    Acme\ExampleBundle\EventListener\SignalListener: ~
 ```
 
 ``` php


DominikaK:
ping @alongosz for review

docs/cookbook/listening_to_core_events.md
```diff
@@ -51,13 +51,13 @@ You now need to register `OnPublishSlot` as a service in the ServiceContainer an
...
-parameters:
-    my.onpublish_slot.class: Acme\ExampleBundle\Slot\OnPublishSlot
- 
 services:
-    my.onpublish_slot:
-        class: '%my.onpublish_slot.class%'
-        arguments: ['@ezpublish.api.service.content']
+    _defaults:
+        autowire: true
+        autoconfigure: false
```
alongosz: @arfaram  Why not `autoconfigure` as well? Although right now for Slots this is not gonna work out of the box (meaning we still need to keep `tags` below), we rather aim to support that in the future.

docs/cookbook/listening_to_core_events.md
```diff
@@ -90,14 +90,14 @@ This very simple example will just log the received signal.
...
-    my.signal_listener.class: Acme\ExampleBundle\EventListener\SignalListener
- 
+``` yaml 
 services:
-    my.signal_listener:
-        class: '%my.signal_listener.class%'
-        arguments: ['@logger']
+    _defaults:
+        autowire: true
+        autoconfigure: false
```
alongosz: Nope :) See [Auto-tagging with autoconfigure doc](https://symfony.com/doc/current/service_container/3.3-di-changes.html#auto-tagging-with-autoconfigure). 

This means that we should either expand this section to include PHP code example or reference that to Symfony doc. The former one is preferred, because links to external services tend to stop working at some point ;)

docs/cookbook/listening_to_core_events.md
arfaram: yes @alongosz the SignalSlotPass require the `signal` attribute.

docs/cookbook/listening_to_core_events.md
arfaram: means the signal type should be checked in the Slot `recieve()` method. IMHO, I found the actual service configuration is more comfortable.

alongosz: @arfaram you're missing the point of defaulting `autoconfigure` to `true`. This is new best practice and should be encouraged even if some of our services don't support it yet.
Here by setting `autoconfigure: true` in `_defaults` you could skip
```yml
       tags:
            - { name: kernel.event_subscriber }
```
which is great benefit once you end up with a lot of such services.

> means the signal type should be checked in the Slot recieve() method. 

this is totally not related to autoconfiguration, see the comment that will follow shortly.

docs/cookbook/listening_to_core_events.md
alongosz: Here you can also keep `autoconfigure: true` to be consistent with the other example.
It does NOT indicate that the Slot will be autoconfigured. Autoconfiguration of Slots is not yet supported.
So you still need to keep
```yaml
        tags:
            - { name: ezpublish.api.slot, signal: ContentService\PublishVersionSignal }
```
which will NOT be overridden by any autoconfiguration since no such setup exists.
This is about advocating best practices and while for this example is not necessary it comes together nicely when we see the second example. Having it different in both examples could be confusing for some developers.

alongosz:
@arfaram I've addressed your concerns, please let me know if this is clear now :)

arfaram:
@alongosz should be ok now

PR #380, opened 2018-08-20
EZP-29442: Quickstart guide

DominikaK:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | https://jira.ez.no/browse/EZP-29442
| Versions      | all

**Note** expand for 1.x, because more commands are needed (creating database etc.)

### DIFF
--- a/docs/getting_started/quickstart.md
+++ b/docs/getting_started/quickstart.md
@@ -2,14 +2,9 @@
 
 This page explains how to quickly install and set up eZ Platform for development.
 
-## Basic requirements
+## Prerequisites
 
-The recommended setup for using eZ Platform is:
-
-- Linux operating system (Debian, Ubuntu or RHEL / CentOS)
-- PHP 7.2
-    - with following packages: `php-cli`, `php-fpm`, `php-mysql`, `php-xml`, `php-mbstring`, `php-intl`, `php-curl`, `php-gd`
-- MySQL
+To quickly install eZ Platform, you need a Linux machine with PHP and MySQL installed.
 
 For more details see [Full requirements](requirements_and_system_configuration.md).
 


docs/getting_started/quickstart.md
```diff
@@ -0,0 +1,58 @@
+# Quickstart
+
+This page explains how to quickly install and set up eZ Platform for development.
+
+## Basic requirements
+
+The recommended setup for using eZ Platform is:
```
MagdalenaZuba: The recommended setup for using eZ Platform is described on the [requirements page](requirements_and_system_configuration.md).

Let's single source requirements page as it will be hard to track and update setup on many pages.

docs/getting_started/quickstart.md
```diff
@@ -0,0 +1,58 @@
...
+``` bash
+composer create-project --keep-vcs ezsystems/ezplatform .
+composer ezplatform-install
+```
+
+For more details see [Install using Composer](install_using_composer.md).
+
+See [Install manually](install_manually.md) to install eZ Platform on a different OS.
+
+!!! tip "Demo site"
```
MagdalenaZuba: As this page explains how to quickly install and set up eZ Platform for development do we want to mention demo here? 
If you want to keep it I would move the note to the top, just under the introduction.

docs/getting_started/quickstart.md
```diff
@@ -0,0 +1,58 @@
...
+
+    ``` bash
+    composer create-project --keep-vcs ezsystems/ezplatform-demo .
+    ```
+
+    The demo site is only a showcase. **Do not use it as a basis for development**.
+
+## Run eZ Platform
+
+To run eZ Platform on the built-in PHP server (for development only):
```
MagdalenaZuba: the point of this page is development set up so I would remove `(for development only)`

docs/getting_started/quickstart.md
DominikaK: Beg to differ :) I'd like this page to be something that a user can get started with without looking at other parts of the doc.
Although I do have doubts about listing the php extensions...

PR #399, opened 2018-09-12
Use the RideController with the ContentService

arfaram:
The RideController is using the repository what is unnecessary in this example. We want to use only the ContentService using constructor injection and using the symfony autowiring.

| Question      | Answer
| ------------- | ---
| Versions      | 2.0+

### DIFF
--- a/docs/tutorials/platform_beginner/4_display_a_list_of_content_items.md
+++ b/docs/tutorials/platform_beginner/4_display_a_list_of_content_items.md
@@ -113,52 +113,12 @@ and renders them using the `list/rides.html.twig` template (line 45).
 
 Because you are using the `ContentService`, you need to register the `HomepageController` as a service.
 
-Create a new file: `src/AppBundle/Resources/config/services.yml`:
+Add the following section in `app/config/services.yml`, under the `services` key:
 
 ``` yaml
-services:
-    _defaults:
-        autowire: true
-        autoconfigure: true
-        public: false
-
-    AppBundle\Controller\:
-        public: true
-        resource: '../../Controller/*'
-```
-
-At the beginning of the `ezplatform.yml` file (before the `ezpublish` key), add the following lines to import the content of the new file:
-
-``` yaml
-imports:
-    - { resource: '@AppBundle/Resources/config/services.yml' }
-```
-
-You also need to use dependency injection to merge your service into the actual container.
-
-Create a new file `src/AppBundle/DependencyInjection/AppExtension.php`:
-
-``` php
-<?php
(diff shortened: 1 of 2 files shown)


DominikaK:
@arfaram:
@ciastektk suggested that the same change you proposed here should be also applied to the HomepageController in an earlier step. I tried doing this, but please verify that I've made no mistakes in the process.

andrerom:
Q: is paging supposed to work? Cause it does not, due to this being done in sub controllers page is not passed on.

I started rewriting this to use content view controller instead after last weeks basic training. It simplifies the code, fixes paging, however open question is how it affects the other examples, ideally no need to explain that concept several times maybe(?)

docs/tutorials/platform_beginner/4_display_a_list_of_content_items.md
```diff
@@ -82,8 +101,65 @@ class HomepageController extends Controller
...
+At the beginning of the `ezplatform.yml` file (before the `ezpublish` key), add the following lines to import the content of the new file:
+
+``` yaml
+imports:
+    - { resource: '@AppBundle/Resources/config/services.yml' }
+```
+
+You also need to use dependency injection to merge your service into the actual container.
+
+Create a new file `src/AppBundle/DependencyInjection/AppExtension.php`:
```
andrerom: Don't think we need this at basic level.

Also Symfony lists services in app/config so if we end up needing to define services we can do that too:
- 3.4: https://github.com/symfony/symfony-standard/blob/3.4/app/config/services.yml
- 4.x: https://github.com/symfony/demo/blob/master/config/services.yaml

andrerom:
@DominikaK If it's ok with you I'll open PR to change it to use contentView controller instead of sub request. Simpler and more eZ'ish anyway ;)

arfaram:
@DominikaK I have some working pagination example with contentView using queryTypes. I will discuss this with @andrerom first to have a basic example.

DominikaK:
@andrerom, it's been some time, but was there ever enough rewrite done here to use in the tutorial?

andrerom:
Unfortunately I can't find the raw material I had on this, so ignore this for now.

andrerom:
@DominikaK Found it, it's raw, but if someone goes over and compared with current tutorial there are several possible imporvments to apply from this:, but there might also be irrelevant changes from PS training material and also it's missing changes done since 2.3.1: https://gist.github.com/andrerom/c410387726d4cbc88e3b75d17f48ff10

DominikaK:
This will be continued, including André's suggestions, in https://jira.ez.no/browse/EZP-30583

PR #401, opened 2018-09-13
EZP-29607: Added component Tables to UI Guidelines documentation

inakijv:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | [EZP-29607](https://jira.ez.no/browse/EZP-29607)
| Versions      | 2.2.2

Most important aspects:
- Added component Tables to UI Guidelines documentation.
(Needed also to adjust `:disabled` state for buttons within tables).

tables - ez platform developer documentation 1

### DIFF
--- a/docs/css/ez-guidelines-buttons.css
+++ b/docs/css/ez-guidelines-buttons.css
@@ -15,6 +15,11 @@
     font-size: 1em;
     color: #fff;
 }
+.ez-guidelines .ez-code-example .btn:disabled,
+.ez-guidelines .ez-code-example .btn[disabled] {
+    cursor: not-allowed;
+    opacity: .3;
+}
 .ez-guidelines .ez-code-example .btn-primary,
 .ez-guidelines .ez-code-example .btn-primary:disabled {
     background-color: #f15a10;
--- a/docs/guidelines/components/tables.md
+++ b/docs/guidelines/components/tables.md
@@ -1,0 +1,202 @@
+# Tables
+
+We use tables as main element for displaying content information.
+
+##<div class="mgt-3 header-line">Basic tables</div>
+**<div class="mgt-minus-2 mgb-3">Basic table</div>**
+<div class="ez-guidelines-tables__with-header mgb-5">
+[[code_example {html}
+<div class="ez-table-header">
+    <div class="ez-table-header__headline">System URL</div>
+</div>
+<table class="table">
+    <thead>
+        <tr>
+            <th>URL</th>
+            <th>Language</th>
+        </tr>
+    </thead>
+    <tbody>
+        <tr>
+            <td>/places-tastes</td>
+            <td>English (United Kingdom)</td>
+        </tr>
+    </tbody>
+</table>
+code_example]]
+</div>
+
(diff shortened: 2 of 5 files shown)


inakijv:
Cannot add reviewers (yet) Ping @DominikaK & @sunpietro

docs/css/ez-guidelines-buttons.css
```diff
@@ -15,6 +15,10 @@
     font-size: 1em;
     color: #fff;
 }
+.ez-guidelines .ez-code-example .btn:disabled {
```
sunpietro: What do you think about adding also: `.ez-guidelines .ez-code-example button[disabled]` selector?

docs/guidelines/components/tables.md
```diff
@@ -0,0 +1,202 @@
...
+</div>
+
+!!! Note
+    This table is displayed within `.n-notifications-modal` modal. Styling relies on this class.
+
+
+##<div class="mgt-3 header-line">List table</div>
+<div class="ez-guidelines-tables__list mgb-5">
+[[code_example {html}
+<table class="table ez-table--list">
```
sunpietro: It should be `table ez-table ez-table--list`

docs/guidelines/components/tables.md
```diff
@@ -0,0 +1,202 @@
...
+    <thead>
+        <tr>
+            <th colspan="2">
```
sunpietro: In general, the structure of `ez-table--list` will be updated very soon.

sunpietro: Please refer to this PR: https://github.com/ezsystems/ezplatform-admin-ui/pull/614

docs/css/ez-guidelines-buttons.css
inakijv: Updated based on review.

docs/guidelines/components/tables.md
inakijv: Good call @sunpietro. It's true (and I agree with you) that it should be that way, but it's what we currently have:

screen shot 2018-09-14 at 3 39 07 pm

This is also a question for you @DominikaK, Shouldn't we reflect here in the Guidelines the current status of each component? Or the standardized one? I don't have any problem updating the PR, but if this section of the documentation is going to be updated based on releases we should reflect that too. I'd go for this final option. But I need to hear you guys first.

docs/guidelines/components/tables.md
inakijv: So based on my previous comment, I'd add this new way in the upcoming 2.3 dev-doc branch, but again I need confirmation.

docs/guidelines/components/tables.md
DominikaK: As a rule we do not name versions in the text, but instead merge changes to proper version branches. So if there is a change that will not apply to 2.2, please add it in a separate PR (to master now, but with a note which versions(s) is should go to).

inakijv: Updated to `table ez-table ez-table--list` as requested.

inakijv:
Updated. Ping @DominikaK & @sunpietro

PR #406, opened 2018-09-18
EZP-29139: Regenerate URL aliases

DominikaK:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | https://jira.ez.no/browse/EZP-29139
| Versions      | all

### DIFF
--- a/docs/guide/url_management.md
+++ b/docs/guide/url_management.md
@@ -155,7 +155,9 @@ To add commands to an existing group, provide the group name and list the comman
 ## Regenerating URL aliases
 
 You can use the `ezplatform:urls:regenerate-aliases` command to regenerate all URL aliases.
-Use is when:
+After the command is applied, old aliases will redirect to the new ones.
+
+Use it when:
 
 - you change URL alias configuration and want to regenerate old aliases
 - you encounter database corruption
@@ -168,5 +170,3 @@ bin/console ezplatform:urls:regenerate-aliases
 ```
 
 Use an `--iteration-count` parameter to define how many Locations should be processed at once, to avoid too much memory use.
-
-After the command is applied, old aliases will redirect to the new ones.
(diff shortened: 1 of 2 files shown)


DominikaK:
ping @alongosz for review

docs/guide/url_management.md
```diff
@@ -151,3 +151,22 @@ A transformation group consists of an array of commands (see [all available comm
...
+
+Before applying the command, back up your database and make sure it is not modified while the command is running.
+
+``` bash
+bin/console ezplatform:urls:regenerate-aliases
+```
+
+Use an `--iteration-count` parameter to define how many Locations should be processed at once, to avoid too much memory use.
+
+After the command is applied, old aliases will redirect to the new ones.
```
alongosz: Hmm, at first I didn't notice that last line and was going to write it is missing ;) Can we advertise it better? (some border or something :D) 
It's one of the key requirements for the feature.

docs/migrating/common_issues.md
```diff
@@ -15,39 +15,19 @@ issues that can occur after migration to eZ Platform.
...
-to the migration table. They will be stored in backup tables named 
-`__migration_backup_custom_alias` and `__migration_backup_global_alias` (created 
-automatically).
-
-It is also possible to skip custom Location and global URL aliases altogether and 
-regenerate only automatically created URL aliases for Locations (use the `autogenerate` 
-action to achieve this). During the script execution the database should not be modified. 
-Since this script can potentially run for a very long time, to avoid memory exhaustion run 
-it in production environment using the `--env=prod` switch.
+    This command replaces the deprecated `ezplatform:regenerate:legacy_storage_url_aliases command.`
```
alongosz: I know you don't like to repeat the doc, but maybe we could also put here info that the reason for deprecating is that the old command didn't keep history? (could be rephrased).

PR #409, opened 2018-09-21
2.3 release notes

DominikaK:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | n/a
| Versions      | all

To do:
- [x] Add screenshots

### DIFF
--- a/docs/releases/ez_platform_v2.3.md
+++ b/docs/releases/ez_platform_v2.3.md
@@ -68,8 +68,6 @@ You can now access and set user preferences in the user menu.
 
 It is covered by the `user/preferences` Policy.
 
-On API side, this is handled by [`UserPreferenceService`](https://github.com/ezsystems/ezpublish-kernel/blob/master/eZ/Publish/API/Repository/UserPreferenceService.php).
-
 ### Dates in preferred timezone
 
 eZ Platform can now display dates across the system using timezone from User Settings.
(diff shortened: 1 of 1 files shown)


arfaram:
I think for the user preferences we should mention that:
- A new policy is now available: User/preferences
- A new API: https://github.com/ezsystems/ezpublish-kernel/blob/master/eZ/Publish/API/Repository/UserPreferenceService.php
- A new database table that must be added: https://github.com/ezsystems/ezpublish-kernel/blob/master/data/mysql/schema.sql#L1577

lserwatka:
@arfaram nope, we don't have to. `ezpreferences` is a legacy table so all instances have it.

docs/releases/ez_platform_v2.3.md
```diff
@@ -0,0 +1,112 @@
...
+eZ Platform can now display dates across the system using timezone from User Settings.
+
+### Improved selection in UDW
+
+Selection of content in Universal Discovery Widget has seen improvements,
+in particular when selecting multiple Content items.
+
+![Multiple selection on UDW](img/2.3_udw_selection.png)
+
+### API improvements
```
andrerom: I think this is covered in https://github.com/ezsystems/developer-documentation/blob/2.2/docs/releases/ez_platform_v2.2.0.md#simplified-use-of-content-and-languages-in-api

The improvment meant for 2.3 `content->getContentType()` did not make it, so that will be in 2.4 instead.

DominikaK: @andrerom, does this mean this change should also apply since 2.3? https://github.com/ezsystems/developer-documentation/pull/413

andrerom: You mean 2.2? yes the doc is valid as of 2.2, 2.3 does not afaik add anything around this.

What is [new in 2.3 in API]
(https://github.com/ezsystems/ezpublish-kernel/compare/7.2...master/) is:
- user preferences service
- asset relation type _(for now just like FIELD relation in API, but features around it is planned for the future, and it's already used in admin UI)_
- `TrashItem->trashed` timestamp for when an item was trashed _(contributed by @thiagocamposviana btw)_

DominikaK: Yes, I mean 2.2 :) And thanks for the clarification, we mixed up the versions, corrected now.

PR #410, opened 2018-09-21
Form Builder

DominikaK:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | n/a
| Versions      | master

### DIFF
--- a/docs/guide/extending_form_builder.md
+++ b/docs/guide/extending_form_builder.md
@@ -47,7 +47,9 @@
     |`integer`|Integer number|
     |`url`|URL|
     |`multiple`|Multiple choice|
-    |`select`|Checkbox|
+    |`select`|Dropdown|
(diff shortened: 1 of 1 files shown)


docs/guide/extending_form_builder.md
```diff
@@ -0,0 +1,120 @@
...
+    - `choices`
+    - `text`
+    - `integer`
+    - `url`
+    - `multiple`
+    - `radio`
+    - `action`
+    - `select`
+
+    New types of fields require a mapper. The mapper must be registered as a service:
```
adamwojs: Information which interface should be implemented is crucial here:  `\EzSystems\EzPlatformFormBuilder\FieldType\Field\FieldMapperInterface`. 

I think we should provide an example, as this is fundamental extension point :wink:

PR #413, opened 2018-09-26
[EZP-28956] - As a Developer I want access to Content on Location object, so that I can use it if I need it

MagdalenaZuba:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | https://jira.ez.no/browse/EZP-28956
| Versions      | 2.3

As a Developer I'd like to get easily access to content item when I have a location. - Documented

### DIFF
--- a/docs/api/public_php_api_locations.md
+++ b/docs/api/public_php_api_locations.md
@@ -102,7 +102,7 @@ $contentService->updateContentMetadata( $contentInfo, $contentUpdateStruct );
 
 ## Accessing content on Location object
 
-When dealing with Location objects *(and Trash objects)*, in PHP you can get access to Content object directly using `$location->getContent()`, in Twig this can also be accessed by `location.content`.
+When dealing with Location objects *(and Trash objects)*, in PHP you can get access to Content item directly using `$location->getContent()`, in Twig this can also be accessed by `location.content`.
 
 This is a lazy property. It will trigger loading of Content when first used. In case of bulk of Locations coming from Search or Location Service, the Content will also be loaded in bulk for the whole Location result set. 
 


docs/api/public_php_api_locations.md
```diff
@@ -99,3 +99,11 @@ $contentUpdateStruct->mainLocationId = 123;
  
 $contentService->updateContentMetadata( $contentInfo, $contentUpdateStruct );
 ```
+
+## Accessing content on Location object
+
+You can search for Locations and reach the content of that Location within Twig templates.
+It was possible by changing `Location->getContent()` to `location.content` and adding lazy properties support in a Twig.
```
DominikaK: `in Twig` or `in a Twig template`

docs/api/public_php_api_locations.md
```diff
@@ -99,3 +99,11 @@ $contentUpdateStruct->mainLocationId = 123;
...
+To make the process more efficient SPI `ContentHandler::loadContentList()` was added and `SearchService::findContent()` was changed to accommodate bulk loading content.
+
+To learn more about this functionality see [Lazy object properties](https://github.com/ezsystems/ezpublish-kernel/blob/master/doc/specifications/api/lazy_properties.md)
```
andrerom: Ok, so if we want to make people aware of lazy properties, which I initially didn't think neded to be mentioned in online doc, but I see it can make sense so people know why it triggers sql calls. 😮 

Then this could be written for 2.2 and up:
```
## Accessing content on Location object

When dealing with Location objects _(and Trash objects)_, in PHP you can get access to content object directly using `$location->getContent()`, in Twig this can also be accessed using `location.content` .

This is a lazy property and will trigger loading of Content when first used, in case of bulk of locations coming from Search or Location Service, the content will also be loaded in bulk for the whole Location result set. To learn more about this functionality see [Lazy object properties](https://github.com/ezsystems/ezpublish-kernel/blob/master/doc/specifications/api/lazy_properties.md)
```

PR #423, opened 2018-10-17
[2.x] Remove composer --no-dev option as we need dev for 2.x install

andrerom:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   |  N/A
| Versions      | 2.x, master

After recent simplification install doc the `--no-dev` option is back in from merges from 1.x, however for development _(development of your own web site)_ needs we need to install with dev packages.

Specifically Symfony dev environment _(default)_ will fail to work as it expects `sensio/generator-bundle` package to be installed which is defined as a dev package in `composer.json`, following symfony standard here. This has been like this since 2.0.

Todo:
- [x] #425 should go in first _(hence WIP label here)_

### DIFF
--- a/docs/cookbook/setting_up_amazon_aws_s3_clustering.md
+++ b/docs/cookbook/setting_up_amazon_aws_s3_clustering.md
@@ -64,6 +64,14 @@ ez_io:
                 adapter: aws_s3_adapter
 ```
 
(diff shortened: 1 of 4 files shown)


andrerom:
Fixed conflict, ready from my side.

PR #426, opened 2018-10-20
Basics -> Supported Response Format

reithor:
"API supports both JSON and XML in terms of response format" seems more approriate,
"response type" or "requested response type" are other options here.

| Question      | Answer
| ------------- | ---
| JIRA Ticket   | 
| Versions      |

### DIFF
--- a/docs/api/rest_api_guide.md
+++ b/docs/api/rest_api_guide.md
@@ -12,7 +12,7 @@ The REST API is available at the URI `/api/ezp/v2` . HTTPS is available as long 
 
 ## Basics
 
-REST (REpresentational State Transfer) is a web services architecture that follows the HTTP Protocol very closely. The eZ Platform REST API supports both [JSON](http://www.json.org/) and [XML](http://www.w3.org/XML/) in terms of response format.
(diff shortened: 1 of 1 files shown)


andrerom:
Technically it is both request (in case of body on creation / update of data) and response.

Maybe "accept type", or something, wdyt @bdunogier ?

bdunogier:
Well, "accept type" is a response header in the HTTP domain, so it might be confusing as well... what about "format", or maybe "exchange format" ?

bdunogier:
Well, "accept type" is a response header in the HTTP domain, so it might be confusing as well... what about "format", or maybe "exchange format" ?

bdunogier:
Well, "accept type" is a response header in the HTTP domain, so it might be confusing as well... what about "format", or maybe "exchange format" ?

bdunogier:
Well, "accept type" is a response header in the HTTP domain, so it might be confusing as well... what about "format", or maybe "exchange format" ?

DominikaK:
I'd vote for `format` by itself, seems to be the least confusing option.

PR #432, opened 2018-10-25
eZ Platform Cloud requirements

DominikaK:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | n/a
| Versions      | 1.13+

PR #440, opened 2018-10-30
[Patform.sh] Update requirements to add mention of advance legacy features

andrerom:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | N/A
| Versions      | 1.13 and higher 

Followup on recent #432.
Adds some more examples to signify what goes as supported vs experimental (custom) when it comes to advance legacy features.

Review ping @rolandbenedetti @larseirik


larseirik:
Looks good to me

PR #441, opened 2018-11-02
Install eZ Platform security update

MagdalenaZuba:
| Question      | Answer
| ------------- | ---
| Versions      | all

Security update

### DIFF
--- a/docs/getting_started/install_ez_platform.md
+++ b/docs/getting_started/install_ez_platform.md
@@ -159,7 +159,7 @@ Future files and directories created by these two users will need to inherit tho
 
 !!! caution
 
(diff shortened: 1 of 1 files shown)


docs/getting_started/install_ez_platform.md
```diff
@@ -159,7 +159,7 @@ Future files and directories created by these two users will need to inherit tho
 
 !!! caution
 
-    For security reasons, in production there is no need for web server to have write permissions to other directories.
+    For security reasons, in production web server should not have write access to other directories then `var`. Skip step above and follow link below for production needs instead.
```
glye:
```suggestion
    For security reasons, in production web server should not have write access to other directories than `var`. Skip the step above and follow the link below for production needs instead.
```

PR #446, opened 2018-11-09
EZEE-2544 Form email notifications do not work

MagdalenaZuba:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | https://jira.ez.no/browse/EZEE-2544
| Versions      | 2.3, master

Documenting Form Builder email notifications customization


docs/cookbook/customizing_form_builder_email_notifications.md
```diff
@@ -0,0 +1,24 @@
...
+    
+    |Parameter|Type|Description|
+    |---------|----|-----------|
+    |`content`|`eZ\Publish\API\Repository\Values\Content\Content`|Name of the form, its Content Type|
+    |`form`|`EzSystems\EzPlatformFormBuilder\FieldType\Model\Form`|Definition of the form|
+    |`data`|`EzSystems\EzPlatformFormBuilder\FieldType\Model\FormSubmission`|Sent data|  
+    
+    By adjusting them to your needs, you will change your email template.
+    
+    ## Configure sender details
```
adamwojs: IMHO this should be moved to https://doc.ezplatform.com/en/2.3/guide/configuration or to similar page dedicated to the Form Builder. Without this configuration the delivery of notification e-mails will be not possible.

docs/cookbook/customizing_form_builder_email_notifications.md
```diff
@@ -0,0 +1,24 @@
+# Customizing Form Builder email notifications
+
+!!! enterprise
+    
+    ## Override email template  
+    
+    Email is one of the Submit button options you can choose from in the Form Builder Interface.
+    It allows you to gather readers' email addresses to send notifications to.
```
DominikaK: I still think it sounds a bit unclear. Now it looks like we're sendind an email to the people who fill in the form (that's how I would understand "readers"), while the email in fact goes to the editor(s) who manage the form. Can you confirm I'm correct, @adamwojs ?

docs/cookbook/customizing_form_builder_email_notifications.md
MagdalenaZuba: I added the information to page dedicated to Form Builder. However I will also leave the same info in the cookbook, as recipes are complete tutorials and we want to avoid adding links to other pages in the documentation for additional steps.

PR #448, opened 2018-11-13
Improvements to eZ Platform beginner tutorial 

andrerom:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | N/A
| Versions      | 1.7+ _except commits made for specific versions and higher_

Several improvements to Beginner tutorial:
- Add tip on usage of `dump()` in order to make people aware how they can themselves look for stuff
- Phrase used for home page was inconsistent, always use name "home page", remove "root folder"
- [v1.10+] Get Content name via `content.name` _(maybe we need to add a note to explain that this maps to `Content->getName()`, but we have not explained API yet so maybe not)_
- [v2.3+] Use Image Asset field => content architecture best practice
- Always use `ez_content:viewAction`
  - pass objects directly to avoid reload => performance best practice
  - Adapt rides to get locations directly => besides perf to make sure link goes to right location if there are several
  - _Both changes have inline explanation of that in followup PR I'll apply on top of this, which will also explain when not to do this (ESI)_

### DIFF
--- a/docs/tutorials/platform_beginner/3_customize_the_front_page.md
+++ b/docs/tutorials/platform_beginner/3_customize_the_front_page.md
@@ -39,7 +39,7 @@ This tells eZ Platform to use the `template` when rendering content with Locatio
     $ php bin/console cache:clear
     ```
 
-## Create template & layout
+## Create template and layout
 
 ### Create the first template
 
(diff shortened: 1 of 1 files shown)


DominikaK:
@andrerom as an explanation, I've moved the tip about `dump()` to a place after we first use a template with variables, to make is easier to understand.

andrerom:
@DominikaK Agreed, my thinking was that advance developers, including those used to eZ Publish, would be curious about this as soon as they are past the "hello word" step. But for developers that are completely new to usage of templates it's better we give a bit more context first, so better like this.

PR #453, opened 2018-11-15
EZP-29835: Cover sorted composer packages in update doc and other smaller improvements 

andrerom:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   |  [EZP-29835](https://jira.ez.no/browse/EZP-29835)
| Versions      | 1.7 - master _(all versions)_

EZP-29835 enabled composer's [sort-packages](https://getcomposer.org/doc/06-config.md#sort-packages), this will once installs have it simplify mergs, but on initial update with this they will get more conflicts then usual so we need to cover that in update doc on how to best get true it.

Besides that some other smaller changes:
- Smaller fixes on version numbers
- Update links to where download files comes from
- Simply instructions around "upstream" handling

+ A 2.4 or 2.3 or 2.2 specific change depending on what release of Commerce we considere as released:  
- add mention of Commerce installations 

TODO:
- [x] Do the change for composer.json on ezplatform:master repo and merge up to all repos

### DIFF
--- a/docs/releases/updating_ez_platform.md
+++ b/docs/releases/updating_ez_platform.md
@@ -32,7 +32,7 @@ or
 git remote add upstream http://github.com/ezsystems/ezcommerce.git
 ```
 
-**1.3** Prepare for pulling changes
+**1.3.** Prepare for pulling changes
 
 ??? note "Adding `sort-packages` option when updating from <=1.7.8, 1.13.4, 2.2.3, 2.3.2"
 


andrerom:
@DominikaK The change in question on meta repos has been done now, so updated PR here to reflect it.

docs/releases/updating_ez_platform.md
```diff
@@ -3,53 +3,106 @@
...
+git remote add upstream http://github.com/ezsystems/ezplatform-ee.git
 or
-git remote add ezplatform-ee http://github.com/ezsystems/ezplatform-ee.git
+git remote add upstream http://github.com/ezsystems/ezcommerce.git
 ```
 
-**1.3.** Then pull the tag into your branch.
+**1.3** Prepare for pulling changes
+
+??? note "Adding `sort-packages` option when updating from <1.7.9, 1.13.5, 2.2.4, 2.3.3, 2.4.0"
```
DominikaK: Could we change the way the versions are listed here? Right now it's a bit confusing, because there is no 2.3.3, 2.4.0 yet (and the difference between `<` and `<=` can be easily overlooked).
Would it be OK to change it to: `updating from <=1.7.8, 1.13.4, 2.2.3, 2.3.2`?

andrerom: yes, that is fine with me

andrerom:
@DominikaK Can you check who in Engineering can do review before we pass to QA?

andrerom:
> TBH these upgrade instructions do not follow the idea behind composer projects. When you do composer create-project you're creating something which is yours based on the given skeleton. We should not require people to keep our own vcs history. If I have my own project I want to have my own history and nothing else.
> 
> So we should rather link to composer.json (or maybe to a patch for it) and also list what changed (both for package versions, new bundles and meta-repository changes which enable new features).

I agree, but we put to much things in meta repos, so if we want people to be able to pull in changes it's the only way we found with @bdunogier some years ago.

Flex will solve this, no meta by us anymore, no problems like this... 🤷‍♂️ 😏

SylvainGuittard:
About `vcs`, @Plopix opened a PR several months ago on the same repo (#284)
I agree with @alongosz when it comes to a project, you don't want to keep the history of eZ Platform.

Plopix:
yes, I agree with @alongosz so much. We don't want the history it does not make sense.

That is not the only problem, to quote myself in the other PR:

> I am pretty convinced that a script that would do a "smart" rsync  will be much better than this procedure. Again that procedure does not work if ezplatform is not the "root directory". that argument alone should be enough to review it.

My opinion is to have a smart script that does not just "rsync" but some other actions to simplify the decisions and actions of the developer who is doing the merge.
Not everything can be automated but most of the things.

Honestly, I have done it on multiple projects where `ezplatform` is not the root directory( like with Launchpad but not only)
=> on Platform.sh you can have multiple `app`. Ex: let's say one in Go and your eZ, in this situation (that will arrive much more often I suppose with time) you don't have eZ at the root or your GIT.

Then the rsync + manual actions that could be automated are pretty efficient.

Also, I agree Flex will simplify that, but it won't solve anything, we should abstract that already and Flex would simplify this abstraction.

This was disgression but well my PR to initiate the idea as 7 months ;)

This PR goes in the good direction !

andrerom:
@SylvainGuittard But this is something we should handle outside of this PR right? Aka this can still be tested by QA so we have it before we do new release? _(and people get blocked)_

SylvainGuittard:
@andrerom indeed, we can handle that outside of this PR. I just wanted to share that it's not the first time that I hear about that from our users. 
Last month on slack:
```
punknroll [Oct 18th at 10:19 AM]
Hi, I tried to update to to latest version as described in the docs https://doc.ezplatform.com/en/latest/releases/updating_ez_platform/ now my masterbranch is "polluted" with all the commits of ezsystems. Any ideas what went wrong here?

kossuth [1 month ago]
Yeah, had the same issue. What I did was I started a "new" project with a clean install of the new eZ version, and copied the new files to my project dir and checked the changes one by one

kossuth [1 month ago]
I'm not sure why the official upgrade is done this way, it really messes up git history
```

So let's not forget it ;)

alongosz:
Yeah, we can move forward with this, sorry for the off-topic Guys, just didn't know where to post that ;)

bdunogier:
Reading this in retrospect, in regards to upgrades: I'm thinking of a procedure centered around updating composer.json and AppKernel. Kind of what Flex does, but adapted to our current situation.

With that procedure + composer update, we should have a simple and yet efficient solution.

andrerom:
@bdunogier your comment is more suited for #284, this is about `sort-packages` option added to composr.json recently.

PR #454, opened 2018-11-16
Added tip about headers length configuration in Varnish

kmadejski:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | ---
| Versions      | `1.13`/`2.2`/`2.3`/`master`

During the discussion with @dspe and @andrerom we figured-out that it's necessary to document tips about dealing with very long `xkey` headers which might be automatically truncated by Varnish and led to cache inconsistencies. 

@dspe can't add you to the reviewers, but I'll appreciate your feedback as well.

### DIFF
--- a/docs/guide/http_cache.md
+++ b/docs/guide/http_cache.md
@@ -609,8 +609,8 @@ based on your decision.
 
 !!! tip Troubleshooting - Content tagged by a big number of tags (too long headers)
     
-    In case of complex Content, for instance Landing Pages with many blocks, you might get into trouble with too long response `xkey` header. Because of this, necessary cache entries may not be tagged properly. You will see `502 Headers too long` errors then.
-    If this is your case, run your Varnish instance(s) with increased values of the following runtime settings:
+    In case of complex Content, for instance Landing Pages with many blocks, you might get into trouble with too long response `xkey` header. Because of this, necessary cache entries may not be tagged properly. You will also see `502 Headers too long` errors.
+    If this is the case, customize the following runtime settings on your Varnish instance(s):
     
     - [http_resp_hdr_len](https://varnish-cache.org/docs/5.1/reference/varnishd.html#http-resp-hdr-len) (e.g. 64k)
     - [http_req_hdr_len](https://varnish-cache.org/docs/5.1/reference/varnishd.html#http-max-hdr) (e.g. 64k)


dspe:
HI @kmadejski . If im right, you need to increase the 3 values when you get a 502 Headers too long errors (to be confirmed). Else its not necessary

kmadejski:
Thanks, @dspe. Added information about `502 Headers...` as you mentioned in https://github.com/ezsystems/developer-documentation/pull/454/commits/621887e693dcc8b57e78548a6941c7b72816ba30.

docs/guide/http_cache.md
```diff
@@ -607,6 +607,17 @@ Note that the system does not add this tag to responses itself, just purges if p
 Response tagging using this tag is currently meant to be done inline in the template logic / views
 based on your decision.
 
+!!! tip Troubleshooting with contents tagged by the huge number of tags (too long headers)
+    
+    In case of complex contents, like for instance Landing Pages with many blocks, you might get into trouble with too long PURGE request `xkey` header what might cause not all of necessary cache entries being cleared. You can see `502 Headers too long` errors then.
```
andrerom: It's currently not a problem for PURGE actually, or should not be, as we do one purge per tag atm: https://github.com/ezsystems/ezplatform-http-cache/blob/master/src/PurgeClient/VarnishPurgeClient.php#L37-L49

However that will change once we move to FOSHttpCache v2.

Anyway, it's for now more a issue for responses I think, afaik tags being stripped off before cache is stored and tagged by xkey, meaning they won't be invalidated by the tags that got stripped.

kmadejski: Yes, you are totally right, changed in https://github.com/ezsystems/developer-documentation/pull/454/commits/e4af94272995a573996d03e4b8469c20862b8a40.

docs/guide/http_cache.md
```diff
@@ -607,6 +607,17 @@ Note that the system does not add this tag to responses itself, just purges if p
 Response tagging using this tag is currently meant to be done inline in the template logic / views
 based on your decision.
 
+!!! tip Troubleshooting - Content tagged by a big number of tags (too long headers)
+    
+    In case of complex Content, for instance Landing Pages with many blocks, you might get into trouble with too long response `xkey` header. Because of this, necessary cache entries may not be tagged properly. You will see `502 Headers too long` errors then.
```
andrerom: 👍 

Could be adapted to something like:
> In case of complex Content, for instance Landing Pages with many blocks, you might get into trouble with too long response `xkey` header. Because of this, necessary cache entries may not be tagged properly, and you'll see `502 Headers too long` errors.
> If this is the case, please customize the following runtime settings on your Varnish instance(s):
> 
>     - [http-resp-hdr-len](https://varnish-cache.org/docs/5.1/reference/varnishd.html#http-resp-hdr-len) _(e.g. 32k)_
>    - [http_req_hdr_len](https://varnish-cache.org/docs/5.1/reference/varnishd.html#http-max-hdr) _(e.g. 64k)_
 >   - [http-resp-size](https://varnish-cache.org/docs/5.1/reference/varnishd.html#http-resp-size) _(e.g. 16k)_

andrerom: Please adjust the example values to something you find sensible, this was only a suggestion for guidline.

PR #456, opened 2018-11-16
Updates to requirements (Intro on Cloud, mariaDB 10.3 support, php versions)

andrerom:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | N/A / [EZP-29740](https://jira.ez.no/browse/EZP-29740)
| Versions      | 1.7+ _(!! See version info on commits !!)_

1.7+:
- Update intro to have correct info on eZ Platform Cloud support instead of implying you can ignore requirements on platform.sh / ez cloud
- mariadb 10.3, getting supported now but not yet recommended on 2.x until it has passed full sanities
- Small: clarification on version usage in regards to performance recommendations

1.13 only:
- php versions issue _(probably merge conflict, 5.6 and 7.1 was missing here)_

1.13+:
- Further fixes on intro to reflect Cloud support on versions where it is supported

PR #457, opened 2018-11-16
[1.7] Update PAAS requirement text

andrerom:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | N/A
| Versions      | 1.7

PR for changes to 1.13+: #456

### DIFF
--- a/docs/getting_started/requirements.md
+++ b/docs/getting_started/requirements.md
@@ -2,7 +2,7 @@
 
 ## Platform as a Service (PaaS)
 
-If you're using a PaaS provider such as our partner [Platform.sh](https://platform.sh/hosting/php/ez/), where we have an single-server setup, and in the future also clustered setup, you can [skip](starting_ez_platform.md#hello-world) this step.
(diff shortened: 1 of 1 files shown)


rolandbenedetti:
Good for me to change that indeed. I would adjust, few typos:

> _As of v1.13 and higher you have the option to subscribe to [eZ Platform Cloud](https://ez.no/Products/eZ-Platform-Cloud), a PaaS offering officially supported by eZ Systems. For other cloud hosting solutions and on-premise or private cloud hosting, you can refer to the requirements below._

DominikaK:
@andrerom if this note refers to 1.13, do we want to add it to the 1.7 branch?

andrerom:
@DominikaK The point of the comment here is more to make it clear we _don't_ support eZ Platform Cloud on 1.7, while also updating the intro across all branches which is out of date.

DominikaK:
Understood, so I second Roland's improvements, they make the fact that Cloud is not supported in 1.7 a bit clearer.

andrerom:
@DominikaK Sure, but what about suggestion in 1.13 PR comment? (last one has suggestion for both 1.7 and 1.13+)

DominikaK:
@andrerom agree, the version from the other PR is even better. If you're OK with that, we can do if with just that one PR, and introduce the changes between 1.7 and 1.13 while merging?

andrerom:
Agreed

PR #461, opened 2018-11-21
EZP-22390: Access controls to custom controllers

DominikaK:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | https://jira.ez.no/browse/EZP-22390
| Versions      | all

### DIFF
--- a/docs/guide/permissions.md
+++ b/docs/guide/permissions.md
@@ -217,9 +217,9 @@ public function performAccessCheck(): void
 - `function` is the function inside the module (e.g. `read`)
 - `limitations` are optional limitations to check against. Here you can provide two keys:
     - `valueObject` is the object you want to check for, for example `ContentInfo`.
-    - `targets` is a table of value objects that are the target of the operation.
-    For example, to check if content can be assigned to a Section, provide the Section as `targets`.
-    `targets` accepts Location, Object State and Section objects.
+    - `targets` are a table of value objects that are the target of the operation.
+    For example, to check if Content can be assigned to a Section, provide the Section as `targets`.
+    `targets` accept Location, Object State and Section objects.
 
 ### Checking user access
 


docs/guide/permissions.md
```diff
@@ -194,3 +194,54 @@ Users are treated like other Content, so to create and modify them the User need
...
+}
+```
+
+`Attribute` accepts three arguments:
+
+- `module` is the Policy module (e.g. `content`)
+- `function` is the function inside the module (e.g. `read`)
+- `limitations` are optional limitations to check against. Here you can provide two keys:
+    - `valueObject` is the object you want to check for, for example `ContentInfo`.
+    - `targets` is a table of value objects that are the target of the operation.
```
MagdalenaZuba: targets are or limitations is above

docs/guide/permissions.md
```diff
@@ -194,3 +194,54 @@ Users are treated like other Content, so to create and modify them the User need
...
+    For example, to check if content can be assigned to a Section, provide the Section as `targets`.
+    `targets` accepts Location, Object State and Section objects.
```
MagdalenaZuba: `targets` accept

docs/guide/permissions.md
```diff
@@ -194,3 +194,54 @@ Users are treated like other Content, so to create and modify them the User need
...
+```
+
+`Attribute` accepts three arguments:
+
+- `module` is the Policy module (e.g. `content`)
+- `function` is the function inside the module (e.g. `read`)
+- `limitations` are optional limitations to check against. Here you can provide two keys:
+    - `valueObject` is the object you want to check for, for example `ContentInfo`.
+    - `targets` is a table of value objects that are the target of the operation.
+    For example, to check if content can be assigned to a Section, provide the Section as `targets`.
```
MagdalenaZuba: Content as below

PR #471, opened 2018-11-29
Adding and overriding block templates

DominikaK:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | n/a
| Versions      | 2.3, master


DominikaK:
Note: it looks like some other solution to this is in the works in https://jira.ez.no/browse/EZEE-2557

PR #474, opened 2018-12-04
Updating to 2.3 - Missing information about new Form Builder logic

mateuszbieniek:
Currently, "update to 2.3" instructions do not cover at all logic behind new Form Builder (that forms are no longer "block values" but content) and how to add them to the Landing Pages.

Basically, this PR states that after following all the steps form docs, Developer has to create new Content Type that would contain Form Field Type.

### DIFF
--- a/docs/releases/updating_ez_platform.md
+++ b/docs/releases/updating_ez_platform.md
@@ -506,8 +506,7 @@ Some versions require updates to the database. Look through [the list of databas
 
     !!! caution "Form (ezform) Field Type"
     
-        In eZ Platform 2.3 forms are content like everything else. After the update, in order to create forms, you have to 
-        add new Content Type (e.g. named "Form") that contains `Form` field (this Content Type can contain other fields 
+        After the update, in order to create forms, you have to add a new Content Type (e.g. named "Form") that contains `Form` Field (this Content Type can contain other fields 
         as well). After that you can use forms inside Landing Pages via Embed block.
 
 ## 5. Dump assets


docs/releases/updating_ez_platform.md
```diff
@@ -504,6 +504,12 @@ Some versions require updates to the database. Look through [the list of databas
         ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
         ```
 
+    !!! caution "Form (ezform) Field Type"
+    
+        In eZ Platform 2.3 forms are content like everything else. After the update, in order to create forms, you have to
```
DominikaK: Can we get rid of the first sentence? This page is complex enough with all the instructions, I don't think we need an explanation here.

docs/releases/updating_ez_platform.md
```diff
@@ -504,6 +504,12 @@ Some versions require updates to the database. Look through [the list of databas
         ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
         ```
 
+    !!! caution "Form (ezform) Field Type"
+    
+        In eZ Platform 2.3 forms are content like everything else. After the update, in order to create forms, you have to 
+        add new Content Type (e.g. named "Form") that contains `Form` field (this Content Type can contain other fields
```
DominikaK: "add a new Content Type" ... "contains a `Form` Field"... and capital "Field" everywhere

mateuszbieniek: Thank you for pointing it out! Fixed.

PR #477, opened 2018-12-07
Beginner tutorial: Fix path for previewing rides

DominikaK:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | n/a
| Versions      | v2+

Fixes a problem when previewing Ride results in an error.

Note: the issue is actually fixed here: https://github.com/ezsystems/ezpublish-kernel/pull/2478, but the new URL is nicer and simpler to use anyway.


andrerom:
+1

Note, while the url is prettier:
- awareness of system url is probably a good thing too
- url alias will not be the exact same as name of the content, by default lowercased and spaces changed and so on..

PR #473, opened 2018-12-04
Clarify use of database update scripts in update guide

DominikaK:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | n/a
| Versions      | all

In the update guide, this points to the main database update script in the section for each version, in case users overlook the general information at the start of the Update database section.

PR #479, opened 2018-12-08
EZP-29874: Add trashed instruction for update doc

andrerom:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | <https://jira.ez.no/browse/EZP-29874
| Versions      | 2.3 and up

Suggestion to reduce number of people running in to this, add section for trashed column in 2.3 while #473 is moving forward.


andrerom:
Yes

PR #490, opened 2018-12-20
Suggestion: Improving general readability

JorgenSolli:
A suggestion on how to increase the readability of the in-line code snippets in the docs.
The current contrast between the background color and the actual color of the text is too low.
The font is also very thin, and thus hard to read.

I've also suggested giving the first row in tables more space to avoid severe linebreaking on long strings.

Current:
image

Suggested changes:
image

### DIFF
--- a/docs/css/custom.css
+++ b/docs/css/custom.css
@@ -245,7 +245,6 @@ body {
     margin: 0;
     box-shadow: 0 1px 1px rgba(0, 0, 0, 0.125);
     color: #2b2b2b;
-    font-size: 80%;
     font-family: monospace;
 }
 


docs/css/custom.css
```diff
@@ -240,10 +240,13 @@ body {
 }
 
 .md-typeset code {
-    background-color: #f0f0f0;
+    background-color: #ededed;
     padding: 1px 4px;
     margin: 0;
-    box-shadow: none;
+    box-shadow: 0 1px 1px rgba(0, 0, 0, 0.125);
+    color: #f05b01;
```
DominikaK:
```suggestion
    color: #2b2b2b;
```

We use a very similar orange color for links, so we can't have monospace in orange as well, but regular dark gray should be readable enough with your other suggestions (also passes WCAG 2 level AA)

docs/css/custom.css
```diff
@@ -240,10 +240,13 @@ body {
...
+    font-size: 80%;
```
DominikaK:
```suggestion
```

docs/css/custom.css
JorgenSolli: Yeah I agree. #2b2b2b works well with monospace.

PR #527, opened 2019-02-08
Solr tested version is 6.6.2

glye:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | N/A
| Versions      | All active branches

The currently tested version is 6.6.2, see https://doc.ezplatform.com/en/2.4/getting_started/requirements/

### DIFF
--- a/docs/guide/solr.md
+++ b/docs/guide/solr.md
@@ -46,9 +46,9 @@ bin/solr start -f -a "-Dsolr.solr.home=multicore"
 
 ##### Solr 6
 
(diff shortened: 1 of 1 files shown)


andrerom:
Actually it's 6.6.5 on solr bundle: https://github.com/ezsystems/ezplatform-solr-search-engine/commit/2e9788642b10a573102285d81ce27b93cd1380b3

But probably better if we skip being specific and say 6.6LTS, goes for both pages across all 1.7+ versions and up.

andrerom:
Edit: Like done for solr.md here: https://github.com/ezsystems/developer-documentation/pull/524/files

PR #530, opened 2019-02-14
Clarify: Use a proper web server in prod, and don't interpret var

glye:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | N/A
| Versions      | All active branches

Clarify that PHP's web server is dev only, and that your production vhost must not interpret anything in the var directory. If similar instructions are elsewhere, please make similar changes there.

### DIFF
--- a/docs/getting_started/install_ez_platform.md
+++ b/docs/getting_started/install_ez_platform.md
@@ -165,7 +165,7 @@ Future files and directories created by these two users will need to inherit tho
 
     For security reasons, in production web server should not have write access to other directories than `var`. Skip the step above and follow the link below for production needs instead.
 
-    You must also make sure that the web server cannot interpret files in the `var` directory through PHP. Follow the instructions on [setting up a virtual host](#set-up-virtual-host), as these take care of this for you.
+    You must also make sure that the web server cannot interpret files in the `var` directory through PHP. To do so, follow the instructions on [setting up a virtual host below](#set-up-virtual-host).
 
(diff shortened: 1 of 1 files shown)


docs/getting_started/install_ez_platform.md
```diff
@@ -161,6 +165,8 @@ Future files and directories created by these two users will need to inherit tho
 
     For security reasons, in production web server should not have write access to other directories than `var`. Skip the step above and follow the link below for production needs instead.
 
+    You must also make sure that the web server cannot interpret files in the `var` directory through PHP. Follow the instructions on [setting up a virtual host](#set-up-virtual-host), as these take care of this for you.
```
DominikaK: Not sure if we need the second sentence, "Set up virtual host" is four lines down, and it's part of the procedure anyway, so it's not likely to be missed.

glye: If we just remove the 2nd sentence, people will wonder "how do I make sure of this?". We got a bug report due to someone not following those virtual host steps, that's why I wanted to clarify it. There's an important security aspect to this. It can be rewritten, of course.

DominikaK:
```suggestion
    You must also make sure that the web server cannot interpret files in the `var` directory through PHP. To do so, follow the instructions on [setting up a virtual host below](#set-up-virtual-host).
```

glye:
So... can I merge, or will you doc people do that? :)

PR #541, opened 2019-03-04
Add tips on layout settings before page migration

jlchassaing:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | 
| Versions      | 2.4

During the page block migration the script will use the defined layout. If it has not been set correctly the default layout will be use and an exception will be thrown if some specific layout is not found

### DIFF
--- a/docs/releases/updating_ez_platform.md
+++ b/docs/releases/updating_ez_platform.md
@@ -387,7 +387,6 @@ Some versions require updates to the database. Look through [the list of databas
         from `ez_systems_landing_page_field_type` under `ezplatform_page_fieldtype` in the new config.
         Otherwise the script will encounter errors.
         
-    !!! tip
         
     You can remove the bundle after the migration is complete.
 


DominikaK:
Thanks for the contribution @jlchassaing 
I think we could make the note a bit more concise, so it's easier to read. What do you think about:

```
This script will use the layout defined in your Landing Page.
To migrate successfully, you need to copy your zone configuration
from `ez_systems_landing_page_field_type` under `ezplatform_page_fieldtype`
in the new config.

Otherwise the script will encounter errors.
```

jlchassaing:
>Thanks for the contribution @jlchassaing  ...

@DominikaK yes that sounds better.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

docs/releases/updating_ez_platform.md
```diff
@@ -380,6 +380,15 @@ Some versions require updates to the database. Look through [the list of databas
...
     3. Run command `bin/console ezplatform:page:migrate`
 
+    !!! tip
+    
+        This script will use the layout defined in your Landing Page.
+        To migrate successfully, you need to copy your zone configuration 
+        from `ez_systems_landing_page_field_type` under `ezplatform_page_fieldtype` in the new config.
+        Otherwise the script will encounter errors.
+        
+    !!! tip
```
DominikaK:
```suggestion
```
One more thing, the second tip tag is not necessary with our rendering.

PR #544, opened 2019-03-05
EZP-29802: Inline custom tags

DominikaK:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | https://jira.ez.no/browse/EZP-29802
| Versions      | master (2.5)

### DIFF
--- a/docs/guide/extending_online_editor.md
+++ b/docs/guide/extending_online_editor.md
@@ -183,7 +183,7 @@ ezrichtext:
 ```
 
 `is_inline` is an optional key.
-The default value is `false`, so if it is not set, the custom tag will be a block tag.
+The default value is `false`, so if it is not set, the custom tag will be treated as a block tag.
 
 ## Custom styles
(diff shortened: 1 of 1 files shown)


docs/guide/extending_online_editor.md
```diff
@@ -167,6 +167,24 @@ ezrichtext.custom_tags.ezfactbox.attributes.name.label: 'Name'
...
+        badge:
+            template: 'AppBundle:field_type/ezrichtext/custom_tag:badge.html.twig'
+            icon: '/bundles/ezplatformadminui/img/ez-icons.svg#bookmark'
+            is_inline: true
+            attributes:
+                # ...
+```
+
+`is_inline` is an optional key.
+The default value is `false`, so if it is not set, the custom tag will be a block tag.
```
MagdalenaZuba: `will be shown as a block tag` or `will be treated as a block tag`?

PR #568, opened 2019-03-28
Solr boosting clarification

kmadejski:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | N/A
| Versions      | 1.7, 1.13, 2.4, master

This PR clarifies:
- scoring on location search while using Solr 4.x
- the configuration of boosting on particular ContentType's fields using 3rd party bundle

### DIFF
--- a/docs/guide/solr.md
+++ b/docs/guide/solr.md
@@ -306,7 +306,7 @@ The configuration above will result in the following boosting (Content Type / Fi
 !!! tip "How to configure boosting on specific fields"
 
     Currently, boosting on particular fields is missing.
-    However, it could be configured using 3rd party [Novactive/NovaeZSolrSearchExtraBundle](https://github.com/Novactive/NovaeZSolrSearchExtraBundle) in case of custom search implementation, for instance for your front-end search form.
+    However, it could be configured using 3rd party [Novactive/NovaeZSolrSearchExtraBundle](https://github.com/Novactive/NovaeZSolrSearchExtraBundle) in case of custom search implementation, e.g. to handle your front-end search form.
     Unfortunately, this doesn't affect search performed in the administration interface. 
     
     The following example presents boosting configuration for Folder's `name` and `description` fields. 


docs/guide/solr.md
```diff
@@ -300,13 +299,82 @@ ez_search_engine_solr:
...
-- `news/title: 3.0`
-- `blog_post/title: 1.5`
 - `news/description: 1.0` (default)
 - `article/text (meta): 5.0`
 - `blog_post/name (meta): 10.0`
 - `article/name (meta): 2.0`
 
+!!! tip "How to configure boosting on specific fields"
+
+    Currently, boosting on particular fields is missing. However, it could be configured using 3rd party [Novactive/NovaeZSolrSearchExtraBundle](https://github.com/Novactive/NovaeZSolrSearchExtraBundle) in case of custom search implementation, for instance for your front-end search form. Unfortunately, this doesn't affect search perfomed in PlatformUI.
```
SylvainGuittard:
```suggestion
    Currently, boosting on particular fields is missing. However, it could be configured using 3rd party [Novactive/NovaeZSolrSearchExtraBundle](https://github.com/Novactive/NovaeZSolrSearchExtraBundle) in case of custom search implementation, for instance for your front-end search form. Unfortunately, this doesn't affect search performed in the administration interface. 
```

SylvainGuittard:
Thanks @kmadejski !

docs/guide/solr.md
```diff
@@ -300,13 +299,82 @@ ez_search_engine_solr:
...
+                    'metaBoost' => [
+                        'custom_folder_name' => 20.0,
+                        'custom_folder_description' => 10.0,
+                    ]
+                ]
+            );
+    
+            $searchResult = $this->searchService->findContent($query);
+            
+            ...
```
mateuszbieniek: Can we put those `...` in the comment? So we will have a valid PHP code?

docs/guide/solr.md
mateuszbieniek: > for instance for your front-end search form
Two "for's" in 3 words, maybe it would be better to rewrite this part to something like:
> for instance to handle your front-end search form

docs/guide/solr.md
kmadejski: This is done on purpose, since this code isn't valid even without these dots 😉

docs/guide/solr.md
kmadejski: Hmm, right. @MagdalenaZuba offered to fix some language mistakes, so I hope she'll cover also this one.

docs/guide/solr.md
andrerom: @kmadejski to avoid this, can we slim down the example to only cover the important bits? So it's clear it's not a copy paste full example

as in, drip the whole class and all, more like:
```
use eZ\Publish\API\Repository\Values\Content\Query;
use Novactive\EzSolrSearchExtra\Query\Content\Criterion\MultipleFieldsFullText;

// (...)

$query = new Query();
$query->query = new MultipleFieldsFullText(
    $queryString,
    [
        'metaBoost' => [
            'custom_folder_name' => 20.0,
            'custom_folder_description' => 10.0,
        ]
    ]
);

$searchResult = $this->searchService->findContent($query);
```

andrerom: Ops, to late :)

PR #569, opened 2019-03-28
EZP-30366 Allow to use custom CK Editor plugins in Alloy Editor

MagdalenaZuba:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | https://jira.ez.no/browse/EZP-30366
| Versions      | master

### DIFF
--- a/docs/guide/extending_online_editor.md
+++ b/docs/guide/extending_online_editor.md
@@ -268,7 +268,7 @@ The configuration is available at:
 ```yaml
 ezrichtext:
     alloy_editor:
-        extra_plugins: [test1, test2]
+        extra_plugins: [plugin1, plugin2]
 ```
 
 The name of a plugin needs to be the same as the one passed to `CKEDITOR.plugins.add` in the plugin source code.


docs/guide/extending_online_editor.md
```diff
@@ -255,3 +255,16 @@ In the example above, the template files for the front end could be:
 ```
 
 Templates for Content View in the Back Office would be `app/Resources/views/themes/admin/field_type/ezrichtext/custom_style/highlighted_word.html.twig` and `app/Resources/views/themes/admin/field_type/ezrichtext/custom_style/highlighted_block.html.twig` respectively (assuming Admin SiteAccess uses the `admin` theme).
+
+## Plugins configuration
+
+To enable new CKEditor plugins in Alloy Editor you need to define them in RichText AlloyEditor Semantic Configuration.
```
alongosz: @dew326 your input here would be valuable. Is it about these https://ckeditor.com/cke4/addons/plugins/all? If so, do we just need to pass them based on that list or are additional actions required. It would be useful to enrich our doc with such precise information.

SylvainGuittard: Definitely agree. We need to be clear here.

dew326: Yes, you need to develop or download a plugin to the CKEditor. You need to add them to the CKEditor plugins by `add` method. If you download from the CKEditor you probably need to only include them on a page.
Example of our plugin: https://github.com/ezsystems/ezplatform-admin-ui/blob/master/src/bundle/Resources/public/js/alloyeditor/src/plugins/ez-add-content.js#L98

SylvainGuittard: Activate a plugin is one thing, make it work with our storage engine is a different story.
@alongosz I think we should be clear on potential limitations, especially with DocBook.

docs/guide/extending_online_editor.md
```diff
@@ -255,3 +255,20 @@ In the example above, the template files for the front end could be:
 ```
 
 Templates for Content View in the Back Office would be `app/Resources/views/themes/admin/field_type/ezrichtext/custom_style/highlighted_word.html.twig` and `app/Resources/views/themes/admin/field_type/ezrichtext/custom_style/highlighted_block.html.twig` respectively (assuming Admin SiteAccess uses the `admin` theme).
+
+## Plugins configuration
+
+If you develop your plugin, you need to add it to the CKEditor plugins by `add` method.
+For more information follow [Creating a CKEditor Plugin tutorial.](https://ckeditor.com/docs/ckeditor4/latest/guide/plugin_sdk_sample.html)
```
michalmilc90:
```suggestion
For more information, follow [Creating a CKEditor Plugin tutorial.](https://ckeditor.com/docs/ckeditor4/latest/guide/plugin_sdk_sample.html)
```

docs/guide/extending_online_editor.md
```diff
@@ -255,3 +255,20 @@ In the example above, the template files for the front end could be:
...
+If you downloaded a plugin from the CKEditor, you need to include it in a page after AlloyEditor is loaded.
+
+To enable your new CKEditor plugin in, define it in the RichText AlloyEditor Semantic Configuration.
+
+The configuration is available at:
+
+```yaml
+ezrichtext:
+    alloy_editor:
+        extra_plugins: [test1, test2]
```
alongosz:
```suggestion
        extra_plugins: [plugin1, plugin2]
```
Below the Yaml block we need to add the following note:
> The name of a plugin needs to be the same as the one passed to `CKEDITOR.plugins.add` in the plugin source code.
>
> Please keep in mind that if a plugin changes RichText input (in `xhtml5/edit` format for DocBook), the changes need to be supported by RichText Field Type. For example, if a plugin adds some class to some element, you need to confirm that this class is stored when saving or publishing content (it could result in either XML validation error or could be just omitted by RichText processor).

MagdalenaZuba: added :)

PR #578, opened 2019-04-02
Rephrase supported setups in requirements

DominikaK:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | n/a
| Versions      | all

Rephrases the section about supported setups in requirements to not be misleading.

### DIFF
--- a/docs/getting_started/requirements.md
+++ b/docs/getting_started/requirements.md
@@ -29,9 +29,9 @@ These setups are tested by QA and are generally recommended setups. For security
 |Filesystem|Linux ext4 / XFS|
 |Package manager|Composer|
 
-### Supported setups
+### Other supported setups
 
-Supported setups are those we perform automated testing on. For security and performance we generally recommend (unless otherwise noted and marked with \*) using the newer versions of components below.
+For security and performance we generally recommend (unless otherwise noted and marked with \*) using the newer versions of components below.
 
 -   OS: Linux
 -   Web Servers:


DominikaK:
Automated testing is not actually performed on the setups listed in the section, as per info from QA (ping @mnocon @m-tyrala, can you confirm?)

mnocon:
@DominikaK yes, the current statement could be misleading, it's partially correct. For example (unless I'm wrong and there are parts of the system where these tests are run):
- we perform automated tests on nginx (latest tagged Docker image), but currently not on Apache (our Docker stack uses nginx: https://github.com/ezsystems/ezplatform/blob/master/.env#L15)
- Both Kernel tests and Behat tests use MariaDB 10.3 (not earlier versions)
- Mysql 8 support (in tests) is something we're still working on (earlier in https://jira.ez.no/browse/EZP-29778 and now continued in https://github.com/ezsystems/ezpublish-kernel/pull/2583 )

But these setups should work, most of the time they contain just slightly older versions of software in recommended section 😉

andrerom:
As for the automated testing, it was the intention to somewhat cover at least min / max versions of supported platforms by separate nightly test runs _(not on travis slowing down every PR)_.

ref: https://doc.ez.no/display/DEV/Test+strategy and other discussions back at the time.

But indeed we don't fully do as it's written, so I'm ok with removing the mention.

DominikaK:
> +1, but as noted in Q&A yesterday this page is missing DEV / build requirement for node.js & yarn.

This will be added in a separate PR, as it only applies to 2.5.

PR #594, opened 2019-04-15
Added info about manuall Redis cache clearing

kmadejski:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | Related to CS-7429
| Versions      | `2.5`

After enabling `ezsystems/symfony-tools` the user might see a notice. To get rid of it, a manual Redis cache clearing is necessary.

### DIFF
--- a/docs/releases/updating_ez_platform.md
+++ b/docs/releases/updating_ez_platform.md
@@ -538,8 +538,7 @@ Some versions require updates to the database. Look through [the list of databas
 
     #### Required manual cache clearing if using Redis
 
-    If you are using Redis as your persistence cache storage and you see a notice message such as:
-    `In AbstractTagAwareAdapter.php line 71: Notice: Undefined index: value`, clear Redis instance manually after the upgrade.
+    If you are using Redis as your persistence cache storage you should always clear it manually after an upgrade.
     You can do it in two ways, by using `redis-cli` and executing the following command:
     
     ```bash


docs/releases/updating_ez_platform.md
```diff
@@ -536,6 +536,20 @@ Some versions require updates to the database. Look through [the list of databas
     bin/console ezplatform:migrate:legacy_matrix
     ```
 
+    #### Required manual cache clearing if using Redis
+
+    If you are using Redis as you persistence cache storage and you see a notice message like `In AbstractTagAwareAdapter.php line 71: Notice: Undefined index: value`, please clear Redis instance manually after the upgrade. You can do it in two ways, by using `redis-cli` and executing following command:
```
andrerom: Actually they should not wait for the notice, they MUST clear cache on upgrades in the first place.
But having note here saying that you'll get such notice, and non working cache if you don't is good so people may find the info when they forget.

kmadejski: Indeed, the point was to put the notice message in the docs, so it can be searchable I hope.

DominikaK:
```suggestion
    If you are using Redis as your persistence cache storage and you see a notice message such as:
    `In AbstractTagAwareAdapter.php line 71: Notice: Undefined index: value`, clear Redis instance manually after the upgrade.
    You can do it in two ways, by using `redis-cli` and executing the following command:
```

andrerom:
I'm wondering if it might be better to [get rid of the notice](https://github.com/ezsystems/symfony-tools/pull/7) and focus this on remembering to clear cache. Something like:
```
    If you are using Redis as your persistence cache storage you should always clear it's cache manually after an upgrade.
    You can do it in two ways, by using `redis-cli` and executing the following command:
```

kmadejski:
@andrerom yes, taking into account what you did in https://github.com/ezsystems/symfony-tools/pull/7 it's better to make this tip more generic as you proposed. I'll update it soon.

docs/releases/updating_ez_platform.md
```diff
@@ -536,6 +536,21 @@ Some versions require updates to the database. Look through [the list of databas
     bin/console ezplatform:migrate:legacy_matrix
     ```
 
+    #### Required manual cache clearing if using Redis
+
+    If you are using Redis as your persistence cache storage you should always clear it manually after an upgrade.
```
andrerom:
```suggestion
    If you are using Redis as your persistence cache storage you should always clear it manually during an upgrade.
```

Unsure, but seems more correct or?

PR #596, opened 2019-04-16
FT API rewrite

DominikaK:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | n/a
| Versions      | 2.x, master

Restructures the Field Type API section.

The content is redivided differently into pages to follow main topics.

Note: the content of the section is reviewed only partially.

Some todos for subsequent PRs:
- unify code examples to either base on one built-in FT (e.g. matrix) or to be generic
- verify the decriptions for all methods
- extend the section on Back Office templates with example templates
- add more doc about the available variables, like in the Template variables
- verify up-to-dateness of the Storage page


docs/api/field_type_form_and_template.md
```diff
@@ -0,0 +1,194 @@
...
+
+You can for example use `simple_block_field`, `simple_inline_field` or `field_attributes` blocks provided in [`content_fields.html.twig`](https://github.com/ezsystems/ezpublish-kernel/blob/v7.5.0/eZ/Bundle/EzPublishCoreBundle/Resources/views/content_fields.html.twig#L496).
+
+!!! caution
+
+    To be able to reuse built-in blocks, your template must inherit from `EzPublishCoreBundle::content_fields.html.twig`.
+
+### Registering your template
+
+To make your template available, you must register it in the system.
```
alongosz: With eZ Design mentioned above this is optional. So we can say something like
> If you don't want to use eZ Design or want to have separate templates per Field Type and/or SiteAccess, you can register a template by the following configuration:

docs/api/field_type_form_and_template.md
```diff
@@ -0,0 +1,194 @@
...
+You can use a separate template for previewing the Field in the Back Office by specifying:
+
+``` yaml
+ezpublish:
+    systems:
+        admin_group:
+            field_templates:
+                - { template: 'AcmeExampleBundle:adminui/field:custom_field_view.html.twig', priority: 10 }
+```
+
```
alongosz: > ... though it is recommended to use eZ Design with `ez_platform_standard_design.override_kernel_templates` set to `true`.
In this case instead of the above config, you can simply prepare `Resources/views/themes/admin/content_fields.html.twig` template without any  extra configuration.

docs/api/field_type_form_and_template.md
```diff
@@ -0,0 +1,194 @@
...
+### Field edit template
+
+To use a template for the Field edit form in the Back Office, you need to specify it in configuration
+under the `twig.form_themes` key:
+
+``` yaml
+twig:
+    form_themes:
+        - 'AcmeExampleBundle:adminui/field:custom_field_template.html.twig'
```
alongosz: I think I sneaked extension point for this in ezsystems/ezplatform-admin-ui#530. I need to confirm this is the same thing (I'm not that fluent in AdminUI voodoo ;) ).

docs/api/field_type_form_and_template.md
```diff
@@ -0,0 +1,194 @@
...
+```
+
+We encourage using custom form types for encapsulation as this makes templating easier by providing Twig block name.
+All built-in Field Types are implemented with this approach. In that case overriding form theme can be done with:
+
+``` html+twig
+{% block custom_fieldtype_widget %}
+    Hello world!
+    {{ block('form_widget') }}
```
alongosz: This calls for more doc about the available variables, the same like in the `Template variables` section. Someone from the Community asked about this, so it's not that obvious. I need to check this along with the previous comment. Some other devs help would be much appreciated.

docs/api/field_type_search.md
```diff
@@ -1,18 +1,17 @@
...
-`getIndexData( Field $field )`
+### `getIndexData( Field $field )`
 
 This method returns the actual index data for the provided `eZ\Publish\SPI\Persistence\Content\Field`. The index data consists of an array of `eZ\Publish\SPI\Persistence\Content\Search\Field` instances. They are described below in further detail.
 
-`getIndexDefinition()`
+### `getIndexDefinition()`
 
-To be able to query data properly an indexable Field Type also is required to return search specification. You must return a HashMap of `eZ\Publish\SPI\Persistence\Content\Search\FieldType` instances from this method, which could look like:
+To be able to query data properly an indexable Field Type also is required to return search specification. You must return an array of `eZ\Publish\SPI\Persistence\Content\Search\FieldType` instances from this method, which could look like:
```
alongosz: any particular reason for replacing HashMap with an array? HashMap is a specific case of an array, which is here relevant. More understandable term here would be "an associative array".

docs/api/field_type_api.md
```diff
@@ -1,37 +1,40 @@
 # Field Type API
 
-eZ Platform can support arbitrary data to be stored in the fields of a content item. In order to support custom data, besides the standard data types, a developer needs to create a custom Field Type.
+eZ Platform can support custom data to be stored in the Fields of a Content item.
+To do so, you need to create a custom Field Type.
 
-The implementation of a custom Field Type is done based on the **FieldType SPI** and its interfaces. These can be found under `eZ\Publish\SPI\FieldType`.
+A custom Field Type is implemented based on the **FieldType SPI** and its interfaces,
```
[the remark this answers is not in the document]
DominikaK: Can we do "A custom Field Type must implement the **FieldType Service Provider Interfaces**..."? Would that be OK subject-matter-wise?

docs/api/field_type_form_and_template.md
DominikaK: I'll add it to the "todo in future" list in PR description, so as not to bloat this PR any further

docs/api/field_type_api.md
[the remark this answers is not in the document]
alongosz: > Can we do "A custom Field Type must implement the **FieldType Service Provider Interfaces**..."? Would that be OK subject-matter-wise?

This is perfect :)

docs/api/field_type_type_and_value.md
```diff
@@ -0,0 +1,166 @@
...
+        'default' => false
+    ],
+    'defaultValue' => [
+        'type' => 'string',
+        'default' => 'Default Value'
+    ]
+];
+```
+
+The settings are mapped into Symfony forms via the [FormMapper](field_type_form_and_template.md#formmapper).
```
Nattfarinn: Ping @alongosz to clarify where and why do we need `getSettingsSchema`, because as far as I know these are not mapped to anything and not needed as long as FieldDefinitionFormMapper::mapFieldDefinitionForm provides correct Symfony Forms fields. Maybe deprecation candidate?

alongosz: > Ping @alongosz to clarify where and why do we need `getSettingsSchema`, because as far as I know these are not mapped to anything and not needed as long as `FieldDefinitionFormMapper::mapFieldDefinitionForm` provides correct Symfony Forms fields. Maybe deprecation candidate?

No idea. It requires deeper investigation.

docs/api/field_type_type_and_value.md
```diff
@@ -0,0 +1,166 @@
...
+
+It is important to note that the schema definitions of the Field Type can be both of arbitrary and serializable format, it is highly recommended to use a simple hash structure.
+
+!!! note
+
+    Since it is not possible to enforce a schema format, the code using a specific Field Type must basically know all Field Types it deals with.
+
+This will also apply to all user interfaces and the REST API, which therefore must provide extension points to register handling code for custom Field Type. These extensions are not defined yet.
+
+### Field Type name
```
adamwojs: I'm missing an explanation of what `eZ\Publish\SPI\FieldType\Nameable` is used for.

```
To be able to generate content name when field is part of name-schema or url-schema you have to implement `eZ\Publish\SPI\FieldType\Nameable` and register this service using the tag `ezpublish.fieldType.nameable`. 
```

Ref. https://github.com/ezsystems/ezpublish-kernel/blob/4eb8e38f020d97e7f8d8ff0e381db0e2b4141a1c/eZ/Publish/SPI/FieldType/Nameable.php

docs/api/field_type_testing.md
```diff
@@ -1,30 +1,36 @@
...
 
 !!! note
 
-    Note that the In-Memory stubs for the Public API integration test suite, do not perform actual Field Type calls, but mainly emulate the behavior of a Field Type for simplicity reasons. 
-    
-    If your Field Type needs to convert data between `storeFieldData()` and `getFieldData()`, you need to implement a `eZ\Publish\API\Repository\Tests\Stubs\PseudoExternalStorage` in addition, which performs this task. Running the tests against the Business Layer implementation of the Public API is not affected by this.
\ No newline at end of file
+    Note that the In-Memory stubs for the Public API integration test suite do not perform actual Field Type calls, but mainly emulate the behavior of a Field Type for the sake of simplicity.
+
+    If your Field Type needs to convert data between `storeFieldData()` and `getFieldData()`, you need to additionally implement an `eZ\Publish\API\Repository\Tests\Stubs\PseudoExternalStorage`, which performs this task. Running the tests against the business layer implementation of the Public API is not affected by this.
```
DominikaK: @adamwojs all remaining paths seem to be correct, with the exception of `eZ\Publish\API\Repository\Tests\Stubs\PseudoExternalStorage`, which does not seem to exist at all. Do you happen to have an idea what replaced it?

docs/api/field_type_form_and_template.md
```diff
@@ -0,0 +1,213 @@
...
+
+If you don't use [eZ Design Engine](../guide/design_engine.md) or want to have separate templates per Field Type and/or SiteAccess,
+you can register a template with the following configuration:
+
+``` yaml
+ezpublish:
+    system:
+        <siteaccess>:
+            field_templates:
+                -
```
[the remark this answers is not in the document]
DominikaK: This should be valid YAMl, see for example in our demo: https://github.com/ezsystems/ezplatform-demo/blob/master/app/config/field_templates.yml#L5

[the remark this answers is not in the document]
michalmilc90: Oh, WOW, I see now! 😄

docs/api/field_type_testing.md
```diff
@@ -1,30 +1,36 @@
...
-This type of integration test ensures, that a Field Type stores its data properly on basis of different Persistence SPI implementations.
+This type of integration test ensures that a Field Type stores its data properly on the basis of different Persistence SPI implementations.
 
 !!! note
 
-    By now, only the Legacy Storage implementation exists.
+    At the moment only Legacy Storage implementation is available.
 
-The integration tests with the Persistence SPI can be found in `eZ\Publish\SPI\Tests\FieldType`. In order to implement a test for your custom Field Type, you need to extend the common base class `eZ\Publish\SPI\Tests\FieldType\BaseIntegrationTest` and implement its abstract methods. As a reference the `KeywordIntegrationTest`, `UrlIntegrationTest` and `UserIntegrationTest` can deal.
+The integration tests with the Persistence SPI is in `eZ\Publish\SPI\Tests\FieldType`. In order to implement a test for your custom Field Type, you need to extend the common base class `eZ\Publish\SPI\Tests\FieldType\BaseIntegrationTest` and implement its abstract methods.
```
michalmilc90:
```suggestion
The integration tests with the Persistence SPI are in `eZ\Publish\SPI\Tests\FieldType`. In order to implement a test for your custom Field Type, you need to extend the common base class `eZ\Publish\SPI\Tests\FieldType\BaseIntegrationTest` and implement its abstract methods.
```

PR #598, opened 2019-04-17
Changed "select" by "click"

nicosomb:
Because we can select a role with the checkbox, it's confusing. I think it's better to say "click on".


michalmilc90:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | n/a
| Versions      | master, TBD

PR #603, opened 2019-04-24
Note about dumping translations after removing web/assets

DominikaK:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | n/a
| Versions      | 2.5, master

Note: not quite happy with the location of this info, as we don't normally cover cases like this, but this was requested in https://github.com/ezsystems/ezplatform/pull/394


emodric:
`encode` -> `encore` :)

PR #613, opened 2019-05-08
EZP-30391: Remove references to Elastic Search

DominikaK:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | https://jira.ez.no/browse/EZP-30391
| Versions      | all


andrerom:
+1, but I'd like us to remove it from earlier versions to, OR if that is not ok change from "experimental" to "deprecated POC" or similar.

docs/api/public_php_api_search.md
```diff
@@ -10,7 +10,7 @@ This section covers how the [`SearchService`](http://apidoc.ez.no/sami/trunk/NS/
 
 Query object contains two properties you can set criteria on `filter` and `query`. You can mix and match use or use both at the same time, there is one distinction between the two:
 
--   `query` Has an effect on scoring *(relevancy)* calculation, and also on the default sorting if `sortClause` is not specified, *when used with Solr and Elastic.* Typically `query` is used for `FullText` search criterion, otherwise you can place everything else on `filter`.
+-   `query` Has an effect on scoring *(relevancy)* calculation, and also on the default sorting if `sortClause` is not specified, *when used with Solr.* Typically `query` is used for `FullText` search criterion, otherwise you can place everything else on `filter`.
```
michalmilc90:
```suggestion
-   `query` — has an effect on scoring *(relevancy)* calculation and on the default sorting, if `sortClause` is not specified *when used with Solr.* Typically, `query` is used for `FullText` search criterion. Otherwise you can place everything else on `filter`.
```

alongosz:
> +1, but I'd like us to remove it from earlier versions too, OR if that is not ok change from "experimental" to "deprecated POC" or similar.
> 
> As in I'd like to avoid developers wasting time with it as they misunderstand and think it might be something they can use and that it might become stable over time _(that specific impl will never be stable which is why it was removed and should have long ago)_

I agree, we probably should not mention something which was never supported as it might be misleading. Especially that we totally dropped testing it even on 1.7.

PR #627, opened 2019-05-22
Update signals_reference.md

mateuszbieniek:
Added information to docs about methods in services that are deprecated + some minor additional changes.

Open question: there are signals that are no longer used anywhere except tests (`AddTranslationInfoSignal`, `SetPriorityOfObjectStateSignal`) shouldn't they be marked in some way in documentation? Now they may be confusing. 
Same for signals emitted only in REST calls (`ContentService\AddRelationSignal`, `ContentService\DeleteRelationSignal`, `ContentTypeService\AssignContentTypeGroupSignal`, `ContentTypeService\CopyContentTypeSignal`, `ContentTypeService\UnassignContentTypeGroupSignal`, `URLAliasService\CreateGlobalUrlAliasSignal`, `URLWildcardService\CreateSignal`, `URLWildcardService\RemoveSignal`, `UserService\AssignUserToUserGroupSignal`, `UserService\CreateUserGroupSignal`, `UserService\DeleteUserGroupSignal`, `UserService\MoveUserGroupSignal`, `UserService\UnAssignUserFromUserGroupSignal`, `UserService\UpdateUserGroupSignal`).

Note: I removed `ContentService::translateVersion` from `TranslateVersionSignal`, since `ContentService::translateVersion` is removed since kernel 7.0

### DIFF
--- a/docs/guide/signals_reference.md
+++ b/docs/guide/signals_reference.md
@@ -21,7 +21,7 @@ All Signals are relative to `eZ\Publish\Core\SignalSlot\Signal` namespace.
 |Signal type|Properties|Triggered by|
 |------|------|------|
 |`ContentService\AddRelationSignal`|`srcContentId` (source contentId, aka referrer)</br>`srcVersionNo`</br>`dstContentId` (destination contentId, aka target)|`ContentService::addRelation()`|
-|`ContentService\AddTranslationInfoSignal`|N/A|`ContentService::addTranslationInfo()`|
+|`ContentService\AddTranslationInfoSignal` (deprecated)|N/A||
 |`ContentService\CopyContentSignal`|`srcContentId` (original content ID)</br>`srcVersionNo`</br>`dstContentId` (contentId of the copy)</br>`dstVersionNo`</br>`dstParentLocationId` (locationId where the content has been copied)|`ContentService::copyContent()`|
 |`ContentService\CreateContentDraftSignal`|`contentId`</br>`versionNo`</br>`userId` (ID of User used to create the draft, or null - current User)|`ContentService::createContentDraft()`|
 |`ContentService\CreateContentSignal`|`contentId`</br>`versionNo`|`ContentService::createContent()`|


docs/guide/signals_reference.md
```diff
@@ -32,7 +32,7 @@ All Signals are relative to `eZ\Publish\Core\SignalSlot\Signal` namespace.
 |`ContentService\HideContentSignal`|`contentId`|`ContentService::hideContent()`|
 |`ContentService\PublishVersionSignal`|`contentId`</br>`versionNo`|`ContentService::publishVersion()`|
 |`ContentService\RevealContentSignal`|`contentId`|`ContentService::revealContent()`|
-|`ContentService\TranslateVersionSignal`|`contentId`</br>`versionNo`</br>`userId`|`ContentService::translationVersion()`|
+|`ContentService\TranslateVersionSignal`|`contentId`</br>`versionNo`</br>`userId`||
```
andrerom: signal should be deprecated (in 7.x) and removed in 8.x if corresponding API has been removed.

andrerom: To clarify: From what I can see `TranslateVersionSignal` is not used by kernel anymore.

So what I had in mind is that we should deprecate it in kernel in 7.5, remove it in master, and reflect the doc here for that _(mark as deprecated here in 2.5 and remove mention in master)_.

mateuszbieniek: @andrerom we should be fine with https://github.com/ezsystems/ezpublish-kernel/commit/8c46b66f3a91b3f03937f4a1448a5e02adea2ca9
Can you please re-review?

andrerom: Well we still need to add ` (deprecated)` here like done below  right?

andrerom: Also on AddTranslationInfoSignal while on it right?

mateuszbieniek: Done

PR #633, opened 2019-05-24
EZP-29301: Custom attributes in OE

DominikaK:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | https://jira.ez.no/browse/EZP-29301
| Versions      | 2.5.2, master

Adds a section about adding custom attributes and CSS classes to elements in the Online Editor.

Config examples based on https://github.com/ezsystems/ezplatform-richtext/pull/39

### DIFF
--- a/docs/guide/extending_online_editor.md
+++ b/docs/guide/extending_online_editor.md
@@ -276,12 +276,12 @@ The available elements are:
 
 !!! caution "Overriding embed templates"
 
-    If you override the default templates to `embedinline`, `embed` or `embedimage` elements,
+    If you override the default templates for `embedinline`, `embed` or `embedimage` elements,
     (e.g. `EzPublishCoreBundle:default:content/embed.html.twig`),
     the data attributes and classes will not be rendered automatically.
 
     Instead, you can make use of the `data_attributes` and `class` properties in your templates.
-    The `ez_data_attributes_serialize` helper enables you to serialize the datab attribute array.
+    The `ez_data_attributes_serialize` helper enables you to serialize the data attribute array.
 
 ### Custom data attributes
 


docs/guide/extending_online_editor.md
```diff
@@ -255,6 +255,85 @@ In the example above, the template files for the front end could be:
...
+                            custom-attribute:
+                                type: boolean
+                                default_value: false
+                            another-attribute:
+                                type: choice
+                                choices: [attr1, attr1]
+                                default_value: attr2
+                                required: false
+                                multiple: true
+```
```
michalmilc90: I'm very happy with the example you provide, but are we sure the developers will now how to configure the default values for `string` and `number` attribute type? :)

docs/guide/extending_online_editor.md
```diff
@@ -255,6 +255,85 @@ In the example above, the template files for the front end could be:
...
+
+### Custom attributes
+
+Custom attributes are configured under the `fieldtypes.ezrichtext.attributes` key.
+
+A custom attribute can belong to one of four types: `choice`, `boolean`, `string`, or `number`.
+You can also set each attribute to be `required` and set its `default_value`.
+
+For the `choice` type, you must provide an array of available `choices`.
+`multiple` enables you to choose whether more than one option can be selected.
```
michalmilc90:
```suggestion
Adding `multiple` enables you to choose whether more than one option can be selected.
```
Just to start a sentence with a capital letter, nothing more than that.

**Question:** Is `multiple` mandatory for `choices`? Meaning, the dev has to provide either `multiple: false` or `multiple: true`?

docs/guide/extending_online_editor.md
```diff
@@ -255,6 +255,85 @@ In the example above, the template files for the front end could be:
...
+                                default_value: false
+                            another-attribute:
+                                type: choice
+                                choices: [attr1, attr1]
+                                default_value: attr2
+                                required: false
+                                multiple: true
+```
+
+This configuration will be output as `data-<attribute_name>` in the corresponding HTML element,
```
michalmilc90:
```suggestion
This configuration will output `data-<attribute_name>` in the corresponding HTML element,
```

Wanted to simplify the sentence. Discard this suggestion if you think it is okay.

docs/guide/extending_online_editor.md
```diff
@@ -255,6 +255,85 @@ In the example above, the template files for the front end could be:
 
 Templates for Content View in the Back Office would be `app/Resources/views/themes/admin/field_type/ezrichtext/custom_style/highlighted_word.html.twig` and `app/Resources/views/themes/admin/field_type/ezrichtext/custom_style/highlighted_block.html.twig` respectively (assuming Admin SiteAccess uses the `admin` theme).
 
+## Custom attributes and classes
```
alongosz: let's use everywhere "data attributes", "attributes" is too generic (e.g class is also an attribute)

docs/guide/extending_online_editor.md
```diff
@@ -255,6 +255,85 @@ In the example above, the template files for the front end could be:
...
+- `ol`
+- `li`
+- `paragraph`
+- `table`
+- `tr`
+- `td`
+
+### Custom attributes
+
+Custom attributes are configured under the `fieldtypes.ezrichtext.attributes` key.
```
alongosz:
```suggestion
Custom attributes are configured under the `fieldtypes.ezrichtext.attributes` key for any given scope (SiteAccess).
```
something like ^ (might need grammar fixes :P)

docs/guide/extending_online_editor.md
```diff
@@ -255,6 +255,85 @@ In the example above, the template files for the front end could be:
...
+
+A custom attribute can belong to one of four types: `choice`, `boolean`, `string`, or `number`.
```
alongosz: Nitpick (if I'm right):
```suggestion
A custom attribute can belong to one of four types: `choice`, `boolean`, `string` or `number`.
```
Do we place comma before "or"?

docs/guide/extending_online_editor.md
alongosz: > **Question:** Is `multiple` mandatory for `choices`? Meaning, the dev has to provide either `multiple: false` or `multiple: true`?

It's not required - by default it's true.

docs/guide/extending_online_editor.md
```diff
@@ -255,6 +255,85 @@ In the example above, the template files for the front end could be:
...
+                                required: false
+                                multiple: true
+```
+
+This configuration will be output as `data-<attribute_name>` in the corresponding HTML element,
+in this example as `data-custom-attribute` and `data-another-attribute`.
+
+### Custom CSS classes
+
+Custom CSS classes are configured under the `fieldtypes.ezrichtext.classes` key.
```
alongosz:
```suggestion
Custom CSS classes are configured under the `fieldtypes.ezrichtext.classes` key for any given scope (SiteAccess).
```

docs/guide/extending_online_editor.md
```diff
@@ -255,6 +255,85 @@ In the example above, the template files for the front end could be:
...
+            fieldtypes:
+                ezrichtext:
+                    attributes:
+                        heading:
+                            custom-attribute:
+                                type: boolean
+                                default_value: false
+                            another-attribute:
+                                type: choice
+                                choices: [attr1, attr1]
```
alongosz: I'd recommend surrounding values with single quotes

docs/guide/extending_online_editor.md
```diff
@@ -255,6 +255,85 @@ In the example above, the template files for the front end could be:
...
+                                default_value: attr2
```
alongosz: this one also
```suggestion
                                default_value: 'attr2'
```
and similar occurrences

docs/guide/extending_online_editor.md
```diff
@@ -255,6 +255,85 @@ In the example above, the template files for the front end could be:
...
+``` yaml
+ezpublish:
+    system:
+        # The configuration only works with an admin (Back Office) SiteAccess
+        <siteaccess>:
+            fieldtypes:
+                ezrichtext:
+                    attributes:
+                        heading:
+                            custom-attribute:
```
alongosz: to be safe these should also be single-quoted

docs/guide/extending_online_editor.md
DominikaK: We can (and want to 😉 ), ref. https://en.wikipedia.org/wiki/Serial_comma

docs/guide/extending_online_editor.md
DominikaK: We've recently decided to unify the convention for code examples and go the opposite way (as in https://github.com/ezsystems/developer-documentation/pull/609)

docs/guide/extending_online_editor.md
DominikaK: I think we can assume this level of fluency with using YAML in our audience.

docs/guide/extending_online_editor.md
alongosz: :scream: good to know :ok_hand:

docs/guide/extending_online_editor.md
```diff
@@ -255,6 +255,99 @@ In the example above, the template files for the front end could be:
...
+### Custom data attributes
+
+Custom data attributes are configured under the `fieldtypes.ezrichtext.attributes` key.
+The configuration is SiteAccess-aware.
+
+A custom data attribute can belong to one of four types: `choice`, `boolean`, `string`, or `number`.
+You can also set each attribute to be `required` and set its `default_value`.
+
+For the `choice` type, you must provide an array of available `choices`.
+Adding `multiple` enables you to choose whether more than one option can be selected.
```
alongosz: Note: `multiple` option proved to be confusing, so right now it's by default `false`, but only for data attributes. For classes it remains `true` by default.

docs/guide/extending_online_editor.md
```diff
@@ -255,6 +255,99 @@ In the example above, the template files for the front end could be:
...
+                            another-attribute:
+                                type: choice
+                                choices: [attr1, attr1]
+                                default_value: attr2
+                                required: false
+                                multiple: true
+```
+
+This configuration will output `data-<attribute_name>` in the corresponding HTML element,
+in this example as `data-custom-attribute` and `data-another-attribute`.
```
alongosz: Actually, I've completely missed this.
For a paragraph if e.g. `value1` was chosen for `my-attribute` custom data attribute, a HTML output would be:
```html
<p data-ezattribute-my-attribute="value1"></p>
```

It might be worth to mention that in case of data attributes, if someone allowed multiple options, e.g. `value1` and `value2` to be chosen, the the rendered HTML for e.g. `my-attribute` would be e.g.:
```html
<p data-ezattribute-my-attribute="value1,value2"></p>
```
It's comma-separated on purpose and need to be handled (e.g split into values on front-end) by a developer if he allowed multiple choices.

PR #634, opened 2019-05-27
[EZP-25721] Warn devs upfront of ConfigResolver initialization order issues

michalmilc90:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | [EZP-25721](https://jira.ez.no/browse/EZP-25721)
| Versions      | 1.7.10, 1.13.6, 2.5.2

### DIFF
--- a/docs/guide/config_dynamic.md
+++ b/docs/guide/config_dynamic.md
@@ -60,8 +60,8 @@ Both `getParameter()` and `hasParameter()` can take three arguments:
     In debug mode, ConfigResolver detects if parameters were loaded prior to initialization of SiteAccess to warn about issues.
     It will log all instances of `getParameter()` that may be used unsafely.
     
-    If a problem occurs, you will reveive a warning that ConfigResolver had been used to load parameter `languages`
-    before SiteAccess was loaded by services: `my.own.service` and `ez.service.used.to.early`.
+    If a problem occurs, you will receive a warning that ConfigResolver had been used to load parameter `languages`
+    before SiteAccess was loaded by services: `my.own.service` and `ez.service.used.too.early`.
        
     To avoid such issues:
     


docs/guide/config_dynamic.md
```diff
@@ -44,6 +44,16 @@ $myParamSettingAdmin = $configResolver->getParameter( 'my_param', 'myapp', 'admi
 // $myParamSetting's value will be 'another value'
 ```
 
+In debug mode, ConfigResolver detects if parameters were loaded prior to initialization of SiteAccess to warn about issues that are hard to resolve.
+It will log all instances of `getParameter()` that may be used unsafely.
+
+Additionally, for class-based services, the warning appears even if a service is recognized by `DynamicSettingsListener` as `$updatableServices` (which is safe).
```
andrerom: I think we can skip this, it’s implied in warning on usages deemed unsafe

docs/guide/config_dynamic.md
```diff
@@ -44,6 +44,16 @@ $myParamSettingAdmin = $configResolver->getParameter( 'my_param', 'myapp', 'admi
...
+
+To avoid issues:
+- Do not use ConfigResolver in service factory.
```
andrerom: Hmm, this can be misleading.
Where did you take this from?

docs/guide/config_dynamic.md
```diff
@@ -44,6 +44,16 @@ $myParamSettingAdmin = $configResolver->getParameter( 'my_param', 'myapp', 'admi
...
+- Instead of using `ctor('$dynamic_param$')`, use `(setter('$dynamic_param$'))` as it allows the system to update the scope changes or use ConfigResolver.
+- You can also use ConfigResolver lazily by [making commands lazy](https://symfony.com/doc/current/console/lazy_commands.html) or configure [lazy services](https://symfony.com/doc/current/service_container/lazy_services.html).
```
andrerom: You can skip configresolver here, it’s not that relevant as it might be indirect usage (you might depend on a regular service, which in turn depends on configresolver)

So somehow state you can avoid the issue by either making command lazy (2.5+ only) or service lazy (does not apply to commands)

docs/guide/config_dynamic.md
```diff
@@ -44,6 +44,16 @@ $myParamSettingAdmin = $configResolver->getParameter( 'my_param', 'myapp', 'admi
...
 ```
 
+In debug mode, ConfigResolver detects if parameters were loaded prior to initialization of SiteAccess to warn about issues that are hard to resolve.
+It will log all instances of `getParameter()` that may be used unsafely.
+
+Additionally, for class-based services, the warning appears even if a service is recognized by `DynamicSettingsListener` as `$updatableServices` (which is safe).
+
+To avoid issues:
+- Do not use ConfigResolver in service factory.
+- Instead of using `ctor('$dynamic_param$')`, use `(setter('$dynamic_param$'))` as it allows the system to update the scope changes or use ConfigResolver.
```
andrerom: “Or use config resolver”
This can be own point about rather injecting config resolver and only using the value needed when necessary, aka load the parameter lazily

docs/guide/config_dynamic.md
michalmilc90: I picked it from [this comment](https://github.com/ezsystems/ezpublish-kernel/pull/2636/files#diff-26112d3c76ce033446c6bf175dbdae01R323), but I may have misinterpreted something. Your suggestion would be to delete this line?

andrerom: So maybe then `Try to avoid eager usage of config resolver (e.g. in service factories)`

docs/guide/config_dynamic.md
```diff
@@ -44,6 +44,16 @@ $myParamSettingAdmin = $configResolver->getParameter( 'my_param', 'myapp', 'admi
 // $myParamSetting's value will be 'another value'
 ```
 
+In debug mode, ConfigResolver detects if parameters were loaded prior to initialization of SiteAccess to warn about issues that are hard to resolve.
+It will log all instances of `getParameter()` that may be used unsafely.
+
+To avoid issues:
+- Avoid eager usage of config resolver (e.g. in service factories).
+- Instead of using `ctor('$dynamic_param$')`, use `(setter('$dynamic_param$'))` as it allows the system to update the scope changes.
```
andrerom:
```suggestion
- Instead of using `ctor('$dynamic_param$')`, use `(setter('$dynamic_param$'))` as it allows the system to update your service with changes on scope changes.
```

docs/guide/config_dynamic.md
```diff
@@ -44,6 +44,16 @@ $myParamSettingAdmin = $configResolver->getParameter( 'my_param', 'myapp', 'admi
...
+- Load the parameter lazily by injecting ConfigResolver with the required value only when necessary.
+- Try using [lazy commands](https://symfony.com/doc/current/console/lazy_commands.html) (applicable only to eZ Platform v2.5 or higher).
```
andrerom:
```suggestion
- Try using [lazy commands](https://symfony.com/doc/current/console/lazy_commands.html) (applicable only to Symfony v3.4 or higher).
```


Unsure about this one, but more precise at least. Up to you.

docs/guide/config_dynamic.md
```diff
@@ -44,6 +44,16 @@ $myParamSettingAdmin = $configResolver->getParameter( 'my_param', 'myapp', 'admi
 // $myParamSetting's value will be 'another value'
 ```
 
+In debug mode, ConfigResolver detects if parameters were loaded prior to initialization of SiteAccess to warn about issues that are hard to resolve.
+It will log all instances of `getParameter()` that may be used unsafely.
+
+To avoid issues:
+- Avoid eager usage of config resolver (e.g. in service factories).
+- Instead of using `ctor('$dynamic_param$')`, use `(setter('$dynamic_param$'))` as it allows the system to update the scope changes.
+- Load the parameter lazily by injecting ConfigResolver with the required value only when necessary.
```
andrerom:
```suggestion
- Load the parameter lazily by injecting ConfigResolver, and get the parameter from it _when_ you need to instead of during construction.
```

andrerom:
Ah, forgot. Maybe we can introduce the section with example from the actual warning so it's possible to find this by searching for parts of it?

e.g.: `WARNING   [app] ConfigResolver was used to load parameter "languages" before SiteAccess was loaded by services: "my.own.service", "ezpublish.service.used.to.early". This can cause issues. Try to use ConfigResolver lazily, make commands that rely on them lazy, or try to mark the service as lazy.`

andrerom:
👌

docs/guide/config_dynamic.md
```diff
@@ -54,6 +54,24 @@ Both `getParameter()` and `hasParameter()` can take three arguments:
 2. `$namespace` - your application namespace, `myapp` in the previous example. If null, the default namespace will be used, which is `ezsettings` by default.
 3. `$scope` - a SiteAccess name. If null, the current SiteAccess will be used.
 
+### ConfigResolver unsafe parameters
```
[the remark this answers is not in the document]
michalmilc90: That's a great idea!

docs/guide/config_dynamic.md
```diff
@@ -54,6 +54,23 @@ Both `getParameter()` and `hasParameter()` can take three arguments:
 2. `$namespace` - your application namespace, `myapp` in the previous example. If null, the default namespace will be used, which is `ezsettings` by default.
 3. `$scope` - a SiteAccess name. If null, the current SiteAccess will be used.
 
+
+!!! note
+
+    In debug mode, ConfigResolver detects if parameters were loaded prior to initialization of SiteAccess to warn about issues.
+    It will log all instances of `getParameter()` that may be used unsafely.
+    
+    If a problem occurs, you will reveive a warning that ConfigResolver had been used to load parameter `languages`
```
DominikaK:
```suggestion
    If a problem occurs, you will receive a warning that ConfigResolver had been used to load parameter `languages`
```

docs/guide/config_dynamic.md
```diff
@@ -54,6 +54,23 @@ Both `getParameter()` and `hasParameter()` can take three arguments:
...
+    before SiteAccess was loaded by services: `my.own.service` and `ez.service.used.to.early`.
```
DominikaK:
```suggestion
    before SiteAccess was loaded by services: `my.own.service` and `ez.service.used.too.early`.
```

PR #641, opened 2019-06-05
Update Page Builder Cache TTL doc

andrerom:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | N/A
| Versions      | 2.5+

This needs review, but attempts to update the example to cover Page Builder and not Landing Page (1.x).

### DIFF
--- a/docs/guide/http_cache.md
+++ b/docs/guide/http_cache.md
@@ -342,10 +342,9 @@ ezpublish:
     However, if the given block value has a since / till date,
     this will be taken into account for the TTL calculation for the block and also for the whole page.
 
-    To overload this behavior, listen to [BlockResponseEvents::BLOCK_RESPONSE](extending_page/#block-render-response).
-    For instance, to disable cache for a custom block you can set a listener's priority to `100`,
-    match for the given block type and call `$event->getResponse()->setPrivate()`.
-    If you only want to adjust the default logic, you can set priority to `-200` and for instance set another TTL.
+    To overload this behavior, listen to [BlockResponseEvents::BLOCK_RESPONSE](extending_page/#block-render-response),
+    and set prioroty to for instance `-200` to adapt what Page Field type does by default. E.g. in order to disable cache
+    for the block use `$event->getResponse()->setPrivate()`.
 
 !!! note "Invalidating Varnish cache using tokens"
 


docs/guide/http_cache.md
```diff
@@ -338,22 +338,14 @@ ezpublish:
...
-    `block_type` should be replaced with the actual block name, e.g. `embed`, `collection`, `schedule`, etc.
-    In the example above `block_type` will be cached for 10 minutes.
-
-    By default blocks are not cached (TTL = 0) for backwards compatibility reasons.
+    Block cache by default respects `$content.ttl_cache$` and `$content.default_ttl$` settings.
+    However, if the given block value has a since / till date,
+    this will be taken into account for the TTL calculation for the block and also for the whole page.
+
+    To overload this behavior, listen to [BlockResponseEvents::BLOCK_RESPONSE](extending_page/#block-render-response).
+    For instance, to disable cache for a custom block you can set a priority of 100,
```
kmadejski:
```suggestion
    For instance, to disable cache for a custom block you can set a listener's priority to `100`,
```

kmadejski: Maybe I'm wrong, but besides suggestion, I'm wondering whether the priority should be `-100`, perhaps? AFAIR the `\EzSystems\EzPlatformPageFieldType\Event\Subscriber\BlockVisibilityTtlHeadersResponseSubscriber` has a priority of `-100`, so if the developer's own listener has priority of `100` and sets different cache settings for the given response, these will be overridden afterward by the PB listeners, right?

andrerom: I might have messed up the order. But the idea here was to run the listener before eZ's if you want to disable cache, and after (ref -200 below) if just want to adjust the headers that gets set by default

andrerom: @kmadejski WDYT? Or should we get someone else to review also maybe to see if this can be simplified?

kmadejski: As for me, the current description is totally fine, I didn't find it too complex. The only doubt is about these priorities. I briefly checked it and it seems that indeed the priority has to be at least `-200` to be able to modify `s-maxage`/`maxage`, I believe it's the same for `public`/`private`.

andrerom: Better?

DominikaK:
@andrerom do you need QA for this? Is there a related ticket?

andrerom:
Not really, but maybe place your laptop in front of @webhdx to get a nay or yey on the change so he is aware of what he might break when he changes it again ;)

PR #649, opened 2019-06-18
Typo :  directed acycling graph

reithor:
Additionaly I am not sure, if the statement 'This example *only* works if your workflow is a [directed acycling graph]' is true.
article_workflow in ee-demo seems to have a directed cycle:
draft -> to_design -> design -> back_to_draft -> draft
https://github.com/ezsystems/ezplatform-ee-demo/blob/master/app/config/workflows.yml

| Question      | Answer
| ------------- | ---
| JIRA Ticket   | 
| Versions      |

### DIFF
--- a/docs/guide/workflow.md
+++ b/docs/guide/workflow.md
@@ -148,11 +148,6 @@
 
     You can use the [`PublishOnLastStageSubscriber.php`](https://github.com/ezsystems/ezplatform-ee-demo/blob/v2.5.0/src/AppBundle/Event/Workflow/PublishOnLastStageSubscriber.php) from eZ Platform demo as a basis for the subscriber.
 
-    !!! note
-
-        This example only works if your workflow is a [directed acyclic graph](https://en.wikipedia.org/wiki/Directed_acyclic_graph)
-        (it doesn't have directed cycles).
-
     The subscriber listens for the `WorkflowEvents::WORKFLOW_STAGE_CHANGE` event
     ([line 61](https://github.com/ezsystems/ezplatform-ee-demo/blob/v2.5.0/src/AppBundle/Event/Workflow/PublishOnLastStageSubscriber.php#L61)).
     When the event occurs, it publishes the relevant Content item


DominikaK:
@damianz5, the example was based on your work, can you comment?

damianz5:
@reithor agree about the acyclic part, this could be linked to https://en.wikipedia.org/wiki/Directed_graph

docs/guide/workflow.md
```diff
@@ -150,7 +150,7 @@
 
     !!! note
 
-        This example only works if your workflow is a [direct acycling graph](https://en.wikipedia.org/wiki/Directed_acyclic_graph)
+        This example only works if your workflow is a [directed acyclic graph](https://en.wikipedia.org/wiki/Directed_acyclic_graph)
```
DominikaK:
```suggestion
        This example only works if your workflow is a [directed graph](https://en.wikipedia.org/wiki/Directed_graph).
```

docs/guide/workflow.md
```diff
@@ -150,7 +150,7 @@
...
         (it doesn't have directed cycles).
```
DominikaK:
```suggestion
```

docs/guide/workflow.md
reithor: Hm,
 "_This example only works if your workflow is a [directed graph]_"
seems to be unnecessesary / redundant in a way - as workflows are supposed to be (or represented by) directed graphs in general.

DominikaK: @reithor, so do you think it would be better to remove the note altogether? Will it be obvious to the reader?

reithor: @DominikaK : ... missed your message, sorry.

'_This example only works if your workflow is a [direct acycling graph]_'
--> has wrong typo and is wrong
'_This example only works if your workflow is a [directed acyclic graph]_'
--> is wrong
'_This example only works if your workflow is a [directed graph]_'
--> gives no additional information

So yes, I would remove the note.

PR #667, opened 2019-07-01
Fix the parameter name to be able to override.

Plopix:
Thanks @maximstrukov

### DIFF
--- a/docs/guide/user_management.md
+++ b/docs/guide/user_management.md
@@ -85,30 +85,30 @@ You can also modify the following form templates:
 **Changing user password:**
 
 ``` yaml
-ezpublish.system.<siteaccess>.user_change_password.templates.form
-ezpublish.system.<siteaccess>.user_change_password.templates.success
+ezsettings.<siteaccess>.user_change_password.templates.form
+ezsettings.<siteaccess>.user_change_password.templates.success
 ```
 
 **Password recovery forms:**
 
 ``` yaml
-ezpublish.system.<siteaccess>.user_forgot_password.templates.form
-ezpublish.system.<siteaccess>.user_forgot_password_success.templates.form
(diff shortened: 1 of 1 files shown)


DominikaK:
@Plopix, you are correct that currently the parameteres are not provided in practical way, but we should actually change them to be displayed in full (tabbed) yaml format, like we do with other configuration. We'll do that in a separate PR. Thanks for pointing this out.

Plopix:
and we don't warn all the users seeing this up to 2.5.1? we lost half a day on that as it was not obvious at all. Thx!

DominikaK:
@Plopix, here's our proposal, formatted to be easier to use directly in code: https://github.com/ezsystems/developer-documentation/pull/670

Plopix:
is this working? I don't think those is a semantic configuration.

adamwojs:
Hi @Plopix!

Sorry for this inconvenience. 

> is this working? I don't think those is a semantic configuration.

"It's complicated" The semantic configuration is available for:

* Change password  (https://github.com/ezsystems/ezplatform-user/blob/master/src/bundle/DependencyInjection/Configuration/Parser/ChangePassword.php#L24-L40)
* User registration (https://github.com/ezsystems/ezplatform-user/blob/master/src/bundle/DependencyInjection/Configuration/Parser/UserRegistration.php#L32-L42)

Update Form for User Settings is configured using Views: https://github.com/ezsystems/ezplatform-user/pull/21

The semantic configuration is missing for User Settings List, Forgot and Reset Password. 

IMHO something worth to fix in 3.0 in one way or another:
A) providing semantic configuration 
B) using ezdesign 

// cc @mikadamczyk @ViniTou

Plopix:
this is probably why for the time being this pr is valid waiting for the semantic configuration to be done.

just my opinion ;)

PR #673, opened 2019-07-04
Update text around Elastic

andrerom:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | N/A
| Versions      | 3.0/master

Just expand on the message to:
1. make it clear it was never supported, and was limited to 1.4.2
2. Reassure people that this is still on our roadmap for a later version

### DIFF
--- a/docs/releases/ez_platform_v3.0_deprecations.md
+++ b/docs/releases/ez_platform_v3.0_deprecations.md
@@ -127,7 +127,8 @@ Deprecated `ezprice` and `ezpage` Field Types have been removed.
 
 ### Elastic Search
 
-Experimental, deprecated and unsupported code for Elastic Search 1.4.2 has been dropped from kernel.
+Experimental, deprecated and unsupported code for Elastic Search 1.4.2 has been dropped from kernel,
+to be replaced with a dedicated bundle for the latest Elastic version in the future.
 
 ### Assetic support
 


docs/releases/ez_platform_v3.0_deprecations.md
```diff
@@ -127,7 +127,7 @@ Deprecated `ezprice` and `ezpage` Field Types have been removed.
 
 ### Elastic Search
 
-Elastic Search support has been dropped.
+Experimental, deprecated & non-supported code for Elastic Search 1.4.2 has been dropped from kernel, a dedicated bundle for latests Elastic version is on the roadmap for later 3.x version.
```
DominikaK: As usual I'm leery of promising what *will* happen in doc.
```suggestion
Experimental, deprecated and unsupported code for Elastic Search 1.4.2 has been dropped from kernel.
```

DominikaK: ...or at least:
```suggestion
Experimental, deprecated and unsupported code for Elastic Search 1.4.2 has been dropped from kernel,
to be replaced with a dedicated bundle for the latest Elastic version in future.  
```

andrerom:
@DominikaK took your second proposal, added "the" when referring to the future. Remove it if it does not belong there.

PR #674, opened 2019-07-04
3.0 field type deprecations clarifications

andrerom:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | N/A
| Versions      | 3.0/master

Make it more clear what has been changed for getName(), and also reassure people that RichText is not gone, just moved.


andrerom:
Small typo, but looks ok 👍 now.

PR #681, opened 2019-07-09
Allow Random sort (Solr)

michalmilc90:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | [EZP-28898](https://jira.ez.no/browse/EZP-28898)
| Versions      | master


docs/guide/search.md
```diff
@@ -342,6 +342,7 @@ The list below presents the Sort Clauses available in the `eZ\Publish\API\Repos
 |`DatePublished`|`?sortDirection`|
 |`Field`|`typeIdentifier` as string</br>`fieldIdentifier` as string</br> `?sortDirection`</br>`?languageCode` as string|
 |`MapLocationDistance `|`typeIdentifier` as string</br>`fieldIdentifier` as string</br>`latitude` as float</br>`longitude` as float</br>`?sortDirection`</br>`?languageCode` as string|
+|`Random`|`?sortDirection`|
```
ViniTou: Random use two arguments as constructor, first is int|null seed and second is sortDirection.

As a side note, seed is not implemented in every storage we support. As for now it is working in SOLR and MySql.

PR #685, opened 2019-07-12
Update EE tutorial script to v3

DominikaK:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | 
| Versions      | 

Updates the tutorial script with changes from https://github.com/ezsystems/developer-documentation/pull/675

To we verified after https://github.com/ezsystems/ezplatform-ee-beginner-tutorial/pull/14 is merged

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -10,7 +10,7 @@
     },
     "require": {
         "behat/behat": "^3.4.3",
-        "ezsystems/repository-forms": "~3.0.0@beta",
+        "ezsystems/repository-forms": "^3.0@dev",
         "symfony/property-access": "^4.3"
     },
     "require-dev": {


tests/src/Context/ConfigurationContext.php
```diff
@@ -165,4 +151,12 @@ private function clearCache(): void
 
         $application->run($input);
     }
+
+    /**
+     * @Given I rebuild Webpack Encore assets
+     */
+    public function rebuildYarn(): void
+    {
+        shell_exec("yarn encore dev");
```
DominikaK: Idea: is it possible to get Symfony env here (with `$APP_ENV` or something similar)?

mnocon: It should be possible to use `ezplatform:encore:compile` here (from https://github.com/ezsystems/ezplatform/blob/master/composer.json#L110)

It will rely on APP_ENV (https://github.com/ezsystems/ezplatform-core/pull/7#issuecomment-484425523)

PR #691, opened 2019-07-17
EZP-30763: Updated UIGuidelines component Tables

inakijv:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | [EZP-30763](https://jira.ez.no/browse/EZP-30763)
| Versions      | 2.5

Tables - eZ Platform Developer Documentation (1)
Tables - eZ Platform Developer Documentation (2)

### DIFF
--- a/docs/guidelines/components/tables.md
+++ b/docs/guidelines/components/tables.md
@@ -95,19 +95,23 @@ code_example]]
             <th>Version</th>
             <th>Modified language</th>
             <th>Contributor</th>
+            <th>Created</th>
+            <th>Last saved</th>
             <th></th>
         </tr>
     </thead>
     <tbody>
         <tr>
-            <td class="ez-checkbox-cell">
+            <td class="ez-table__cell ez-table__cell--has-checkbox">
                 <input type="checkbox">
             </td>
-            <td>1</td>
-            <td>English (United Kingdom)</td>
-            <td>Administrator User</td>
-            <td>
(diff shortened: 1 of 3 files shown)


inakijv:
ping @dew326 @tischsoic

docs/guidelines/components/tables.md
```diff
@@ -119,6 +123,34 @@ code_example]]
...
+    <thead>
+        <tr>
+            <th>Name</th>
+            <th>Limitations</th>
+            <th>Stages</th>
+        </tr>
+    </thead>
+    <tbody>
+        <tr>
+            <td class="ez-table__cell--top"><a href="#">Article Workflow</a></td>
```
dew326: I don't think we have class `ez-table__cell--top` in our CSSes.

inakijv: True, it's part of `ezplatform-workflow`. Should we consider adding it or we better wait?

dew326: `workflow` is a EE feature, this might be confusing for the open-source users that this class doesn't work. But on the other hand, we do not have documentation for the EE users.

inakijv: Updated based on discussion 👍

inakijv:
Updated PR based on discussion:
UI Guidelines cannot include (nor create) special documentation for the EE users (we could add specific labels, but this option needs further discussion).

PR #741, opened 2019-08-28
CS-7683 EZ on debian and Sorl 6 Java downgrade

MagdalenaZuba:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | https://jsd.ez.no/browse/CS-7683
| Versions      | all

### DIFF
--- a/docs/getting_started/requirements.md
+++ b/docs/getting_started/requirements.md
@@ -27,18 +27,12 @@ These setups are tested by QA and are generally recommended setups. For security
 
 |||
 |------|------|
-|Search|Solr (recommended; for performance, features and search quality):</br></br>Solr 6.x *Currently tested with Solr 6.6LTS*</br></br>Oracle Java/Open JDK: 8 |
+|Search|Solr (recommended; for performance, features and search quality):</br></br>Solr 6.x *Currently tested with Solr 6.6LTS*</br></br>Oracle Java/Open JDK: 8</br></br>*NOTE: If your Java version is higher than 8, you need to downgrade it to Oracle Java/Open JDK: 8.* |
 |Graphic Handler|GraphicsMagick or ImageMagick or GD|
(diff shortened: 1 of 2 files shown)


docs/getting_started/requirements.md
```diff
@@ -34,6 +34,12 @@ These setups are tested by QA and are generally recommended setups. For security
 |Package manager|Composer (recent stable version)|
 |Asset manager|`Node.js` 10.15.3 LTS</br>`yarn` 1.15.2 or higher|
 
+!!! caution "Java version"
+
+    If your Java version is higher than 8, you need to downgrade it to Oracle Java/Open JDK: 8. 
+    You can do it by following your favorite tutorial, for example:
+    [How to downgrade Java on Ubuntu/Kali Linux.](https://flicsdb.com/how-to-downgrade-java-on-ubuntu-kali-linux/)
```
adamwojs: Is there any official documentation for debian/red hat/oracle about installing concrete java version?

MagdalenaZuba: I could find only this: https://developers.redhat.com/blog/2018/12/10/install-java-rhel8/ or we can skip a link altogether.

mateuszbieniek: Maybe it is just me, but can't we find a better spot/less confusing way to mention it? Right now it just "caution" under the requirements table - if I did not pay attention reading the whole doc I could get a wrong idea that Java is required for Platform to work and I have to install it in version 8, which is not true. 
But as I said, it may be just me...

MagdalenaZuba: It is the only place we mention Java in the documentation, but if it's confusing in this format, we can move it inside the table as a plain text under `Oracle Java/Open JDK: 8` requirement.
The second note is in 'Download and configure solr' section.

PR #754, opened 2019-09-05
Add caution about $_POST PHP ini limit while editing large CTs

konradoboza:
Simple note about possible memory-related issues when working with large Content Types.

### DIFF
--- a/docs/guide/content_model.md
+++ b/docs/guide/content_model.md
@@ -127,7 +127,15 @@ They determine what Fields of what Field Types will be included in all Content i
 
 !!! caution
     
-    In case of Content Types containing many Field Types you should be aware of possible memory-related issues with publishing/editing. They are caused by the limitation of how many `$_POST` input variables can be accepted. The easiest way to fix them is by increasing the `max_input_vars` value in the `php.ini` configuration file. You may also experience performance problems with such large Content Types, in particular when you have many Content items.
+    In case of Content Types containing many Field Types you should be aware of possible memory-related issues with publishing/editing.
+    They are caused by the limitation of how many `$_POST` input variables can be accepted.
+    
+    The easiest way to fix them is by increasing the `max_input_vars` value in the `php.ini` configuration file.
+    Note that this solution is not universally recommended and you're proceeding on your own risk.
+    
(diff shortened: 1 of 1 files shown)


docs/guide/content_model.md
```diff
@@ -125,6 +125,10 @@ They determine what Fields of what Field Types will be included in all Content i
 
     You can assign each Field defined in a Content Type to a group by selecting one of the groups in the Category drop-down. [Available groups can be configured in the content repository](configuration.md#content-repository-configuration).
 
+!!! caution
+    
+    In case of Content Types containing many Field Types you should be aware of possible memory-related issues on publishing/editing. They are caused by the limitation of how many `$_POST` input variables may be accepted. The easiest way to fix them is increasing `max_input_vars` value in `php.ini` configuration file.
```
glye:
```suggestion
    In case of Content Types containing many Field Types you should be aware of possible memory-related issues on publishing/editing. They are caused by the limitation of how many `$_POST` input variables may be accepted. The easiest way to fix them is increasing `max_input_vars` value in `php.ini` configuration file. You may also experience performance problems with such large Content Types, in particular when you have many Content items.
```
Should we define a threshold, beyond which you should consider doing things differently? 20? 30?

michalmilc90:
```suggestion
In case of Content Types containing many Field Types you should be aware of possible memory-related issues with publishing/editing. They are caused by the limitation of how many `$_POST` input variables can be accepted. The easiest way to fix them is by increasing the `max_input_vars` value in the `php.ini` configuration file.
```

konradoboza: Not sure about this one, haven't tested any number in particular. :/

konradoboza:
@glye added suggestion regarding performance drop when multiple Content items using large CT.

docs/guide/content_model.md
glye: Yeah, I didn't mean a scientifically valdi limit, just a fencepost-ish thing, saying something like "beyond this you are entering uncharted territory, we don't intend our software to be used like this and haven't tested it, perhaps you should consider other ways of doing what you need to do".
CC @ezsystems/documentation-team

PR #761, opened 2019-09-11
JavaScript client removal

MagdalenaZuba:
| Question      | Answer
| ------------- | ---
| Versions      | 2.5 and master

JavaScript client was removed in v2.2.


michalmilc90:
Might be a silly question, but do we have it covered in the release notes?

PR #764, opened 2019-09-12
Cookbook cleanup: Relocating 'Display content using simple templates'

michalmilc90:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | n/a
| Versions      | all

In this PR, 'Display content using simple templates' is merged with 'templates.md'


docs/guide/displaying_content_with_simple_templates.md
```diff
@@ -0,0 +1,96 @@
+# Content item rendering
```
DominikaK: The title is a bit too close to "Content rendering" we have above, I'm thinking something line "Templating basics"?
(You could also change file name, as links will be broken here anyway)

docs/guide/templates.md
```diff
@@ -70,13 +72,24 @@ This is what individual keys in the configuration mean:
...
+Each Content item can be rendered differently, using different templates, depending on the type of view it is displayed in.
+The default, built-in views are:
+ 
+- **full** – used when the Content item is displayed by itself, as a full page
+- **line** – used when it is displayed as an item in the list, for example a listing of contents of a folder
+- **embed** – used when one Content item is embedded in another, as a block
+- **embed-inline** – used when a Content item is embedded inline in another block 
 
-See [View provider configuration](content_rendering#configuring-views-the-viewprovider) for more details.
+Other, custom view types can be created, used for example for [embedding one Content item in another](templates.md#embedding-content-items), but only these four have built-in controllers in the system.
```
MagdalenaZuba:
```suggestion
Other, custom view types can be created, used for example for [embedding one Content item in another](#embedding-content-items), but only these four have built-in controllers in the system.
```

docs/guide/templates.md
```diff
@@ -70,13 +72,24 @@ This is what individual keys in the configuration mean:
...
+For more details, see [View provider configuration](content_rendering#configuring-views-the-viewprovider).
```
MagdalenaZuba:
```suggestion
For more details, see [View provider configuration](content_rendering.md#configuring-views-the-viewprovider).
```

docs/guide/templates.md
```diff
@@ -291,12 +368,18 @@ Linking to other Locations is done with a [native `path()` Twig helper](http://
 <a href="{{ path( location ) }}">Some link to a Location</a>
 ```
 
-If you don't have the Location object, but only its ID, you can generate the URLAlias the following way:
+If you don't have the Location object, but only its Location ID, you can generate the URLAlias the following way:
```
MagdalenaZuba:
```suggestion
If you don't have the Location object, but only its ID, you can generate the URL alias the following way:
```
Is that Location ID or a Content ID. From the context I think it is the other but not sure, so I would left ID to be safe.

PR #778, opened 2019-09-25
EZP-30911: Added option to pass `location-id` param to URL alias reg command

michalmilc90:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | [EZP-30911](https://jira.ez.no/browse/EZP-30911)
| Versions      | 2.5

The `ezplatform:urls:regenerate-aliases` now accepts an additional `--location-id` parameter.

### DIFF
--- a/docs/guide/url_management.md
+++ b/docs/guide/url_management.md
@@ -171,7 +171,7 @@ bin/console ezplatform:urls:regenerate-aliases
 
 You can also extend the command by the following parameters:
 
-- `--iteration-count` — to define how many Locations should be processed at once to avoid too much memory use
+- `--iteration-count` — to define how many Locations should be processed at once to reduce memory usage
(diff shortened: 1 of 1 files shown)


docs/guide/url_management.md
```diff
@@ -169,7 +169,10 @@ Before applying the command, back up your database and make sure it is not modif
 bin/console ezplatform:urls:regenerate-aliases
 ```
 
-Use an `--iteration-count` parameter to define how many Locations should be processed at once, to avoid too much memory use.
+You can also extend the command by the following parameters:
+
+- `--iteration-count` — to define how many Locations should be processed at once to avoid too much memory use
```
MagdalenaZuba:
```suggestion
- `--iteration-count` — to define how many Locations should be processed at once to reduce memory usage
```

PR #781, opened 2019-09-26
EZP-30948: Remove Setup and Design folders

DominikaK:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | https://jira.ez.no/browse/EZP-30948

Also clears up incorrect naming of Content instead of Home.

### DIFF
--- a/docs/guide/content_management.md
+++ b/docs/guide/content_management.md
@@ -30,13 +30,12 @@ Under this root there are preset top level Locations in each installation which 
 
 ![Content and Media top Locations](img/content_structure_media_library.png)
 
-#### Home
+#### Content
 
-**Home** is the top level Location for the actual contents of a site.
+The top level Location for the actual contents of a site
+can be viewed by selecting the **Content structure** tab in the Content mode interface.
 This part of the tree is typically used for organizing folders, articles, information pages, etc.
-This means that it contains the actual content structure of the site,
-which can be viewed by selecting the **Content structure** tab in the Content mode interface.
-The default ID number of the **Home** Location is 2, but it can be [modified via configuration](config_repository.md#top-level-locations).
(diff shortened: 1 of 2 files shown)


docs/guide/content_management.md
```diff
@@ -30,13 +30,13 @@ Under this root there are preset top level Locations in each installation which
 
 ![Content and Media top Locations](img/content_structure_media_library.png)
 
-#### Content
+#### Home
 
-**Content** is the top level Location for the actual contents of a site.
+**Home** is the top level Location for the actual contents of a site.
```
alongosz: Hmm, this is actually more complicated. It's a name of a Content at this top Location. In EE it's "Home", but in OSS it's "eZ Platform". So on an abstract level we can call it "Content". Maybe it should be rephrased so rather abstract term is used, not suggesting any name. Top level location for "Content structure" also would sound good here.

docs/guide/content_management.md
```diff
@@ -30,13 +30,13 @@ Under this root there are preset top level Locations in each installation which
...
 This part of the tree is typically used for organizing folders, articles, information pages, etc.
 This means that it contains the actual content structure of the site,
 which can be viewed by selecting the **Content structure** tab in the Content mode interface.
-The default ID number of the **Content** Location is 2, but it can be [modified via configuration](config_repository.md#top-level-locations).
+The default ID number of the **Home** Location is 2, but it can be [modified via configuration](config_repository.md#top-level-locations).
```
alongosz: I think we should mention here that this is for back-office / AdminUI purpose only. Repository has no knowledge of "Content structure" Location configuration (maybe except for SiteAccess tree root).

DominikaK: Added clarification in the link's target page.

PR #846, opened 2019-10-22
Update twig_functions_reference.md to include information on "original" image variation

mateuszbieniek:
There is no place in documentation that mentions `original` image variation, which returns unmodified, "original" image variation, but it is mentioned in product files, for example, https://github.com/ezsystems/ezpublish-kernel/blob/7.5/eZ/Bundle/EzPublishCoreBundle/Resources/views/content_fields.html.twig#L425

### DIFF
--- a/docs/guide/twig_functions_reference.md
+++ b/docs/guide/twig_functions_reference.md
@@ -266,7 +266,7 @@ The filters also accept an optional `timezone` parameter for displaying date and
 |---------------|--------------------------------------------------------|-------------------------------------------|
 | `field`       | `eZ\Publish\API\Repository\Values\Content\Field`       | The image Field                           |
 | `versionInfo` | `eZ\Publish\API\Repository\Values\Content\VersionInfo` | The VersionInfo that the Field belongs to |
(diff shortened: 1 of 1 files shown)


docs/guide/twig_functions_reference.md
```diff
@@ -266,7 +266,7 @@ The filters also accept an optional `timezone` parameter for displaying date and
 |---------------|--------------------------------------------------------|-------------------------------------------|
 | `field`       | `eZ\Publish\API\Repository\Values\Content\Field`       | The image Field                           |
 | `versionInfo` | `eZ\Publish\API\Repository\Values\Content\VersionInfo` | The VersionInfo that the Field belongs to |
-| `variantName` | `string`                                               | Name of the image variation to be used        |
+| `variantName` | `string`                                               | Name of the image variation to be used. To display original image alias use `original` as a image variation |
```
michalmilc90:
```suggestion
| `variantName` | `string`                                               | Name of the image variation to be used. To display original image alias, use `original` as an image variation |
```

PR #849, opened 2019-10-24
Text style update

MagdalenaZuba:
| Question      | Answer
| ------------- | ---
| Versions      | all

We wanted to unify style of all pages with text.


MagdalenaZuba:
@inakijv ping for review

PR #856, opened 2019-11-04
Suggest the design engine for default templates

bdunogier:
In my opinion, it would be the preferred way for most.
Maybe a link to a general page about the design engine would be nice.

Are there other places where we recommend config for templates ?

### DIFF
--- a/docs/guide/content_rendering.md
+++ b/docs/guide/content_rendering.md
@@ -193,7 +193,7 @@ parameters:
     ezplatform.default_view_templates.content.full: "content/view/full.html.twig"
 ```
 
-Alternatively, you can do it using the design engine. For example, if your theme is `site`,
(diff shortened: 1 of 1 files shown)


docs/guide/content_rendering.md
```diff
@@ -193,6 +193,9 @@ parameters:
     ezplatform.default_view_templates.content.full: "content/view/full.html.twig"
 ```
 
+Alternatively, the design engine can be used for this. If your theme is `site`, create `themes/site/default/content/{viewType}.html.twig` to use it instead of the core one.
```
DominikaK:
```suggestion
Alternatively, you can do it using the design engine. For example, if your theme is `site`,
create `themes/site/default/content/<viewType>.html.twig` to override the core template.
```

PR #857, opened 2019-11-06
enableSingleRuntimeChunk() caution to avoid having multiple JS calls

konradoboza:
Recently @arfaram discovered an issue regarding multiple calls when adding a single entry to custom Webpack configuration. This caution explains what is the cause and how to fix that unwanted behavior.

### DIFF
--- a/docs/guide/bundles.md
+++ b/docs/guide/bundles.md
@@ -155,7 +155,7 @@ add a `Resources/encore/ez.webpack.custom.config.js` file, for example:
 
 !!! tip
 
-    If you don't plan to add multiple entry files on the same page in your custom config, you need to use `disableSingleRuntimeChunk()` funtion to avoid adding a separate `runtime.js` file. Otherwise your JS code may be ran multiple times. By default `enableSingleRuntimeChunk()` function is used.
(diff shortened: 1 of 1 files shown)


docs/guide/bundles.md
```diff
@@ -153,6 +153,10 @@ add a `Resources/encore/ez.webpack.custom.config.js` file, for example:
 	module.exports = customConfig;
 ```
 
+!!! tip
+
+    If you don't plan to add multiple entry files on the same page in your custom config, you need to use `disableSingleRuntimeChunk()` funtion to avoid adding a separate `runtime.js` file. Otherwise your JS code may be ran multiple times. By default `enableSingleRuntimeChunk()` function is used.
```
michalmilc90:
```suggestion
    If you don't plan to add multiple entry files on the same page in your custom config, use the `disableSingleRuntimeChunk()` funtion to avoid adding a separate `runtime.js` file. Otherwise, your JS code may be run multiple times. By default, the `enableSingleRuntimeChunk()` function is used.
```

michalmilc90:
note: merge to 2.5.x and master

PR #915, opened 2020-01-16
Fix backup db script to cover PostgreSQL

DominikaK:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | n/a
| Versions      | 2.5, master

Database dump command on the Backup did not cover PostgreSQL

### DIFF
--- a/docs/guide/backup.md
+++ b/docs/guide/backup.md
@@ -23,7 +23,7 @@ var/logs/*
  
 ```
 # MySQL
-mysqldump -u root --add-drop-table <database_name> > db_backup.sql
+mysqldump -u <database_user> --add-drop-table <database_name> > db_backup.sql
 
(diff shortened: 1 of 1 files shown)


docs/guide/backup.md
```diff
@@ -22,7 +22,11 @@ var/logs/*
 3\. Create a dump of the database:
  
 ```
-mysqldump -u root --add-drop-table example > db_backup.sql
+# MySQL
+mysqldump -u root --add-drop-table <database_name> > db_backup.sql
```
alongosz: Let's also change "root" to `<database_user>`. If a production setup uses root user it's rather bad, so unlikely to happen.
```suggestion
mysqldump -u <database_user> --add-drop-table <database_name> > db_backup.sql
```

PR #917, opened 2020-01-16
2.5.6 introduce password expiration and not recovery

arfaram:
In 2.5.6 we introduce a new column in the ezsuer DB table which is in defined in dbupdate-7.5.4-to-7.5.5.sql file. This column will be used to save current time when user password is changed. The Password Recovery process is using the same process like forgot password introduced in previous release.

| Question      | Answer
| ------------- | ---
| Versions      | 2.5.6

### DIFF
--- a/docs/updating/4_update_2.5.md
+++ b/docs/updating/4_update_2.5.md
@@ -86,7 +86,7 @@ bin/console cache:pool:clear cache.redis
 
 ## Updating to 2.5.6
 
-### Password Expiration
+### Password expiration
 
 Run the following script to update the database:
 


docs/updating/4_update_2.5.md
```diff
@@ -86,7 +86,7 @@ bin/console cache:pool:clear cache.redis
 
 ## Updating to 2.5.6
 
-### Password recovery
+### Password Expiration
```
MagdalenaZuba:
```suggestion
### Password expiration
```

PR #932, opened 2020-02-04
EZP-30583: Improve beginner tutorial code

DominikaK:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | https://jira.ez.no/browse/EZP-30583
| Versions      | master

- Homepage now uses a Query Type instead of a custom controller to render Rides
- Homepage is paginated using `pagingQueryAction`
- RideController uses the Content Service instead of injecting the repository

Additionally:
- Removes mention of using the built-in php server
- Fixes including jquery
- Removes unneeded step when setting up webpack encore (cf. https://github.com/ezsystems/ezplatform/pull/449)

### DIFF
--- a/docs/tutorials/platform_beginner/5_display_a_list_of_content_items.md
+++ b/docs/tutorials/platform_beginner/5_display_a_list_of_content_items.md
@@ -39,7 +39,7 @@ In `templates/full/home_page.html.twig` replace the "Hello world" with a table t
 {% endblock %}
 ```
 
-The `rides` variable you use in line 5 above needs to contains a list of all Rides.
+The `rides` variable you use in line 5 above needs to contain a list of all Rides.
 To get this list, you will use a Query Type.
 
 ## Create a QueryType for the home page
@@ -111,7 +111,7 @@ You defined the name `Ride` in the Query Type file in the `getName` method.
 
 Using the `pagingQueryAction` of the built-in `ez_query` controller (line 6)
 enables you to automatically get paginated results.
-You set the limit of results per page in the `limit` parameter.
+You can set the limit of results per page in the `limit` parameter.
 
 ### View types
 


docs/tutorials/platform_beginner/3_customize_the_front_page.md
```diff
@@ -76,11 +76,13 @@ To add a template like this to your site, create a `main_layout.html.twig` file
...
     <title>eZ Platform Beginner Tutorial</title>
 
-    {{  encore_entry_link_tags('tutorial') }}
+    {{ encore_entry_link_tags('tutorial') }}
 
     <link href='http://fonts.googleapis.com/css?family=Oswald' rel='stylesheet' type='text/css'>
     <link href='http://fonts.googleapis.com/css?family=Roboto+Slab' rel='stylesheet' type='text/css'>
     <link href='http://fonts.googleapis.com/css?family=Fjalla+One' rel='stylesheet' type='text/css'>
+
+    <script src="{{ asset('bundles/ezplatformadminuiassets/vendors/jquery/dist/jquery.min.js') }}"></script>
```
dew326: It's safer to include jquery before files that need it. So I would move it before `{{ encore_entry_link_tags('tutorial') }}`

docs/tutorials/platform_beginner/5_display_a_list_of_content_items.md
```diff
@@ -6,130 +6,121 @@ In this step you will display a table of all Rides on the front page.
...
-``` html+twig hl_lines="5"
+``` html+twig hl_lines="5 18"
 {% extends "main_layout.html.twig" %}
 
 {% block content %}
     <div class="col-xs-10 col-xs-offset-1 text-justified">
-        {{ render_esi( controller( 'App\\Controller\\HomepageController::getAllRidesAction' ) ) }}
+        {% for ride in rides.currentPageResults %}
+            {% if loop.first %}
+                <table class="table table-hover">
```
dew326: Probably what is inside the `loop.first` and `loop.last` ifs doesn't have to be in the for loop.

docs/tutorials/platform_beginner/7_embed_content.md
```diff
@@ -62,36 +62,36 @@ Add the template for the line view of a Landmark by creating `templates/line/lan
...
+            {{ ez_render_field( content, 'photo', { parameters: { 'alias': 'landmark_list', 'class': 'img-rounded'}}) }}
+        </a>
+    </div>
+
+    {# MODAL #}
+    <div class="bike-modal modal fade" id="bikeModal{{ content.id }}" tabindex="-1" role="dialog" aria-hidden="true">
+        <div class="modal-content">
+            <div class="close-modal" data-dismiss="modal">
+                <div class="lr">
+                    <div class="rl">
```
dew326: Those `divs` with class `rl` and `lr` looks strange. What is the purpose of having it?

DominikaK: They are responsible for creating the "X" from the exit button from the module: 
image
There's probably  more elegant way to do it?

docs/tutorials/platform_beginner/7_embed_content.md
```diff
@@ -116,27 +116,25 @@ namespace App\Controller;
...
-     * Action used to display a ride
-     *    - Adds the list of all related Landmarks to the response.
-     *
-     * @param ContentView $view
-     *
-     * @return ContentView $view
-     */
+    private $contentService;
+
+    public function __construct(ContentService $contentService)
```
ViniTou: Not sure it this is mentioned anywhere, but that will only works with `autowiring` enabled

DominikaK: The assumption is that the tutorial is started from a clean install, and it is enabled by default.

docs/tutorials/platform_beginner/5_display_a_list_of_content_items.md
```diff
@@ -6,130 +6,117 @@ In this step you will display a table of all Rides on the front page.
...
+                    {{ pagerfanta( rides, 'ez', {'routeName': location } ) }}
+                </div>
+            </div>
+        {% endif %}
     </div>
 {% endblock %}
 ```
 
-The application will look for a `getAllRidesAction` inside the `HomepageController.php` located in `App\Controller`.
+The `rides` variable you use in line 5 above needs to contains a list of all Rides.
```
MagdalenaZuba:
```suggestion
The `rides` variable you use in line 5 above needs to contain a list of all Rides.
```

docs/tutorials/platform_beginner/5_display_a_list_of_content_items.md
```diff
@@ -6,130 +6,117 @@ In this step you will display a table of all Rides on the front page.
...
+                        limit: 4
+                        assign_results_to: rides
 ```
 
+The `query_type` parameter in line 13 indicates which Query Type to use.
+You defined the name `Ride` in the Query Type file in the `getName` method.
+
+Using the `pagingQueryAction` of the built-in `ez_query` controller (line 6)
+enables you to automatically get paginated results.
+You set the limit of results per page in the `limit` parameter.
```
MagdalenaZuba:
```suggestion
You can set the limit of results per page in the `limit` parameter.
```

PR #936, opened 2020-02-11
EZP-31316: Query FT pagination

DominikaK:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | https://jira.ez.no/browse/EZP-31316
| Versions      | 2.5, master (3.0)

### DIFF
--- a/docs/guide/controllers.md
+++ b/docs/guide/controllers.md
@@ -595,6 +595,18 @@ The following variables are available in parameter expressions:
 
 You can paginate the query results by checking the **Enable pagination** box and selecting a limit of results per page.
 
+The following optional parameters are available:
+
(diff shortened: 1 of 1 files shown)


docs/guide/controllers.md
```diff
@@ -615,6 +619,10 @@ Query results are provided to the template in the `items` variable:
         "viewType": itemViewType
     })) }}
 {% endfor %}
+
+{% if isPaginationEnabled %}
```
bdunogier: We must document that variable, as well as `pageParameter` below.

In addition, there are three extra parameters that can be passed to the `ez_render_field` twig function: `enablePagination`, `disablePagination` and `itemsPerPage`. You can find the documentation for them in https://github.com/ezsystems/ezplatform-query-fieldtype/blob/master/doc/howto/customize_rendering.md.

PR #941, opened 2020-02-18
EZP-30751: DFS file permissions are ignored

konradoboza:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | [EZP-30751](https://jira.ez.no/browse/EZP-30751)
| Versions      | 2.5

This PR contains an example config that can be used for specifying permissions for files generated when NFS adapter is in use.

### DIFF
--- a/docs/guide/file_management.md
+++ b/docs/guide/file_management.md
@@ -61,7 +61,7 @@ Default values are 0644 for files and 0755 for directories.
 
 !!! note
 
-    In case when NFS [adapter](file_management.md#adapter) is in use, you should configure file permissions under `oneup_flysystem` key instead, as follows:
+    When using the NFS [adapter](file_management.md#adapter), configure file permissions under the `oneup_flysystem` key instead, as follows:
     
     ``` yaml
     oneup_flysystem:


docs/guide/file_management.md
```diff
@@ -59,6 +59,22 @@ Default values are 0644 for files and 0755 for directories.
 
     Make sure to configure permissions using a number and *not* a string. If you write "0644" it will *not* be interpreted by PHP as an octal number, and unexpected permissions will be applied.
 
+!!! note
+
+    In case when NFS [adapter](file_management.md#adapter) is in use, you should configure file permissions under `oneup_flysystem` key instead, as follows:
```
DominikaK:
```suggestion
    When using the NFS [adapter](file_management.md#adapter), configure file permissions under the `oneup_flysystem` key instead, as follows:
```

andrerom:
hmm, is this possible rather a bug @bdunogier ? _(that SA config is not passed on to NFS adapter as well)_

bdunogier:
If it isn't mapped from ez settings to flysystem's, it does look like a bug indeed.

DominikaK:
@andrerom @konradoboza, if this is a bug, what do you want to do with this PR? Merge it until the issue is resolved, or close with the hope it's fixed quickly?

konradoboza:
In my opinion, we should merge this PR as-is for now, as this topic isn't covered by our documentation at all. After the NFS adapter code cleanup we can adjust this config accordingly.

andrerom:
@konradoboza That is fine, but please pass the bug to thridline so it's fixed, and point to this being reverted once that is done.

konradoboza:
@andrerom done.

PR #968, opened 2020-03-13
EZP-31415: Sibling Criterion

DominikaK:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | https://jira.ez.no/browse/EZP-31415

Note: Composite Criterion will be documented together with https://jira.ez.no/browse/EZP-26577


docs/guide/search/criteria_reference/sibling_criterion.md
```diff
@@ -0,0 +1,15 @@
...
+
+## Arguments
+
+- `locationId` - int representing the Location ID.
+- `parentLocationId` - int representing the parent Location ID.
+
+## Example
+
+``` php
+$query->query = new Criterion\Sibling(59, 2);
```
adamwojs: Instantiation using constructor is correct but in most cases better way is to use named constructor
`\eZ\Publish\API\Repository\Values\Content\Query\Criterion\Sibling::from Location` 

```
$location = $locationService->loadLocation(59);
// ...
$query->query = Criterion\Sibling::fromLocation($location);
```

DominikaK: Is it the only Criterion this applies to? The example was written this way to be consistent with other Criteria ref pages.

adamwojs: > Is it the only Criterion this applies to?
Currently yes

PR #980, opened 2020-03-23
FIX : Change name of supporter adapter Redis

amine-betari:
- Change name of supporter adapter Redis

-  Symfony\Component\Cache\Adapter\TagAware\RedisTagAwareAdapter instead of Symfony\Component\Cache\Adapter\RedisTagAwareAdapter

### DIFF
--- a/docs/guide/persistence_cache.md
+++ b/docs/guide/persistence_cache.md
@@ -135,7 +135,7 @@ for information on how to connect to Redis.
 
 There are two Redis adapters available out of the box that fit different needs.
 
-##### `Symfony\Component\Cache\Adapter\RedisTagAwareAdapter`
+##### `Symfony\Component\Cache\Adapter\TagAware\RedisTagAwareAdapter`
 
 **Requirement**: Redis server configured with eviction [`maxmemory-policy`](https://redis.io/topics/lru-cache#eviction-policies):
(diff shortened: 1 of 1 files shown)


amine-betari:
OK, maybe my mistake. Because I am in an ez platform 2.5.6 project and we still have this old namespace. (which works)

amine-betari:
but in the line 171 configuration, we put the correct namespace ?

```` yaml
services:
    cache.redis:
        # NOTE: This optimized Redis Adapter is avaiable as of 2.5LTS via https://github.com/ezsystems/symfony-tools
        class: Symfony\Component\Cache\Adapter\TagAware\RedisTagAwareAdapter
        parent: cache.adapter.redis
        tags:
            - name: cache.pool
              clearer: cache.app_clearer
              provider: 'redis://secret@example.com:1234/13'
              # Default CACHE_NAMESPACE value, see app/config/cache_pool/cache.redis.yaml for usage with e.g. multi repo.
              namespace: 'ez'`

andrerom:
As of 2.5.9 that is they wrong one, so it should be updated.

It herby is updated for next release:
https://github.com/ezsystems/ezplatform/commit/74740128e798cced11982146345573b2a077b0ba

PR #981, opened 2020-03-24
Renamed ezstudio:scheduled:publish command to ezplatform:scheduled:run

DominikaK:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | https://github.com/ezsystems/date-based-publisher/pull/164


andrerom:
Is this valid for earlier versions too?

DominikaK:
@andrerom the original PR was merged to master only, so presumably not

PR #986, opened 2020-03-25
EZP-31220: Symfony 5

DominikaK:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | https://jira.ez.no/browse/EZP-31220
| Versions      | master, 3.0

To do:
- [x] Update links to Symfony doc

Note: Changes resulting from switch to Sf5 that concern upgrading mentioned in https://github.com/ezsystems/developer-documentation/pull/953/commits/dd6e0f2b68865c1f7c01591c12cc06c911ded2dd

### DIFF
--- a/docs/guide/personalization.md
+++ b/docs/guide/personalization.md
@@ -106,7 +106,7 @@ ezrecommendation:
                     method: basic
                     login: 12345
                     password: 1234-5678-9012-3456-7890
-                ddocument_root: '%kernel.project_dir%/public/var/export/'
+                document_root: '%kernel.project_dir%/public/var/export/'
 ```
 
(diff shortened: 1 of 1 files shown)


docs/guide/personalization.md
```diff
@@ -106,7 +106,7 @@ ezrecommendation:
                     method: basic
                     login: 12345
                     password: 1234-5678-9012-3456-7890
-                document_root: '%kernel.root_dir%/../web/var/export/'
+                document_root: '%kernel.project_dir%/web/var/export/'
```
webhdx: If this should be exposed via http server then it should be rather `document_root: '%kernel.project_dir%/public/var/export/'`. @ciastektk @damianz5 correct me if I'm wrong.

PR #990, opened 2020-03-26
Added note for "upload_max_filesize" PHP setting

konradoboza:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | [EZP-31016](https://jira.ez.no/browse/EZP-31016)
| Versions      | `2.5`, `3.0`, `master`

Currently, the default max file size that can be uploaded via Image Field Type is set to 10MB. However, BO validation relies on `upload_max_filesize` setting taken from `php.ini`. Therefore, when the user has the default value which is 2MB, saving a draft is not possible without decreasing the Field Type "Maximum file size" value.

This issue applies only to Image Field Type. @ezsystems/documentation-team feel free to move the note as you see fit.

### DIFF
--- a/docs/api/field_type_reference.md
+++ b/docs/api/field_type_reference.md
@@ -880,7 +880,7 @@ The Image Field Type supports one `FieldDefinition` option: the maximum size for
       
 !!! note
 
-    As the default value for maximum size is set to 10MB we recommend setting `upload_max_filesize` key in `php.ini` configuration file to value equal or higher than that. It will prevent validation errors while editing Content Types.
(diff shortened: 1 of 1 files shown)


docs/api/field_type_reference.md
```diff
@@ -877,6 +877,10 @@ The Image Field Type supports one `FieldDefinition` option: the maximum size for
 !!!note
 
     Maximum size is rounded to 1 MB (legacy storage limitation).
+      
+!!! note
+
+    As the default value for maximum size is set to 10MB we recommend setting `upload_max_filesize` key in `php.ini` configuration file to value equal or higher than that. It will prevent validation errors while editing Content Types.
```
DominikaK:
```suggestion
    As the default value for maximum size is set to 10MB, we recommend setting the `upload_max_filesize` key in the `php.ini` configuration file to a value equal to or higher than that. It will prevent validation errors while editing Content Types.
```

PR #992, opened 2020-03-30
Adding button placement configuration for custom tag button placement

janit:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | N/A
| Versions      | 3.0

In eZ Platform 3.0 custom tags need to be explicitly placed into a menu in the system. See richtext bundle configuration for `ezyoutube`, `eztwitter` and `ezfacebook` included: https://github.com/ezsystems/ezplatform-richtext/blob/8f453bc780e7dd9ebcdce3a71e77f5be5a999fd2/src/bundle/Resources/config/prepend/ezpublish.yaml#L366

This PR adds the example button placement into two examples configuration blocks.

### DIFF
--- a/docs/extending/extending_online_editor.md
+++ b/docs/extending/extending_online_editor.md
@@ -31,6 +31,16 @@ ezplatform:
                             buttons:
                                 ezfactbox:
                                     priority: 5
+                        ezfactbox:
+                            buttons:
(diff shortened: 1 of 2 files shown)


janit:
I noticed this also needs toolbar icons defined. I will add them.

PR #997, opened 2020-03-31
Updating,  specify aws s3 bucket access

lionelakpagni:
Precising that AWS S3 bucket should be set as public

| Question      | Answer
| ------------- | ---
| JIRA Ticket   | 
| Versions      |

### DIFF
--- a/docs/guide/clustering_aws_s3.md
+++ b/docs/guide/clustering_aws_s3.md
@@ -26,7 +26,7 @@ The secret key cannot be retrieved again after the key has been created, so don'
 
 !!! note
 
-    Make sure that your bucket is [configured as Public](https://docs.aws.amazon.com/AmazonS3/latest/user-guide/block-public-access-bucket.html) to avoid facing 403 errors, as the current S3 handler requires files to be stored publicly for linking to work as intended.
+    Make sure that your bucket is [configured as Public](https://docs.aws.amazon.com/AmazonS3/latest/user-guide/block-public-access-bucket.html) to avoid facing 403 errors, as the current S3 handler is meant to store files publicly so they can be served directly from S3.
     
 ## Set up eZ Platform for AWS S3
 


docs/guide/clustering_aws_s3.md
```diff
@@ -24,6 +24,11 @@ See <http://docs.aws.amazon.com/AmazonS3/latest/dev/s3-access-control.html>.
 The secret key cannot be retrieved again after the key has been created, so don't lose it.
 (However, you can create new keys if needed.)
 
+!!! note
+
+    Make sure that your bucket is configurated as Public to avoid facing 403 errors.
```
konradoboza:
```suggestion
    Make sure that your bucket is configured as "public" to avoid facing 403 errors.
```

DominikaK:
```suggestion
    Make sure that your bucket is [configured as Public](https://docs.aws.amazon.com/AmazonS3/latest/user-guide/block-public-access-bucket.html) to avoid facing 403 errors.
```

docs/guide/clustering_aws_s3.md
```diff
@@ -24,6 +24,11 @@ See <http://docs.aws.amazon.com/AmazonS3/latest/dev/s3-access-control.html>.
...
+    [Set bucket as public](https://docs.aws.amazon.com/AmazonS3/latest/user-guide/block-public-access-bucket.html).
```
DominikaK:
```suggestion
```

docs/guide/clustering_aws_s3.md
```diff
@@ -24,6 +24,10 @@ See <http://docs.aws.amazon.com/AmazonS3/latest/dev/s3-access-control.html>.
 The secret key cannot be retrieved again after the key has been created, so don't lose it.
 (However, you can create new keys if needed.)
 
+!!! note
+
+    Make sure that your bucket is [configured as Public](https://docs.aws.amazon.com/AmazonS3/latest/user-guide/block-public-access-bucket.html) to avoid facing 403 errors.
```
andrerom: Unsure if this is correct, but if it is maybe we should specify:

```suggestion
    Make sure that your bucket is [configured as Public](https://docs.aws.amazon.com/AmazonS3/latest/user-guide/block-public-access-bucket.html) to avoid facing 403 errors, as the current S3 handler requires files to be stored publicly for linking to work as intended.
```

@lionelakpagni / @konradoboza Is this correct though?

konradoboza: Not sure, I haven't tested it. The images are served from S3 over HTTP, and they need to be public for that most likely. Otherwise, they won't be present in the Back Office due to the mentioned 403 errors (to the best of my understanding). That being said, I am ok with your suggestion.

lionelakpagni: What I understood is that current handler creates files with a public visibility, so this might be in conflict with a private bucket configuration , for, this one wont allow public files. 
So I agree with you @andrerom

docs/guide/clustering_aws_s3.md
```diff
@@ -24,6 +24,10 @@ See <http://docs.aws.amazon.com/AmazonS3/latest/dev/s3-access-control.html>.
 The secret key cannot be retrieved again after the key has been created, so don't lose it.
 (However, you can create new keys if needed.)
 
+!!! note
+
+    Make sure that your bucket is [configured as Public](https://docs.aws.amazon.com/AmazonS3/latest/user-guide/block-public-access-bucket.html) to avoid facing 403 errors, as the current S3 handler requires files to be stored publicly for linking to work as intended.
```
andrerom, line 29: @DominikaK updated text here

note: "linking" is kind of missleading on second looki t could be rewritten as
```suggestion
    Make sure that your bucket is [configured as Public](https://docs.aws.amazon.com/AmazonS3/latest/user-guide/block-public-access-bucket.html) to avoid facing 403 errors, as the current S3 handler is meant to store files publicly so they can be served directly from S3.
```

PR #1000, opened 2020-03-31
Added 'access_control' usage example

konradoboza:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | none
| Versions      | `2.5`, `3.0`, `master`

There are some API endpoints that can be blocked for an Anonymous User and for which there are no dedicated permissions/limitations. This PR adds an example usage of `access_control` mechanism. @ezsystems/documentation-team feel free to place the note somewhere else if needed.

### DIFF
--- a/docs/api/general_rest_usage.md
+++ b/docs/api/general_rest_usage.md
@@ -210,7 +210,7 @@ Session-based is the default authentication method, as it is needed for UI.
 
 !!! note "Limiting anonymous access to metadata over REST API"
     
-    There are some API endpoints that might be accessible to the Anonymous User outside of the Back Office due to missing permissions or limitations.
+    Some API endpoints accessible to the Anonymous User return metadata you might not want to expose, due to insufficient permission limitations.
     To prevent that, you can rely on the Symfony securing URL patterns mechanism called [access_control](https://symfony.com/doc/3.4/security/access_control.html).
     The example below shows you how to block listing Content Types for the non-authenticated users.
     


konradoboza:
Is there any reward for opening #1000 PR? 😄

docs/api/general_rest_usage.md
```diff
@@ -208,6 +208,17 @@ The REST API supports two authentication methods:
 
 Session-based is the default authentication method, as it is needed for UI.
 
+!!! note
+    
+    There are some API endpoints that might be accessible for the Anonymous User outside the Back Office due to missing permissions or limitations. To prevent that, you can rely on the Symfony securing URL patterns mechanizm called [access_control](https://symfony.com/doc/3.4/security/access_control.html). The example below shows how to disallow listing Content Types for the not-authenticated user.
```
andrerom, line 213: Look good 👍  Only things I see to potentially improve would be to consider to add a header, and maybe tweak the first sentence a bit.

```suggestion
!!! note "Anonymous access to meta data over REST"
    
    Some API endpoints accessible for the Anonymous User return metadata you might not want exposed, due to their nature in PHP API they don't have own permission limitations. To prevent that, you can rely on the Symfony securing URL patterns mechanizm called [access_control](https://symfony.com/doc/3.4/security/access_control.html). The example below shows how to disallow listing Content Types for the not-authenticated user.
```

konradoboza: From my perspective it looks good, I will let @ezsystems/documentation-team decide how to put this.

MagdalenaZuba, line 213:
```suggestion
    There are some API endpoints that might be accessible to the Anonymous User outside of the Back Office due to missing permissions or limitations.
    To prevent that, you can rely on the Symfony securing URL patterns mechanism called [access_control](https://symfony.com/doc/3.4/security/access_control.html).
    The example below shows you how to block listing Content Types for the non-authenticated users.
```

docs/api/general_rest_usage.md
```diff
@@ -208,6 +208,19 @@ The REST API supports two authentication methods:
 
 Session-based is the default authentication method, as it is needed for UI.
 
+!!! note
+    
+    There are some API endpoints that might be accessible to the Anonymous User outside of the Back Office due to missing permissions or limitations.
```
MagdalenaZuba, line 213:
```suggestion
    Some API endpoints accessible to the Anonymous User return metadata you might not want to expose, due to insufficient permission limitations.
```

PR #1001, opened 2020-04-01
Update 3.0 update project doc for Symfony 5

andrerom:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   |  N/A
| Versions      |  3.0/master

Not verified if the steps themselves are valid for Symfony 5 here, but added links and updated 4 => 5 for Symfony references.

### DIFF
--- a/docs/releases/ez_platform_v3.0_project_update.md
+++ b/docs/releases/ez_platform_v3.0_project_update.md
@@ -1,15 +1,15 @@
 # eZ Platform v3.0 project update
 
 Updating to v3.0 requires you to adapt your project and your custom code.
-This is due to the switch from Symfony 3 to Symfony 4, as well as due to deprecation of certain features.
+This is due to the switch from Symfony 3 to Symfony 5, as well as due to deprecation of certain features.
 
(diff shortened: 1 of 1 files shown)


andrerom:
Another topic in terms of terminology: This page seems to be about `Upgrading` from for instance 2.5, and not about `Updating` from 3.0.0 to 3.0.1.

DominikaK:
@andrerom, there is an open PR under QA right now which handles this and other necessary steps when upgrading/migrating to v3: https://github.com/ezsystems/developer-documentation/pull/953

andrerom:
Ok, I missed that.

Note in terms of terminology, unsure if you agree but I would expect it somewhat like this:
- Update: 3.0.x to 3.0.y to 3.1.z (?)
- Upgrade: 2.x to 3.x
- Migration: IMO implies much more work, so typically  eZ Publish _(typically from 5.4)_ to eZ Platform _(typically to 3.x for 3.x branch)_

DominikaK:
Moved the page to the `updating` folder under the name "Upgrading to v3"

PR #1023, opened 2020-04-15
Update recommendation bundle for v3

DominikaK:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | https://jira.ez.no/browse/EZP-31121
| Versions      | 3.0, master

### DIFF
--- a/docs/guide/personalization.md
+++ b/docs/guide/personalization.md
@@ -599,7 +599,7 @@ if the content Fields were previously exported by the export script.
 You can retrieve data returned from the recommendation engine and modify it before displaying.
 
 To do it, subscribe to `RecommendationResponseEvent`.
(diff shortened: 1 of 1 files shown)


docs/guide/personalization.md
```diff
@@ -592,6 +594,24 @@ if the content Fields were previously exported by the export script.
     }
     ```
 
+### Modifying recommendation data
+
+You can retrieve data returned from the recommendation engine and modify it before displaying.
+
+To do it, subscribe to `RecommendationResponseEvent`.
+See [`Event/Subscriber/RecommendationEventSubscriber.php`](https://github.com/ezsystems/ezrecommendation-client/blob/master/src/lib/Event/Subscriber/RecommendationEventSubscriber.php) for an example:
```
MagdalenaZuba, line 602: remember to change to a specific version later on

```suggestion
See [`Event/Subscriber/RecommendationEventSubscriber.php`](https://github.com/ezsystems/ezrecommendation-client/blob/master/src/lib/Event/Subscriber/RecommendationEventSubscriber.php) for example:
```

PR #1040, opened 2020-05-01
Update requirements.md to add php-json module

KSATE:
Add php-json module because it's required to use composer

| Question      | Answer
| ------------- | ---
| JIRA Ticket   | 
| Versions      |

### DIFF
--- a/docs/getting_started/requirements.md
+++ b/docs/getting_started/requirements.md
@@ -22,7 +22,7 @@ These setups are tested by QA and are generally recommended setups. For security
 |Web Server|Nginx 1.14</br>Apache 2.4|Nginx 1.16</br>Apache 2.4|Nginx 1.14</br>Apache 2.4|
 |DBMS|MariaDB 10.3|MariaDB 10.3</br>MySQL 8.0|MariaDB 10.3</br>MySQL 8.0|
 |PHP|PHP 7.3|PHP 7.3|PHP 7.3|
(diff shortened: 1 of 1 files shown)


andrerom:
But this is bulti-in to PHP isn't it? (checking)

andrerom:
From docs it seems to be built-in: https://www.php.net/manual/en/json.requirements.php

For Debian/Ubuntu however you are right, it's been split out into a package:
- https://packages.debian.org/buster/php-json
- https://packages.ubuntu.com/eoan/php-json

For Redhat/Centos however someone will need to check as it seems to have been built in in earlier versions at least: https://access.redhat.com/solutions/38251

So @DominikaK you'll need QA on this.

micszo:
@andrerom we checked on CentOS with PHP 5.6 and 7.1/7.2. 
On 5.6 there is no php-json package, it seems to be part of php-common (at least yum tries to update that one when you try to install php-json).
On 7.1/7.2 there is a php-json package.
The article you linked to is from 2011, PHP 7.0 was released in 2015 so that would fit.

andrerom:
Ok, so then this change is safe for 2.5 and 3.0 and master (3.0), thanks @micszo  👍 


_(It might change [again with PHP 8.0](https://wiki.php.net/rfc/always_enable_json), but we'll see once that is relevant for us  ;))_

PR #1056, opened 2020-05-19
Change "Layout" path segment in content URL

reithor:
In current clean installation `/view/content/<contentId>/full/true/<locationId>` and `/view/content/<contentId>/full/false/<locationId>`
will both render the content embeded in pagelayout (strings 'true' and 'false' both are non-empty strings).
 Using `/view/content/<contentId>/full/1/<locationId>` seems to be better as `/view/content/<contentId>/full/0/<locationId>` will actually change the rendering.

| Question      | Answer
| ------------- | ---
| JIRA Ticket   |  N/A
| Versions      |

### DIFF
--- a/docs/guide/content_rendering.md
+++ b/docs/guide/content_rendering.md
@@ -6,7 +6,7 @@ eZ Platform comes with a native controller to display your content, known as the
 
 It can also be called straight by its direct URI: 
 
-`/view/content/<contentId>/full/true/<locationId>`
+`/view/content/<contentId>/full/1/<locationId>`
 
 `/view/content/<contentId>`
 


docs/guide/content_rendering.md
```diff
@@ -6,7 +6,7 @@ eZ Platform comes with a native controller to display your content, known as the
 
 It can also be called straight by its direct URI: 
 
-`/view/content/<contentId>/full/true/<locationId>`
+`/view/content/<contentId>/full/1/<locationId>`
```
andrerom, line 9: is this maybe implying we have a bug, or are we sure it should be 0/1 ?

reithor: NOTE:
current 2.5 version (ee demo with view.yml removed) behaves the same:
http://3.250.111.56:8001/view/content/221/full/0/223 => embeded in pagelayout
http://3.250.111.56:8001/view/content/221/full/false/223 => not embeded in pagelayout

andrerom: @alongosz WDYT?

alongosz: > @alongosz WDYT?

Well... It really depends on the expectation here. If we wanted to be able to use "true" or "false" then this is a bug. But maybe we didn't?
It's one of those "who needs strict types in PHP" confusion, though here we deal with array, so nothing on language level to be done (yet).
https://github.com/ezsystems/ezpublish-kernel/blob/v7.5.7/eZ/Publish/Core/MVC/Symfony/View/ParametersInjector/NoLayout.php#L27

`(bool)'false'` is actually `true`

Let's ask @bdunogier what he meant (sorry, git blame does not forget nor forgive ;) )

Side: default/fallback value in the linked line is inconsistent with route configuration, for which the fix is still pending via ezsystems/ezpublish-kernel/pull/2656. I'm actually not sure there if this is BC break or not.

bdunogier: No need for forgiveness, it was a while ago ;)

If possible, I'd try to support both `0`/`1` and `false`/`true`. The parameter is called `layout`, so yes/no on it is fairly straightforward.

Or am I missing something ?

alongosz:
> If possible, I'd try to support both `0`/`1` and `false`/`true`. The parameter is called `layout`, so yes/no on it is fairly straightforward.
> 
> Or am I missing something ?

Ok. I would actually go for literal boolean only, but since the intention was as described above, its a bug.
Reported as [EZP-31662](https://jira.ez.no/browse/EZP-31662).
I have to close this PR as invalid, sorry :-) 

Thank you for bringing our attention to the issue though. URLs with "true"/"false" should work as well, we'll fix that.

PR #1065, opened 2020-06-01
Fix example ValueObjectVisitor

skrosoft:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | N/A
| Versions      | All versions

Fix about incorrect use of EzSystems\EzPlatformRest\Output\Generator object during example source code

### DIFF
--- a/docs/api/extending_the_rest_api.md
+++ b/docs/api/extending_the_rest_api.md
@@ -155,8 +155,10 @@ class Hello extends ValueObjectVisitor
 {
     public function visit( Visitor $visitor, Generator $generator, $data )
     {
-        $generator->startValueElement( 'Hello', $data->name );
-        $generator->endValueElement( 'Hello' );
+        $generator->startObjectElement('Hello');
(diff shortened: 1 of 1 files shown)


alongosz:
@DominikaK  this entire "Extending the REST API" guide is a bit outdated:
* for eZ Platform 3.0, Symfony 5+: It's no longer advocated to use bundles in projects, unless they're to be reused as dependencies. 
* for all versions: code examples should follow eZ Platform Code style (or at least PSR2) - e.g. no spaces before & after method arguments:
```php
        $generator->startObjectElement('Hello');
        $generator->startValueElement('Name', $data->name);
        $generator->endValueElement('Name');
        $generator->endObjectElement('Hello');
```
but this is out of scope here as it applies to the entire doc.

DominikaK:
@alongosz thanks for pointing it out, we'll add this to the list of pages to review shortly.

DominikaK:
Will be covered in https://jira.ez.no/browse/EZP-31690

PR #1068, opened 2020-06-02
Fix purge type for Varnish

andrerom:
ref: https://github.com/ezsystems/ezplatform/pull/522
Relevant for 3.0 and up.

### DIFF
--- a/docs/guide/http_cache.md
+++ b/docs/guide/http_cache.md
@@ -137,7 +137,7 @@ Settings for purge servers can be configured per SiteAccess group or SiteAccess:
 ``` yaml
 ezplatform:
     http_cache:
-        purge_type: http
+        purge_type: varnish
 
     system:
         my_siteacess:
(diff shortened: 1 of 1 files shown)


xrow:
Do we need new ez release? I still have an error on 3.0.3

andrerom:
@xrow What error? you did not mention any error anywhere.

xrow:
image

andrerom:
@xrow This PR _(like the other one you commented on as well)_ is about fixing the doc so it correctly states you should use `varnish` not `http`, did you try that?

xrow:
I looks the env var is not being picked put. Need to look into it.

xrow:
Hmmm... I have now this error wiht 3.0.4

You have requested a non-existent service "cache.redis".

image

andrerom:
That seems like something rather relevant to the CACHE_DSN issue you reported separately, not so much HttpCache?

xrow:
This is true. I just wanted to report my issues with the latest version.

PR #1074, opened 2020-06-09
Fix locale.yaml link

lionelakpagni:
link should be ... locale.yml instead of ...locale.yaml
Ok for 1.13 and 2.5 doc

| Question      | Answer
| ------------- | ---
| JIRA Ticket   | 
| Versions      |

### DIFF
--- a/docs/guide/internationalization.md
+++ b/docs/guide/internationalization.md
@@ -192,7 +192,7 @@ ezplatform:
         eng-DE: en_DE
 ```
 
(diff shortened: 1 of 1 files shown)


alongosz:
> Should this rather be fixed in kernel? Wasn't all yml files intended to be renamed to yaml in 3.x to follow Symfony?
> 
> Edit: But by now that would kind of be a break so I would understand if this should for now be fixed in doc like this instead.

Yeah, we've missed the kernel ones (it's mostly fixed in other packages AFAIR). The safest course of action would be to use `yaml` for new files only and old ones rename in Kernel's next major.

PR #1086, opened 2020-06-19
Query FT offset and Field config

DominikaK:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | https://github.com/ezsystems/ezplatform-query-fieldtype/pull/40, https://github.com/ezsystems/ezplatform-query-fieldtype/pull/42
| Versions      | master, but all

Also fixes an error in one of the related templates.

**Note**: merge also to 2.5 and 3.0 after new releases

### DIFF
--- a/docs/guide/controllers.md
+++ b/docs/guide/controllers.md
@@ -718,14 +718,14 @@ For example:
 {{ ez_render_field(content, 'posts', {'parameters': {'enablePagination': true, 'itemsPerPage': 8}}) }}
 ```
 
-You can also define an offset for the results. Provide an offset in the Query Type, or in parameters:
+You can also define an offset for the results. Provide the offset in the Query Type, or in parameters:
 
 ```
 offset: 3
 ```
 
-If pagination is disabled and offset is defined, the query's offset is added to the offset calculated for a page
-(for example, with `offset = 5` and `itemsPerPage = 10`, first page starts with 5, second page starts with 15, etc.).
+If pagination is disabled and an offset value is defined, the query's offset is added to the offset calculated for a page
+(for example, with `offset = 5` and `itemsPerPage = 10`, the first page starts with 5, the second page starts with 15, etc.).
 
 Without offset defined, pagination defines the starting number for each page
 (for example, with `itemsPerPage = 10`, first page starts with 0, second page starts with 10, etc.).


docs/guide/controllers.md
```diff
@@ -718,6 +718,12 @@ For example:
 {{ ez_render_field(content, 'posts', {'parameters': {'enablePagination': true, 'itemsPerPage': 8}}) }}
 ```
 
+You can also define an offset for the results. Provide an offset in the Query Type, or in parameters:
+
+```
+offset: 3
```
bdunogier, line 724: We may have to explain what will happen with the offset:
- if no offset is defined in the query
  - if pagination is disabled, it will use 0
  - if pagination is enabled, it will add the itemsPerPage count to the offset (page1 = 10, page2 = 20, etc)
- if an offset is defined in the query
  - if pagination is disabled, it will use that offset
  - if pagination is disabled, the query's offset will be added to the offset computed for a page (offset = 5, itemsPerPage = 10, page0 = offset 5, page1 = offset 15, page 2 = offset 25)

bdunogier: Looks good to me !

docs/guide/controllers.md
```diff
@@ -718,6 +718,18 @@ For example:
...
+```
+
+If pagination is disabled and offset is defined, the query's offset is added to the offset calculated for a page
+(for example, with `offset = 5` and `itemsPerPage = 10`, first page starts with 5, second page starts with 15, etc.).
```
dabrt, line 728:
```suggestion
(for example, with `offset = 5` and `itemsPerPage = 10`, the first page starts with 5, the second page starts with 15, etc.).
```

docs/guide/controllers.md
```diff
@@ -718,6 +718,18 @@ For example:
 {{ ez_render_field(content, 'posts', {'parameters': {'enablePagination': true, 'itemsPerPage': 8}}) }}
 ```
 
+You can also define an offset for the results. Provide an offset in the Query Type, or in parameters:
```
dabrt, line 721:
```suggestion
You can also define an offset for the results. Provide the offset in the Query Type, or in parameters:
```

docs/guide/controllers.md
```diff
@@ -718,6 +718,18 @@ For example:
...
+
+```
+offset: 3
+```
+
+If pagination is disabled and offset is defined, the query's offset is added to the offset calculated for a page
```
dabrt, line 727:
```suggestion
If pagination is disabled and an offset value is defined, the query's offset is added to the offset calculated for a page
```

PR #1087, opened 2020-06-22
DB update caution: Do not force it

glye:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | N/A
| Versions      | All

Don't use the force, Luke.

Doc page: https://doc.ezplatform.com/en/latest/updating/4_update_database/

Ref https://ezcommunity.slack.com/archives/C3YFR5ZM4/p1592810966225800?thread_ts=1592403114.218800&cid=C3YFR5ZM4

### DIFF
--- a/docs/updating/4_update_database.md
+++ b/docs/updating/4_update_database.md
@@ -31,4 +31,6 @@ Only after applying all changes your database will work properly.
 
     After updating the database, clear the cache.
     
-    Do not use `--force` argument for `mysql` / `psql` commands when performing update queries. If there is any problem applying the updates, it is best if the query fails immediately, so you can fix the underlying problem before trying to execute the update again. If you leave this for later you risk ending up with an incompatible database, though the problems might not surface until later.
+    Do not use `--force` argument for `mysql` / `psql` commands when performing update queries.
+    If there is any problem during the update, it is best if the query fails immediately, so you can fix the underlying problem before you execute the update again.
+    If you leave this for later you risk ending up with an incompatible database, though the problems might not surface immediately.


docs/updating/4_update_database.md
```diff
@@ -30,3 +30,5 @@ Only after applying all changes your database will work properly.
     Always back up your data before running any database update scripts.
 
     After updating the database, clear the cache.
+    
+    Do not use `--force` on update queries. If there is any problem applying the updates, it is best if the query fails immediately, so you can fix the underlying problem before trying to execute the update again. If you leave this for later, you risk ending up with an incompatible database, and larger problems further down the road.
```
andrerom, line 34:
```suggestion
    Do not use `--force` argument for `mysql` / `psql` when performing update queries. If there is any problem applying the updates, it is best if the query fails immediately, so you can fix the underlying problem before trying to execute the update again. If you leave this for later, you risk ending up with an incompatible database that might not surface before further down the road.
```

glye: Ok, good, but then I'll tune the last sentence a bit further.

docs/updating/4_update_database.md
```diff
@@ -30,3 +30,5 @@ Only after applying all changes your database will work properly.
     Always back up your data before running any database update scripts.
 
     After updating the database, clear the cache.
+    
+    Do not use `--force` argument for `mysql` / `psql` commands when performing update queries. If there is any problem applying the updates, it is best if the query fails immediately, so you can fix the underlying problem before trying to execute the update again. If you leave this for later, you risk ending up with an incompatible database, that might not surface before further down the road.
```
glye, line 34:
```suggestion
    Do not use `--force` argument for `mysql` / `psql` commands when performing update queries. If there is any problem applying the updates, it is best if the query fails immediately, so you can fix the underlying problem before trying to execute the update again. If you leave this for later you risk ending up with an incompatible database, though the problems might not surface until later.
```

docs/updating/4_update_database.md
```diff
@@ -30,3 +30,5 @@ Only after applying all changes your database will work properly.
     Always back up your data before running any database update scripts.
 
     After updating the database, clear the cache.
+    
+    Do not use `--force` argument for `mysql` / `psql` commands when performing update queries. If there is any problem applying the updates, it is best if the query fails immediately, so you can fix the underlying problem before trying to execute the update again. If you leave this for later you risk ending up with an incompatible database, though the problems might not surface until later.
```
MagdalenaZuba, line 34:
```suggestion
    Do not use `--force` argument for `mysql` / `psql` commands when performing update queries.
    If there is any problem during the update, it is best if the query fails immediately, so you can fix the underlying problem before you execute the update again.
    If you leave this for later you risk ending up with an incompatible database, though the problems might not surface immediately.
```

PR #1090, opened 2020-06-25
Created example using "url:" parameter name when defining repository …

vidarl:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | N/A
| Versions      | 3.0

In 3.0, the default doctrine config uses `url:`, instead of `host`, `port`, `dbname` etc.
I think the documentation should also showcase the use of `url:` to avoid confusion..

### DIFF
--- a/docs/guide/config_repository.md
+++ b/docs/guide/config_repository.md
@@ -104,7 +104,7 @@ ezplatform:
 ```
 # .env.local
 
-SECOND_DATABASE_URL=otherdb://otheruser:otherpasswd@otherhost:3306/otherdbname?serverVersion=mariadb-10.3.0
+SECOND_DATABASE_URL=otherdb://otheruser:otherpasswd@otherhost:otherport/otherdbname?otherdbserversion
 ```
 
 ## Field groups configuration


andrerom:
+1, but proper person to do dev review here would be for instance @alongosz

docs/guide/config_repository.md
```diff
@@ -89,6 +101,18 @@ ezplatform:
...
 ```
 
+```
+# .env.local
+DATABASE_USER=ezplatform
+DATABASE_PASSWORD=password
+DATABASE_NAME=ezplatform
+DATABASE_HOST=localhost
+DATABASE_PORT=3306
+DATABASE_PLATFORM=mysql
```
alongosz, line 111: @vidarl do we really need all those variables? It's done like that for Docker setup, but I don't think it applies to a second instance as well.

vidarl: You are right. In this case I could use the actual values directly when setting `SECOND_DATABASE_URL`

vidarl: Sorry @alongosz, forgot about your suggestion.. Looks OK now ?

dabrt: Guys, sorry to interfere, but based on this PR:
https://github.com/ezsystems/ezplatform/pull/576
should you include the DATABASE_VERSION parameter here as well?

vidarl: Ok, I was not aware of that [fresh change](https://github.com/ezsystems/ezplatform/pull/576)...
I have added DATABASE_VERSION now

docs/guide/config_repository.md
```diff
@@ -89,6 +101,12 @@ ezplatform:
             repository: second_repository
 ```
 
+```
+# .env.local
+
+SECOND_DATABASE_URL=otherdb://otheruser:otherpasswd@otherhost:3306/otherdbname?serverVersion=mariadb-10.3.0
```
alongosz, line 107: I would either skip the port or state `:otherdbport`, the same goes for `mariadb-10.3.0`.
But that's just minor nitpick, should be known to someone who even tries have multiple connections to different instances.

vidarl, line 107:
```suggestion
SECOND_DATABASE_URL=otherdb://otheruser:otherpasswd@otherhost:otherport/otherdbname?otherdbserversion
```

vidarl: Okay, changed into `otherdbport` and `otherdbserversion`

DominikaK (approved):

PR #1116, opened 2020-07-27
Add note on permissions when using API in document field mappers

lionelakpagni:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | 
| Versions      | 1.13, 2.5 and 3.5

Make more explicit the usage of Sudo or SPI layer.

### DIFF
--- a/docs/guide/search/solr.md
+++ b/docs/guide/search/solr.md
@@ -576,7 +576,9 @@ my_webinar_app.webinar_event_title_fulltext_field_mapper:
 	
 !!! caution "Permission issues when using Repository API in document field mappers"
 
-    Document field mappers are low level, they expect to be able to index all content regardless of current user permissions. If you use Repository API in your custom document field mappers you'll need to use [sudo](https://doc.ezplatform.com/en/2.5/api/public_php_api/#using-sudo), alternativly use Persistance SPI layer as showcased in example above.
+    Document field mappers are low level. They expect to be able to index all content regardless of current user permissions.
+    If you use PHP API in your custom document field mappers, you need to apply [`sudo()`](../../api/public_php_api.md#using-sudo),
+    otherwise use the Persistence SPI layer as in the example above.
 
 ## Configuring Solr Replication (master/slave)
 


docs/guide/search/solr.md
```diff
@@ -501,7 +501,11 @@ These can be used on the extension points by registering them with the container
 
 The following example shows how to index data from the parent Location content, in order to make it available for full-text search on the children content.
 It is based on the use case of indexing webinar data on the webinar events, which are children of the webinar. Field mapper could then look like this:
+	
+	!!! caution "checking permissions"
 
+You may face some permissions issue when trying to load contents using contentService. To avoid this, use [sudo](https://doc.ezplatform.com/en/2.5/api/public_php_api/#using-sudo).
+If you dont want to use sudo, you can use internalLoadContentById (SPI layer) that dont check permissions.
```
andrerom, line 508: I would rather make the caution about wrong usage of API in this layer somehow, and I would skip mentioning any internal methods. Unsure how to word it, but something like:

```suggestion
	!!! caution "Permission issues when using Repository API in document field mappers"

Document field mappers are low level, they expect to be able to index all content regardless of current user permissions. For indexing Content the system itself is thus using SPI Persistence layer, so you you use API in your custom document field mappers you'll need to use [sudo](https://doc.ezplatform.com/en/2.5/api/public_php_api/#using-sudo).
```


But, someone in engineering overseeing search should review this.

docs/guide/search/solr.md
```diff
@@ -573,6 +573,11 @@ my_webinar_app.webinar_event_title_fulltext_field_mapper:
         - {name: ezpublish.search.solr.field_mapper.content}
 ```
 
+	
+!!! caution "Permission issues when using Repository API in document field mappers"
+
+    Document field mappers are low level, they expect to be able to index all content regardless of current user permissions. If you use Repository API in your custom document field mappers you'll need to use [sudo](https://doc.ezplatform.com/en/2.5/api/public_php_api/#using-sudo), alternativly use Persistance SPI layer as showcased in example above.
```
DominikaK, line 579:
```suggestion
    Document field mappers are low level. They expect to be able to index all content regardless of current user permissions.
    If you use PHP API in your custom document field mappers, you need to apply [`sudo()`](../../api/public_php_api.md#using-sudo),
    otherwise use the Persistence SPI layer as in the example above.
```

DominikaK (approved):

PR #1124, opened 2020-08-05
Prevent confusion if visitors comming from v3.x Documentation

pboethig:
In order to save time on v3x readings

If I am an EZ Platform v3.x developer and come from the V3.x docs context I do not want to struggle in v2.x bundle documenations.

I expect, that I do not get bundle documenation on v3.x docs.

### DIFF
--- a/docs/tutorials/field_type/creating_a_tweet_field_type.md
+++ b/docs/tutorials/field_type/creating_a_tweet_field_type.md
@@ -4,6 +4,8 @@
 
     The code created in this tutorial is available on GitHub: (https://github.com/ezsystems/TweetFieldTypeBundle).
 
+# This tutorial is only processable in EZ Platform version < 3.x 
+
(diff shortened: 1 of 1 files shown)


DominikaK:
We are currently working on some redesigns to the documentation layout which will make it much clearer which version of the doc you are reading. Hopefully, once this work is completed, you shouldn't encounter confusion like thin anymore.

Because of multiple extensive changes between v2 and v3 of the product, it is not viable to cover all the differences with such warnings.

pboethig:
Please link on generic doc: https://doc.ezplatform.com/en/3.1/extending/extending_field_type/

DominikaK:
@pboethig, it is not clear to me what you are referring to. The "Creating custom Field Type" page for v3 does not link to the "Creating a Tweet Field Type" tutorial, which, as you rightly noticed, exists only in the v2 version of the doc.

PR #1127, opened 2020-08-11
Extending page doc update (EZEE-2947)

dabrt:
| Question      | Answer |
| ------------- | --- |
| JIRA Ticket   | EZEE-2947 |
| Versions      | master |

### DIFF
--- a/docs/extending/extending_page.md
+++ b/docs/extending/extending_page.md
@@ -483,7 +483,7 @@
                 thumbnail: assets/images/ez-icons.svg#block-visible-recurring
                 views:
                     default:
-                        template: @ezdesign/blocks/translate_block.html.twig
+                        template: "@ezdesign/blocks/translate_block.html.twig"
                         name: Default view
                         priority: -255
                 attributes:


docs/extending/extending_page.md
```diff
@@ -455,6 +455,163 @@
...
+      ``` json
+      "require": {
+          (...)
+          "ezplatform-i18n/ezplatform-i18n-fr_fr": "^3.0",
+          (...)
+        }
+      ```
+
+    Then, create a translatable Page block by adding the following YAML configuration
+    in an application or a bundle, under the `ezplatform_page_fieldtype` key...
```
dabrt, line 480:
```suggestion
    in an application or a bundle, under the `ezplatform_page_fieldtype` key:
```

docs/extending/extending_page.md
```diff
@@ -455,6 +455,163 @@
...
+                        type: string
+                        name: E-mail address
+                        validators:
+                            regexp:
+                                options:
+                                    pattern: '/^\S+@\S+\.\S+$/'
+                                message: Provide a valid e-mail address
+    ```
+
+    ... and, in the `blocks` directory, a corresponding Twig template called `translate_block.html.twig`:
```
dabrt, line 511:
```suggestion
    In the `blocks` directory, add a corresponding Twig template called `translate_block.html.twig`:
```

docs/extending/extending_page.md
```diff
@@ -455,6 +455,163 @@
...
+
+    ```
```
dabrt, line 513:
```suggestion
    ``` html+twig
```

docs/extending/extending_page.md
```diff
@@ -455,6 +455,163 @@
...
+    <h1>{{ name|default('Hello stranger') }}!</h1>
+    ```
+
+    Next, implement the logic that is responsible for label translation. Begin with registering a new service in the `config/services.yaml` file:
+
+    ``` yaml
+    App\Event\Subscriber\TranslateBlockNameSubscriber:
+        tags:
+            - { name: event_subscriber }
+	  ```
```
dabrt, line 523:
```suggestion
    ```
```

docs/extending/extending_page.md
```diff
@@ -455,6 +455,163 @@
     }
     ```
 
+    ### Block name translation
+
+    A practical example of how you can use the `BlockDefinitionEvents` API is translating the block name.
+    You can modify the `name` attribute of the Page block, so that it displays a translation
```
DominikaK, line 461:
```suggestion
    You can modify the `name` attribute of the Page block so that it displays a translation
```

docs/extending/extending_page.md
```diff
@@ -455,6 +455,163 @@
...
+    in one of the defined languages.
+
+    The example uses a [Symfony Translator](https://github.com/symfony/symfony/blob/5.1/src/Symfony/Component/Translation/Translator.php) module and its `trans()` method.
+    The method takes three arguments: an identifier of the block name label, an array of parameters,
```
DominikaK, line 465:
```suggestion
    The method takes three arguments: an identifier of the block name, an array of parameters,
```

docs/extending/extending_page.md
```diff
@@ -455,6 +455,163 @@
...
+    and the domain of the translation.
+
+    Start with adding the requirement to install the language package to the `composer.json` file,
```
DominikaK, line 468: Maybe use `composer require`, like we do in https://doc.ezplatform.com/en/master/guide/back_office_translations/#installing-new-translations?
Not sure if either is the recommended way.

docs/extending/extending_page.md
```diff
@@ -455,6 +455,163 @@
...
+      ``` json
+      "require": {
+          (...)
+          "ezplatform-i18n/ezplatform-i18n-fr_fr": "^3.0",
+          (...)
+        }
+      ```
+
+    Then, create a translatable Page block by adding the following YAML configuration
+    in an application or a bundle, under the `ezplatform_page_fieldtype` key:
```
DominikaK, line 480:
```suggestion
    under the `ezplatform_page_fieldtype` key:
```

docs/extending/extending_page.md
```diff
@@ -455,6 +455,163 @@
...
+
+    Next, implement the logic that is responsible for label translation. Begin with registering a new service in the `config/services.yaml` file:
+
+    ``` yaml
+    App\Event\Subscriber\TranslateBlockNameSubscriber:
+        tags:
+            - { name: event_subscriber }
+    ```
+
+    Then, implement an event subscriber that listens to the block definition event. For example, create a `src/Event/Subscriber/TranslateBlockNameSubscriber.php` file that contains the following code:
```
DominikaK, line 525:
```suggestion
    Then, implement an event subscriber that listens to the block definition event. For example, create an `src/Event/Subscriber/TranslateBlockNameSubscriber.php` file that contains the following code:
```

docs/extending/extending_page.md
```diff
@@ -455,6 +455,163 @@
...
+    After you add new files with translations, run `php bin/console cache:clear` to clear the cache.
+
+    The language to display is selected automatically based on [user preferences or browser setup](../../guide/back_office_translations/#selecting-back-office-language).
+
+    !!! note "Additional information"
+
+        For more information, see the following articles:
+
+        - [Back office translations](../guide/back_office_translations.md)
+
```
DominikaK, line 610:
```suggestion
```

docs/extending/extending_page.md
```diff
@@ -455,6 +455,163 @@
...
+        - [Symfony translations](https://symfony.com/doc/current/translation.html)
+
```
DominikaK, line 612:
```suggestion
```

DominikaK (approved):

mnocon (commented):
I was able to translate my block:
obraz
but I think there are things we could improve here.

1) Template path:
```
In the `blocks` directory, add a corresponding Twig template called `translate_block.html.twig`:
```
a) The blocks directory doesn't exist by default, I had to create it. Also the whole path is `templates/blocks` - maybe it could added here to avoid confusion for people less familiar with Symfony?
b) Should we promote the use of Design Engine? Then it would be `@ezdesign/blocks/translate_block.html.twig` in config and path `templates/themes/standard/blocks`.

2) Nitpick: IMHO creating a class first and then registering it as a Symfony service is a better order - it feels weird to register a service that doesn't exist yet.

3) The `TranslateBlockNameSubscriber` is missing a `use EzSystems\EzPlatformPageFieldType\FieldType\Page\Block\Definition\Event\BlockAttributeDefinitionEvent` call and crashes without it.

4) About translations (I've also looked at https://doc.ibexa.co/en/latest/guide/back_office_translations/):
a) Should we add an explanation where does this XML come from? Is it possible to generate it somehow and just fill in the translation messages? I know nothing about translations and that part felt a bit like magic 😄 
b) Right now there is no default translation for the Block - is it possible to extend this example a bit to show how it can be done?

mnocon (approved):
Looks good, just one thing to fix before merging.

docs/extending/extending_page.md
```diff
@@ -455,6 +455,186 @@
...
+    ezplatform_page_fieldtype:
+        blocks:
+            translate_block:
+                name: Translatable Block
+                # The group that contains the block in the Elements menu
+                category: Example
+                thumbnail: assets/images/ez-icons.svg#block-visible-recurring
+                views:
+                    default:
+                        template: @ezdesign/blocks/translate_block.html.twig
```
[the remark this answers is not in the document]
dabrt: Thank you, Marek!

PR #1130, opened 2020-08-21
Minor improvements of Upgrading eZ Platform to v3 page

MagdalenaZuba:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | Feedback meeting
| Versions      | all


andrerom (approved):
+1 LGTM

Only ultra nitpick on the image would be that `config` would be sorted before `Resources` on disk usually, but it's way better already so 👍

PR #1132, opened 2020-08-24
[3.x] Rewrite & broadly expand HttpCache doc

andrerom:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | https://github.com/ezsystems/developer-documentation/pull/1071
| Versions      | 3.x / master

This is 3.x version of the changes to HttpCache doc merged for 2.5 in https://github.com/ezsystems/developer-documentation/pull/1071


docs/guide/http_cache.md
```diff
@@ -72,573 +80,753 @@ fos_http_cache:
...
+Typically, you should not override these settings unless you know what you are doing.
 
-    ##### Clear the cache
+``` yaml
+fos_http_cache:
+    proxy_client:
+        default: varnish
+        varnish:
+            http:
+                servers: ['$http_cache.purge_servers$']
```
alongosz, line 481: Q: does this dynamic configuration still work? AFAIK it's always gonna be the one from `default` scope at compile time and having it here might lead to incorrect expectation by an integrator. Not sure what would be an alternative in this case though, except static container parameter.

alongosz (changes requested):

docs/guide/http_cache.md
andrerom: Good point, it should be 

```suggestion
                servers: ['%purge_server%']
```

andrerom: _Ignore prev comment_

Good catch, then we have an issue in v3, what I have pasted here comes directly from: https://github.com/ezsystems/ezplatform-http-cache/blob/master/src/Resources/config/fos_http_cache.yml


_Side: Have someone grepped fro usage of `$` config in 3.x for EE / Commerce?_

docs/guide/http_cache.md
```diff
@@ -72,573 +80,753 @@ fos_http_cache:
...
+and specify the URL Varnish can be reached on (in `config/packages/ezplatform.yaml`):
+
+
+| Configuration | Parameter| Environment variable| Possible values|
+|---------|--------|--------|----------|
+| ezpublish.http_cache.purge_type | `purge_type` | `HTTPCACHE_PURGE_TYPE` | local, varnish, fastly |
+| ezpublish.system.(scope).http_cache.purge_servers | `purge_server`* | `HTTPCACHE_PURGE_SERVER`* | Array of URLs to proxies when using Varnish or Fastly (`https://api.fastly.com`) |
+| ezpublish.system.(scope).http_cache.varnish_invalidate_token | `varnish_invalidate_token` | `HTTPCACHE_VARNISH_INVALIDATE_TOKEN` | (Optional) For token based authentication |
+| ezpublish.system.(scope).http_cache.fastly.service_id | `fastly_service_id` | `FASTLY_SERVICE_ID` | Service ID to authenticate with Fastly |
+| ezpublish.system.(scope).http_cache.fastly.key | `fastly_key` | `FASTLY_KEY` | Service key/token to authenticate with Fastly |
```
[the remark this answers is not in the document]
andrerom: Fixed in 89d9ae2

docs/guide/http_cache.md
```diff
@@ -72,573 +80,753 @@ fos_http_cache:
...
+  
+    ```php
+    Request::setTrustedProxies([$request->server->get('REMOTE_ADDR')], Request::HEADER_X_FORWARDED_ALL);
+    ```
+
+    When trusting remote IP like this, make sure your application is only accessible through Varnish.
+    If it is accessible in other ways you may end up trusting e.g. IP of client browser instead which would be a serious security issue.
+
+    Make sure that **all** traffic always comes from the trusted proxy/load-balancer,
+    and there is no other way to configure it.
```
vidarl, line 165:
```suggestion
    and that there is no other way to configure it.
```

better?

docs/guide/http_cache.md
```diff
@@ -72,573 +80,753 @@ fos_http_cache:
...
+                purge_servers: [http://my.varnish.server:8081]
 ```
 
-!!! tip "Environment Variables"
+!!! note "Invalidating Varnish cache using tokens"
 
-    eZ Platform uses environment variables by default, so you can define those values in the environment.
-    See [Update your Virtual Host](#update-your-virtual-host)
+    In setups where the Varnish server IP can change (for example on platform.sh/eZ Platform Cloud),
+    you can use token-based cache invalidation via [ez_purge_acl.](https://github.com/ezsystems/ezplatform-http-cache/blob/v2.1.0/docs/varnish/vcl/varnish5.vcl#L174)
```
vidarl, line 205: hmm.. We should likely document how to configure purge token too..

vidarl (approved):
Looks good besides some nitpick and @alongosz's regarding config

docs/guide/http_cache.md
andrerom: Indeed

docs/guide/http_cache.md
andrerom: @alongosz Have this been reported already or How should we report this to engineering?

docs/guide/http_cache.md
andrerom: Indeed, but unless you have proposal on that I suggest that should be followup.

docs/guide/http_cache.md
alongosz: > _Side: Have someone grepped fro usage of `$` config in 3.x for EE / Commerce?_

We've tested dynamic settings removal, so yes, but indeed it seems we've missed some configs.

> @alongosz Have this been reported already or How should we report this to engineering?

Recently (surprisingly all at once, though the issue exists since 3.0) there have been separate per-case reports of some things related to dynamic configuration not working.

We should indeed look at the bigger picture and grep the entire configuration. I need to sync with @lserwatka on that.

> 
> Also: Can we move forward with this PR despite this? After all the block here is just reflecting what's in the product.

Hmm, it depends how broken it is. If that settings injects default context (as I implied in my first comment), then it's fine, we'll probably catch it when grepping. But... the recent reported cases (for different configs) showed us that it's injected as-is, so in this case it would be set to `$http_cache.purge_servers$` verbatim. If that's the case here as well, we should provide something static as a temporary fix (or make an actual fix so the config works - probably better).

alongosz: Grouped as the Epic [EZP-31950](https://jira.ez.no/browse/EZP-31950). @mikadamczyk is looking into this now.

mikadamczyk: I checked how this works. I set different servers for two SiteAccesses. The dynamic setting is parsed on each request in `ezplatform-http-cache/src/ProxyClient/HttpDispatcherFactory.php:39`

```
        foreach ($servers as $server) {
            if (!$this->dynamicSettingParser->isDynamicSetting($server)) {
                $allServers[] = $server;
                continue;
            }

            $settings = $this->dynamicSettingParser->parseDynamicSetting($server);
            $configuredServers = $this->configResolver->getParameter(
                $settings['param'],
                $settings['namespace'],
                $settings['scope']
            );
            $allServers = array_merge($allServers, (array)$configuredServers);
        }
```

For each SiteAccess I received a correct server URLs

alongosz: Thank you for checking @mikadamczyk, looks like in this case false alarm. We can resolve this.

alongosz (approved):
No more remarks from my side.

andrerom:
Rebased and ready for doc review @DominikaK / @dabrt

DominikaK (approved):
Some minor text polishing will be done in a separate PR to cover all versions, as it will not require dev effort.

docs/guide/http_cache.md
```diff
@@ -72,573 +80,752 @@ fos_http_cache:
...
 
-For custom or eZ Platform controllers still using `X-Location-Id`, a dedicated response listener
-`XLocationIdResponseSubscriber` handles translating this to tags so the cache can be properly invalidated by this bundle.
 
-!!! note
+### Response tagging in Templates
 
-    This is currently marked as deprecated. For rendering content it is advised to refactor to use Content View.
-    For other needs there is an FOS tag handler for Twig and PHP that can be used.
+1\. `ez_http_tag_location()`
```
vidarl, line 793: Actually, the function is named [´ez_http_cache_tag_location`](https://github.com/ezsystems/ezplatform-http-cache/blob/master/src/Twig/ContentTaggingExtension.php#L40).

The nameing was changed [here](https://github.com/ezsystems/ezplatform-http-cache/commit/01e29e4c784bc3d17c7b7a4c9e67e284d3e99340) but neither developer-documentation or markdown files in ezplatform-http-cache was updated accordingly, 

The `ez_http_tag_relation_ids()` and `ez_http_tag_relation_location_ids()` functions were added later, but their nameing is not consistent. They should both be renamed with a `ez_http_cache` prefix, right ? @alongosz

DominikaK: Ok, we'll change that in another PR together with some more proofing of the text.

andrerom: Ok, possible fix on product side in https://github.com/ezsystems/ezplatform-http-cache/pull/135

PR #1133, opened 2020-08-25
Update custom policies for Symfony 5 compliance.

rhidja:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | NA
| Versions      | 3.1

### DIFF
--- a/docs/guide/custom_policies.md
+++ b/docs/guide/custom_policies.md
@@ -94,36 +94,12 @@ It is not possible to remove an existing module, function or limitation from a P
 
 ## Integrating the `PolicyProvider` into EzPublishCoreBundle
 
-For a `PolicyProvider` to be active, it must be properly declared in EzPublishCoreBundle.
-A bundle just has to retrieve CoreBundle's DIC extension and call `addPolicyProvider()`. This must be done in the bundle's `build()` method.
-
-``` php
-namespace Acme\ExampleBundle\AcmeExampleBundle;
-
-use Symfony\Component\HttpKernel\Bundle\Bundle;
-
-class AcmeExampleBundle extends Bundle
-{
-    public function build(ContainerBuilder $container)
-    {
-        parent::build($container);
-
-        // ...
- 
-        // Retrieve "ezpublish" container extension
-        $eZExtension = $container->getExtension('ezpublish');
-        // Add the policy provider
-        $eZExtension->addPolicyProvider(new MyPolicyProvider());
-    }
-}
-```
-
-If you are using eZ Platform in version 3.1 or newer, you must use the `App\Kernel` to register a new policy provider:
+For a `PolicyProvider` to be active, you have to register it in the class `src/Kernel.php`:
 
 ```php
 namespace App;
 
-use App\Security\AppPolicyProvider;
+use App\Security\MyPolicyProvider;
 use Symfony\Bundle\FrameworkBundle\Kernel\MicroKernelTrait;
 use Symfony\Component\HttpKernel\Kernel as BaseKernel;
 
@@ -138,21 +114,19 @@ class Kernel extends BaseKernel
         // Retrieve "ezpublish" container extension
         $eZExtension = $container->getExtension('ezpublish');
         // Add the policy provider
-        $eZExtension->addPolicyProvider(new AppPolicyProvider());
+        $eZExtension->addPolicyProvider(new MyPolicyProvider());
     }
 }
 ```
-
 
 ## Integrating custom Limitation types with the UI
 
 To provide support for editing custom policies in the Back Office, you need to implement [`EzSystems\EzPlatformAdminUi\Limitation\LimitationFormMapperInterface`](https://github.com/ezsystems/ezplatform-admin-ui/blob/master/src/lib/Limitation/LimitationFormMapperInterface.php).
 
-Next, register the service in DIC (Dependency Injection Container) with the `ez.limitation.formMapper` tag and set the `limitationType` attribute to the Limitation type's identifier:
+Next, register the service with the `ez.limitation.formMapper` tag and set the `limitationType` attribute to the Limitation type's identifier:
 
 ```yaml
-app.security.limitation.custom_limitation.mapper:
-    class: 'App\Security\Limitation\Mapper\CustomLimitationFormMapper'
+App\Security\Limitation\Mapper\CustomLimitationFormMapper:
     arguments:
         # ...
     tags:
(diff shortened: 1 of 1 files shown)


docs/guide/custom_policies.md
```diff
@@ -118,6 +118,32 @@ class AcmeExampleBundle extends Bundle
 }
 ```
 
+If you are using Symfony 5 you have to use the `App\Kernel` to register a new policy provider :
```
andrerom, line 121: as this is targeting 3.1 and higher, do we need to mention things like `if you are suing sf 5` ?

rhidja: Can replace Symfony 5, EzPlatform 3.1

andrerom:
@dabrt Change should rather replace the way it was done before which is no longer valid _(**if** that is the case)_, rather then a If / else depending on version _(already covered by branching)_.

docs/guide/custom_policies.md
```diff
@@ -118,6 +118,32 @@ class AcmeExampleBundle extends Bundle
 }
 ```
 
+If you are using EzPlatform 3.0+ you have to use the `App\Kernel` to register a new policy provider :
```
dabrt, line 121:
```suggestion
If you are using EzPlatform 3.0+ you must use the `App\Kernel` to register a new policy provider:
```

dabrt (changes requested):

docs/guide/custom_policies.md
andrerom: That is not my point, your suggestion towards 3.1 doc will result in 3.1 doc containing two instructions for how to do this. One which you seems to indicate does not work on 3.1 and then a second way for 3.1

Summary: 3.1 branch is 3.1 doc => 3.1 doc only need instructions for how you do things on 3.1, we don't need both here.

dabrt (approved):

DominikaK:
@rhidja , I've had to modify branch history, can you please rebase this PR against the current 3.1 branch?

rhidja:
@DominikaK rebase done!

andrerom:
@dabrt thanks for review ping, but I can't review as long as I don't quite get why we don't just adjust the code example above.

As in the change here seems to indicate the way to do it changed with 3.x, so why keep doc on how it's done in earlier releases? Or is there some reason to show both approaches?

rhidja:
@andrerom @dabrt  il removed the policy definion for old ezpaltform versions.

docs/guide/custom_policies.md
```diff
@@ -125,8 +126,7 @@ To provide support for editing custom policies in the Back Office, you need to i
 Next, register the service in DIC (Dependency Injection Container) with the `ez.limitation.formMapper` tag and set the `limitationType` attribute to the Limitation type's identifier:
```
andrerom, line 125: Not part of the changes here but, @DominikaK we kind of stopped shortening Dependency Injection Container didn't we?

andrerom (approved):

docs/guide/custom_policies.md
DominikaK: If you're referring to using the DIC acronym, then yes, we need to unify the way we refer to registering services (in most places we just say "register the service..." without spelling out DIC). To be done in some other PR throughout the whole doc.

rhidja: @andrerom, @DominikaK done.

mikadamczyk (approved):

docs/guide/custom_policies.md
```diff
@@ -94,26 +94,27 @@ It is not possible to remove an existing module, function or limitation from a P
...
+    protected function configureContainer(ContainerBuilder $container, LoaderInterface $loader): void
+    {
         // ...
- 
+        
         // Retrieve "ezpublish" container extension
         $eZExtension = $container->getExtension('ezpublish');
         // Add the policy provider
-        $eZExtension->addPolicyProvider(new MyPolicyProvider());
+        $eZExtension->addPolicyProvider(new AppPolicyProvider());
```
micszo, line 117: It looks like now it's inconsistent, L43 & L69 still use "MyPolicyProvider".

docs/guide/custom_policies.md
```diff
@@ -122,11 +123,10 @@ class AcmeExampleBundle extends Bundle
 
 To provide support for editing custom policies in the Back Office, you need to implement [`EzSystems\EzPlatformAdminUi\Limitation\LimitationFormMapperInterface`](https://github.com/ezsystems/ezplatform-admin-ui/blob/master/src/lib/Limitation/LimitationFormMapperInterface.php).
```
micszo, line 124: Some example code here would be really useful. 🙂 It's rather a general statement.

docs/guide/custom_policies.md
```diff
@@ -135,11 +135,10 @@ app.security.limitation.custom_limitation.mapper:
 
 If you want to provide human-readable names of the custom Limitation values, you need to implement [`EzSystems\EzPlatformAdminUi\Limitation\LimitationValueMapperInterface`](https://github.com/ezsystems/ezplatform-admin-ui/blob/master/src/lib/Limitation/LimitationValueMapperInterface.php).
```
micszo, line 136: Here also some example would be useful.

mikadamczyk (commented):
It would be great to add information to the doc about how to create custom limitation e.g. CustomLimitationType.php and CustomLimitation.php classes with sample configuration 
```
    Security\Limitation\CustomLimitationType:
        tags:
            - {name: ezpublish.limitationType, alias: CustomLimitation}
``` 
to make it more clear

docs/guide/custom_policies.md
rhidja: @micszo done!

docs/guide/custom_policies.md
rhidja: @micszo it is too long to do it, i sugest to merge first this PR, to correct the current documentation. After i think it is better to split this chapter, and create an other one for custom limitations, to explain how to create CustomLimitation, CustomLimitationType,  LimitationFormMapperInterface, imitationValueMapperInterface.

docs/guide/custom_policies.md
rhidja: @micszo same thing too long to do it.

docs/guide/custom_policies.md
micszo: Yes, internally we also came to the conclusion that this will have to be tackled separately. \cc @dabrt

micszo (approved):
LGTM. 👍 

Let's handle the big issues separately.

docs/guide/custom_policies.md
dabrt: A ticket was created to handle your suggestions to this article:
https://issues.ibexa.co/browse/EZP-32155
Should you have any more comments, you can add them directly there.

micszo: Thanks!

PR #1142, opened 2020-09-21
EZP-31912: [2.5] Removed legacy references from File Management doc and fixed paths

alongosz:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | [EZP-31912](https://jira.ez.no/browse/EZP-31912)
| Relates to | ezsystems/ezplatform-kernel#110
| Versions      | 2.5, for 3.x see (#1148)

Those are the doc issues I've found so far while investigating the mentioned JIRA tickets. Let's wait with review until all JIRA issues are fixed because there might be an update here.

Note: there should be also separate doc PR for 3.1, related to [EZP-31740](https://jira.ez.no/browse/EZP-31740), about feature changes due to config not working anymore, as described in ezsystems/ezplatform#596

TODO:
- [x] revise updated static server config suggestion with respect to security

### DIFF
--- a/docs/guide/file_management.md
+++ b/docs/guide/file_management.md
@@ -41,7 +41,7 @@ The 'default' Flysystem adapter's directory is based on your site settings, and 
 
 !!! note
 
-    When Legacy Bridge is enabled the path will be automatically set to
+    When Legacy Bridge is enabled, the path will be automatically set to
     `%ezpublish_legacy.root_dir%/$var_dir$/$storage_dir$` instead.
 
 #### Configure the permissions of generated files
@@ -142,7 +142,8 @@ For [more information about REST API see the documentation](../api/rest_api_guid
 
 ### IO URL decoration
 
-By default, images and binary files referenced by content will be served from the same server as the application, for example `/var/site/storage/images/3/6/4/6/6463-1-eng-GB/kidding.png`. This is the default semantic configuration:
+By default, images and binary files that are referenced by the content will be served from the same server as the application, for example `/var/site/storage/images/3/6/4/6/6463-1-eng-GB/kidding.png`. 
+This is the default semantic configuration:
 
 ``` yaml
 ezpublish:
@@ -158,7 +159,8 @@ ezpublish:
 
 One common use case is to use an optimized nginx to serve images in an optimized way. The example image
 above could be made available as `http://static.example.com/var/site/storage/images/3/6/4/6/6463-1-eng-GB/kidding.png`
-by setting up a separate server that maps `/path/to/ezplatform/web/var` directory. The configuration would be as follows:
+by setting up a separate server that maps the `/path/to/ezplatform/web/var` directory. 
+The configuration would be as follows:
 
 ``` yaml
 ezpublish:
@@ -170,8 +172,9 @@ ezpublish:
 
 !!! caution
 
-    For security reasons it is recommended not to map `/path/to/ezplatform/web/` as a static server
-    Document Root. `/var/` directory should be mapped to `/path/to/ezplatform/web/var` directly instead.
+    For security reasons, do not map `/path/to/ezplatform/web/` as
+    Document Root of the static server. 
+    Map the `/var/` directory directly to `/path/to/ezplatform/web/var` instead.
 
 
 !!! note


docs/guide/file_management.md
```diff
@@ -37,7 +37,7 @@ ez_io:
                 adapter: default
 ```
 
-The 'default' Flysystem adapter's directory is based on your site settings, and will automatically be set to `%ezpublish_legacy.root_dir%/$var_dir$/$storage_dir$` (for example: `/path/to/ezpublish_legacy/var/ezdemo_site/storage`).
+The 'default' Flysystem adapter's directory is based on your site settings, and will automatically be set to `%webroot_dir%/$var_dir$/$storage_dir$` (for example: `/path/to/ezplatform/web/var/site/storage`).
```
andrerom, line 40: The original path will still be automatically correct if legacy bridge is installed, afaik.

alongosz: > The original path will still be automatically correct if legacy bridge is installed, afaik.

Hmm, ok. I can make a note here. Because w/o LB this piece of the doc is incorrect.

alongosz: > > The original path will still be automatically correct if legacy bridge is installed, afaik.
> 
> Hmm, ok. I can make a note here. Because w/o LB this piece of the doc is incorrect.

Added in relevant places via 2455169. Dropped /var/site/storage (or /var/ezdemo_site/storage) suffix because the doc root in this case needs to be set to the web root so the given examples make sense.

alongosz:
@ezsystems/documentation-team @ezsystems/php-dev-team this is ready for a review. I've attached the PR to the latest IO URL prefix fix JIRA issue, since it mostly is related to that.

I felt static server configuration was misleading by suggesting that `web/var/site/storage` should be mapped as Document Root of a static server. In fact the path which is generated starts with `/var` so the mapping should correspond rather to a public directory of the installation. 

This is clearly a security issue because static server would be able to serve front controllers as well. Therefore I've added a note explaining that var dir should be mapped directly to `/var` with respect to static server Doc Root.

Please let me know if this needs better explaining or is it clear to everyone, not just to me ;)

security note update cc @glye

docs/guide/file_management.md
```diff
@@ -37,7 +37,12 @@ ez_io:
...
+
+!!! note
+
+    When Legacy Bridge is enabled the path will be automatically set to
```
dabrt, line 44:
```suggestion
    When Legacy Bridge is enabled, the path will be automatically set to
```

docs/guide/file_management.md
```diff
@@ -137,7 +142,7 @@ For [more information about REST API see the documentation](../api/rest_api_guid
 
 ### IO URL decoration
 
-By default, images and binary files referenced by content will be served from the same server as the application, for example `/var/ezdemo_site/storage/images/3/6/4/6/6463-1-eng-GB/kidding.png`. This is the default semantic configuration:
+By default, images and binary files referenced by content will be served from the same server as the application, for example `/var/site/storage/images/3/6/4/6/6463-1-eng-GB/kidding.png`. This is the default semantic configuration:
```
dabrt, line 145:
```suggestion
By default, images and binary files that are referenced by the content will be served from the same server as the application, for example `/var/site/storage/images/3/6/4/6/6463-1-eng-GB/kidding.png`. 
This is the default semantic configuration:
```

docs/guide/file_management.md
```diff
@@ -151,7 +156,9 @@ ezpublish:
 
 ### Using a static server for images
 
-One common use case is to use an optimized nginx to serve images in an optimized way. The example image above could be made available as `http://static.example.com/images/3/6/4/6/6463-1-eng-GB/kidding.png` by setting up a server that uses `ezpublish/ezpublish_legacy/var/ezdemo_site/storage`. The configuration would be as follows:
+One common use case is to use an optimized nginx to serve images in an optimized way. The example image
+above could be made available as `http://static.example.com/var/site/storage/images/3/6/4/6/6463-1-eng-GB/kidding.png`
+by setting up a separate server that maps `/path/to/ezplatform/web/var` directory. The configuration would be as follows:
```
dabrt, line 161:
```suggestion
by setting up a separate server that maps the `/path/to/ezplatform/web/var` directory. 
The configuration would be as follows:
```

docs/guide/file_management.md
```diff
@@ -161,14 +168,25 @@ ezpublish:
                 url_prefix: 'http://static.example.com/$var_dir$/$storage_dir$'
 ```
 
+!!! caution
+
+    For security reasons it is recommended not to map `/path/to/ezplatform/web/` as a static server
```
dabrt, line 173:
```suggestion
    For security reasons, do not map `/path/to/ezplatform/web/` as
```

docs/guide/file_management.md
```diff
@@ -161,14 +168,25 @@ ezpublish:
...
+    Document Root. `/var/` directory should be mapped to `/path/to/ezplatform/web/var` directly instead.
```
dabrt, line 174:
```suggestion
    Document Root of the static server. 
    Map the `/var/` directory directly to `/path/to/ezplatform/web/var` instead.
```

dabrt (changes requested):

glye (approved):
LGTM. I have a more general concern about static serving, see protected issue https://jira.ez.no/browse/EZP-31150 - it isn't directly related to the PR, but affects the same doc.

alongosz:
> LGTM. I have a more general concern about static serving, see protected issue https://jira.ez.no/browse/EZP-31150 - it isn't directly related to the PR, but affects the same doc.

Hmm, indeed it's the way it was designed to work (probably before the era where where those issues started to matter). @glye it's best that you discuss this with PMs, because if to be implemented, quite complicated feature. I mean serving protected content is moderately easy, but scaling it so it's performs with our perm system is another story. Or we could document the behavior separately ;)

alongosz:
@dabrt thx for the review, committed the changes.

dabrt (approved):

andrerom (approved):

kaff (approved):

PR #1150, opened 2020-09-24
added missing information about adding styles for aligning images in rich text

mateuszdebinski:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | [EZP-31790](https://jira.ez.no/browse/EZP-31790)
| Versions      | 1.13 and higher

### DIFF
--- a/docs/guide/templates.md
+++ b/docs/guide/templates.md
@@ -207,9 +207,10 @@ See [Custom rendering logic](controllers.md#custom-rendering-logic) for more in
 ## Rendering Rich Text
 
 ### Alignment of images
-Images that have been added to rich text and has set alignment, have one of three class: align-left, align-right, or align-center. By default, these classes have no implementation on the Front, so for the alignment to work properly, you need to implement them in your styles. 
 
-Below is an example styles of classes implementation for alignment
+Images that have been added to Rich Text with alignment set have one of following classes: `align-left`, `align-right`, or `align-center`.
+
+By default, these classes have no specific styling, so for the alignment to work properly, you need to add them to your styles, for example:
 
 ``` css
 .align-left {


andrerom (approved):

andrerom:
LGTM.
---
For future version, might there be a way for us to inject CSS on front ala how pagebuilder is injecting JS on front? @dew326 
If so there should be a way to opt out though, just would be good incase we need to provide standard RichText styling for reuse on front.

dew326:
I think Page Builder is injecting only into the iframe when the preview is shown in the admin interface. 
I don't think we include any CSS/JS in the front, well I hope.

docs/guide/templates.md
```diff
@@ -204,6 +204,30 @@ While it is possible to template a whole website using only Twig, a custom PHP c
 
 See [Custom rendering logic](controllers.md#custom-rendering-logic) for more information.
 
+## Rendering Rich Text
+
+### Alignment of images
+Images that have been added to rich text and has set alignment, have one of three class: align-left, align-right, or align-center. By default, these classes have no implementation on the Front, so for the alignment to work properly, you need to implement them in your styles.
```
DominikaK, line 210:
```suggestion
### Alignment of images

Images that have been added to Rich Text with alignment set have one of following classes: `align-left`, `align-right`, or `align-center`.
By default, these classes have no specific styling, so for the alignment to work properly, you need to add them to your styles, for example:
```

docs/guide/templates.md
```diff
@@ -204,6 +204,30 @@ While it is possible to template a whole website using only Twig, a custom PHP c
 
 See [Custom rendering logic](controllers.md#custom-rendering-logic) for more information.
 
+## Rendering Rich Text
+
+### Alignment of images
+Images that have been added to rich text and has set alignment, have one of three class: align-left, align-right, or align-center. By default, these classes have no implementation on the Front, so for the alignment to work properly, you need to implement them in your styles. 
+
+Below is an example styles of classes implementation for alignment
```
DominikaK, line 212:
```suggestion
```

DominikaK (changes requested):

DominikaK (approved):

PR #1153, opened 2020-10-01
Fix Solr boost caution to be correct

andrerom:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   |  N/A
| Versions      | 2.5+

The caution was not clear enough and was contradictive, as it made no sense to then have this doc if boosting where not at all supported by the bundle.

### DIFF
--- a/docs/guide/search/solr.md
+++ b/docs/guide/search/solr.md
@@ -307,7 +307,8 @@ Here are the most common issues you may encounter:
 
 !!! caution "Index time boosting"
 
-    Index time boosting was deprecated in Solr 6.5 and removed in Solr 7.0. Until query time boosting is implemented there is no way to boost currently in the bundle out of the box.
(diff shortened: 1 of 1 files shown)


docs/guide/search/solr.md
```diff
@@ -307,7 +307,7 @@ Here are the most common issues you may encounter:
 
 !!! caution "Index time boosting"
 
-    Index time boosting is not available in the Solr Search Engine Bundle.
+    Index time boosting was deprecated in Solr 6.5 and removed in Solr 7.0. Until query time boosting is implemented there is no way to boost currently in the bundle out of the box.
```
DominikaK, line 310:
```suggestion
    Index time boosting was deprecated in Solr 6.5 and removed in Solr 7.0.
    Until query time boosting is implemented, there is no way to boost in the bundle out of the box.
```

DominikaK (approved):

PR #1155, opened 2020-10-01
Fix link to cron instructions for Link Manager

Steveb-p:
This fixes the hash part of the link to match the "id" attribute on the linked page.

Btw, why is the Link Manager even mentioned at this point? I do feel like it doesn't really belong there, as a casual user I would still be just setting up the platform, so this particular note feels like it's pointing to something that I should eventually do, but at this point it's kinda distracting.

https://doc.ezplatform.com/en/latest/getting_started/install_ez_platform/#install-ez-platform_1
image

### DIFF
--- a/docs/getting_started/install_ez_platform.md
+++ b/docs/getting_started/install_ez_platform.md
@@ -178,7 +178,7 @@ This operation is performed only once when you install eZ Platform Enterprise Ed
 
 !!! tip "Enabling Link manager"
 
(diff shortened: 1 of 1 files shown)


DominikaK (approved):

DominikaK:
The order of steps, including the Link Manager note, will be changed in this PR: https://github.com/ezsystems/developer-documentation/pull/1151

PR #1187, opened 2020-10-20
Security advice for Symfony app secret

glye:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | N/A
| Versions      | 1.13 and newer

The Symfony app secret is crucial to security. This is _somewhat urgent_, it would be good to have it released before 3.2 stable, when a new wave of installs is coming.

- "up to 32 characters" is wrong. [Symfony says "around 32"](https://symfony.com/doc/current/reference/configuration/framework.html#secret), which means we should consider that the minimum, and longer may be a good idea.
- Added caution box, syntax not verified.
- The same change should be done in [installing-on-mac-os-and-windows.md](https://github.com/ezsystems/developer-documentation/blob/7584ebbe08e5239ce3c65de372afcf3fc211580f/docs/community_resources/installing-on-mac-os-and-windows.md#change-installation-parameters) and any other relevant places.
- In later versions, at least 3.x, there's a Tip box about storing db credentials in .env.local, not in version control. The same applies to the app secret. This can be merged into the caution box, as that's also a security concern.

Heads up, I will propose a new doc page soon called Security Checklist, where we can expand more on things like this. But this PR should not wait until that is ready.

### DIFF
--- a/docs/community_resources/installing-on-mac-os-and-windows.md
+++ b/docs/community_resources/installing-on-mac-os-and-windows.md
@@ -99,10 +99,23 @@ composer install
 
 After a moment the installer will ask you to provide a few parameters:
 
-1. Choose a [secret](https://symfony.com/doc/2.8/reference/configuration/framework.html#secret). It should be a random string consisting of up to 32 characters, numbers, and symbols. This is used by Symfony for generating [CSRF tokens](https://symfony.com/doc/2.8/security/csrf_in_login_form.html), [encrypting cookies](https://symfony.com/doc/2.8/security/remember_me.html), and for creating signed URIs when using [ESI (Edge Side Includes)](https://symfony.com/doc/2.8/http_cache/esi.html).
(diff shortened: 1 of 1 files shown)


docs/getting_started/install_ez_platform.md
```diff
@@ -105,11 +105,19 @@ composer create-project --no-dev --keep-vcs ezsystems/ezplatform .
...
 1. Select a `database_name` or accept the default one.
 1. Provide your `database_user` and `database_password`.
 
+!!! caution
+
+    The app secret is crucial to the security of your installation. Be careful about how you generate it, and how you store it. Here's one way to generate a 64 characters long, secure random string as your secret, in PHP:
+    ``` php
+    print bin2hex(random_bytes(32));
+    ```
+    Make sure you do not commit the secret to version control systems, or share it with anyone who does not strictly need it. If you have any suspicion that it may have been exposed, replace it with a new secret. The same goes for other secrets, like database password, Varnish invalidate token, JWT passphrase, etc.
```
DominikaK, line 119:
```suggestion
    The app secret is crucial to the security of your installation. Be careful about how you generate it, and how you store it.
    Here's one way to generate a 64 characters long, secure random string as your secret, in PHP:
    
    ``` php
    print bin2hex(random_bytes(32));
    ```

    Do not commit the secret to version control systems, or share it with anyone who does not strictly need it.
    If you have any suspicion that the secret may have been exposed, replace it with a new one.
    The same goes for other secrets, like database password, Varnish invalidate token, JWT passphrase, etc.
```

DominikaK (approved):

docs/getting_started/install_ez_platform.md
glye: 👍

mateuszdebinski (approved):

andrerom (approved):

PR #1191, opened 2020-10-21
Use only satis hostname, not path

glye:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | N/A
| Versions      | All supported where this is included in doc

https://doc.ibexa.co/en/3.2/getting_started/install_ez_platform/#set-up-authentication-tokens
_"Ibexa DXP and Ibexa Commerce subscribers have access to commercial packages at..."_

The text is not quite correct, since Commerce packages are located at updates.ez.no/bul_com/. If we only show the hostname it's correct, and simpler than showing two urls.

Another question is if we should also link to https://updates.ibexa.co/ here, but that's for later I think. CC @andrerom

### DIFF
--- a/docs/getting_started/install_ez_platform.md
+++ b/docs/getting_started/install_ez_platform.md
@@ -56,7 +56,7 @@ composer -V
 
     ### Set up authentication tokens
 
-    [[= product_name_ee =]] and [[= product_name_com =]] subscribers have access to commercial packages at [updates.ez.no/bul/](https://updates.ez.no/bul/).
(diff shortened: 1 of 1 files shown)


DominikaK:
@glye, I've had to modify branch history, can you please rebase this PR against the current 3.2 branch?

andrerom:
> Another question is if we should also link to https://updates.ibexa.co/ here, but that's for later I think. CC @andrerom

not yet, this is planned for 3.3 but we still need to prepare it for use

DominikaK (approved):

PR #1194, opened 2020-10-21
Update available password hashes

glye:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | N/A
| Versions      | All supported versions

Updates the list of password hashes we support. Adds the "new" ones, and marks the old as deprecated/removed in given versions.

Doc: https://doc.ibexa.co/en/latest/api/field_type_reference/#php-api-field-type_23
Deprecation notice: https://doc.ibexa.co/en/latest/releases/ez_platform_v3.0_deprecations/#password-hashes
Source: https://github.com/ezsystems/ezpublish-kernel/blob/18c6d271f194a95fbb81ca56fd63e59593547ff3/eZ/Publish/API/Repository/Values/User/User.php#L54

### DIFF
--- a/docs/api/field_type_reference.md
+++ b/docs/api/field_type_reference.md
@@ -2257,9 +2257,12 @@ This Field Type validates and stores information about a user.
 
 !!! caution
 
-    Using the MD5-based deprecated hash types is a security risk, because if the hashes are leaked, they are too easily broken by brute force attacks. The plaintext type offers no security. It was only ever intended for testing, and should never be used now.
+    Using the MD5-based deprecated hash types is a security risk, because if the hashes are leaked, they are too easily broken by brute-force attacks.
+    The plaintext type offers no security. It was only ever intended for testing, and should never be used now.
 
-    We strongly recommend switching to one of the new hash types. If you do, it will be used for new users. Existing users will also have their hashes updated to the new type when they log in. (A mass update of all hashes is not possible, because this requires knowing the passwords, which only the users themselves do.)
+    We strongly recommend switching to one of the new hash types. If you do, it will be used for new users.
+    Existing users will also have their hashes updated to the new type when they log in.
+    (A mass update of all hashes is not possible, because this requires knowing the passwords, which only the users themselves do.)
 
     Removal notice: https://doc.ibexa.co/en/latest/releases/ez_platform_v3.0_deprecations/#password-hashes
 


docs/api/field_type_reference.md
```diff
@@ -2240,17 +2240,20 @@ This Field Type validates and stores information about a user.
...
-|`eZ\Publish\Core\Repository\Values\User\User::PASSWORD_HASH_MD5_USER`|MD5 hash of the password and username.|
-|`eZ\Publish\Core\Repository\Values\User\User::PASSWORD_HASH_MD5_SITE`|MD5 hash of the password, username and site name.|
-|`eZ\Publish\Core\Repository\Values\User\User::PASSWORD_HASH_PLAINTEXT`|Passwords are stored in plaintext, should not be used for real sites.|
+|`eZ\Publish\Core\Repository\Values\User\User::DEFAULT_PASSWORD_HASH`|Default password hash, used when none is specified, may change over time.|
+|`eZ\Publish\Core\Repository\Values\User\User::PASSWORD_HASH_PHP_DEFAULT`|Passwords hashed by PHP's default algorithm, which may change over time.|
+|`eZ\Publish\Core\Repository\Values\User\User::PASSWORD_HASH_BCRYPT`|Bcrypt hash of the password.|
+|`eZ\Publish\Core\Repository\Values\User\User::PASSWORD_HASH_MD5_PASSWORD`|MD5 hash of the password, not recommended. Deprecated in v1.13, removed in v3.0.|
+|`eZ\Publish\Core\Repository\Values\User\User::PASSWORD_HASH_MD5_USER`|MD5 hash of the password and username. Deprecated in v1.13, removed in v3.0.|
```
DominikaK, line 2254:
```suggestion
|`eZ\Publish\Core\Repository\Values\User\User::PASSWORD_HASH_MD5_USER`|MD5 hash of the password and username. Deprecated.|
```

docs/api/field_type_reference.md
```diff
@@ -2240,17 +2240,20 @@ This Field Type validates and stores information about a user.
...
+|`eZ\Publish\Core\Repository\Values\User\User::PASSWORD_HASH_MD5_SITE`|MD5 hash of the password, username and site name. Deprecated in v1.13, removed in v3.0.|
```
DominikaK, line 2255:
```suggestion
|`eZ\Publish\Core\Repository\Values\User\User::PASSWORD_HASH_MD5_SITE`|MD5 hash of the password, username and site name. Deprecated.|
```

docs/api/field_type_reference.md
```diff
@@ -2240,17 +2240,20 @@ This Field Type validates and stores information about a user.
...
+|`eZ\Publish\Core\Repository\Values\User\User::PASSWORD_HASH_PLAINTEXT`|Passwords are stored in plaintext, should not be used for real sites. Deprecated in v1.13, removed in v3.0.|
```
DominikaK, line 2256:
```suggestion
|`eZ\Publish\Core\Repository\Values\User\User::PASSWORD_HASH_PLAINTEXT`|Passwords are stored in plaintext, should not be used for real sites. Deprecated.|
```
Here and in the two cases above we can jut remove the line in branches for v3

docs/api/field_type_reference.md
```diff
@@ -2240,17 +2240,20 @@ This Field Type validates and stores information about a user.
...
 ###### Available password hash types
 
 |Constant|Description|
 |------|------|
-|`eZ\Publish\Core\Repository\Values\User\User::PASSWORD_HASH_MD5_PASSWORD`|MD5 hash of the password, not recommended.|
-|`eZ\Publish\Core\Repository\Values\User\User::PASSWORD_HASH_MD5_USER`|MD5 hash of the password and username.|
-|`eZ\Publish\Core\Repository\Values\User\User::PASSWORD_HASH_MD5_SITE`|MD5 hash of the password, username and site name.|
-|`eZ\Publish\Core\Repository\Values\User\User::PASSWORD_HASH_PLAINTEXT`|Passwords are stored in plaintext, should not be used for real sites.|
+|`eZ\Publish\Core\Repository\Values\User\User::DEFAULT_PASSWORD_HASH`|Default password hash, used when none is specified, may change over time.|
+|`eZ\Publish\Core\Repository\Values\User\User::PASSWORD_HASH_PHP_DEFAULT`|Passwords hashed by PHP's default algorithm, which may change over time.|
```
DominikaK, line 2251: Just to make sure, you mean that the *passwords* may change overtime, not the *algorithm*? In that case:

```suggestion
|`eZ\Publish\Core\Repository\Values\User\User::PASSWORD_HASH_PHP_DEFAULT`|Passwords hashed by PHP's default algorithm. The passwords may change over time.|
```

glye: No, the algorithm may change. Passwords too, but that's not the point 😊

docs/api/field_type_reference.md
glye: Completely remove the lines in v3 doc, agree.
I'd say it can useful for a v1 / v2 dev to know when exactly these were deprecated, and it's common practice in the code itself to include it.

glye: It might even be wise to add an alert box here, to make it very clear that using these deprecated hash types is a security risk.

glye: Added the warning box in new commit.

docs/api/field_type_reference.md
```diff
@@ -2240,17 +2240,28 @@ This Field Type validates and stores information about a user.
...
+|`eZ\Publish\Core\Repository\Values\User\User::PASSWORD_HASH_MD5_USER`|MD5 hash of the password and username. Deprecated in v1.13, removed in v3.0.|
+|`eZ\Publish\Core\Repository\Values\User\User::PASSWORD_HASH_MD5_SITE`|MD5 hash of the password, username and site name. Deprecated in v1.13, removed in v3.0.|
+|`eZ\Publish\Core\Repository\Values\User\User::PASSWORD_HASH_PLAINTEXT`|Passwords are stored in plaintext, should not be used for real sites. Deprecated in v1.13, removed in v3.0.|
+
+!!! caution
+
+    Using the MD5-based deprecated hash types is a security risk, because if the hashes are leaked, they are too easily broken by brute force attacks. The plaintext type offers no security. It was only ever intended for testing, and should never be used now.
+
+    We strongly recommend switching to one of the new hash types. If you do, it will be used for new users. Existing users will also have their hashes updated to the new type when they log in. (A mass update of all hashes is not possible, because this requires knowing the passwords, which only the users themselves do.)
```
DominikaK, line 2262:
```suggestion
    Using the MD5-based deprecated hash types is a security risk, because if the hashes are leaked, they are too easily broken by brute-force attacks.
    The plaintext type offers no security. It was only ever intended for testing, and should never be used now.

    We strongly recommend switching to one of the new hash types. If you do, it will be used for new users.
    Existing users will also have their hashes updated to the new type when they log in.
    (A mass update of all hashes is not possible, because this requires knowing the passwords, which only the users themselves do.)
```

DominikaK (approved):

PR #1197, opened 2020-10-23
Update doc to reflect that Solr8 is supported

andrerom:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | N/A
| Versions      | 3.x, 2.5 _(new there, needs QA)_

Recent request for Solr 8, however technically it's supported.
It's just not recommended _yet_, as LTS ([8.8](https://cwiki.apache.org/confluence/display/SOLR/Roadmap)) is not out yet and supported by Platform.sh.

Solr 8 support:
- Code (3.x): https://github.com/ezsystems/ezplatform-solr-search-engine/pull/187
- Platform.sh (8.6): https://docs.platform.sh/configuration/services/solr.html

### DIFF
--- a/docs/guide/search/solr.md
+++ b/docs/guide/search/solr.md
@@ -26,7 +26,7 @@ The example presents a configuration with a single core. For configuring Solr in
 
 ##### Solr 7
 
-Download and extract Solr *(here using 7.7.2 as an example)*:
+Download and extract Solr, for example, in version 7.7.2:
 
 - [solr-7.7.2.tgz](http://archive.apache.org/dist/lucene/solr/7.7.2/solr-7.7.2.tgz) or [solr-7.7.2.zip](http://archive.apache.org/dist/lucene/solr/7.7.2/solr-7.7.2.zip)
 


konradoboza (approved):

andrerom:
Note to doc team, might be this should be applied to 2.5 also if this is Tested and merged before this is merged: 
https://github.com/ezsystems/ezplatform-solr-search-engine/pull/189

docs/getting_started/requirements.md
```diff
@@ -56,7 +56,7 @@ For security and performance we generally recommend (unless otherwise noted) usi
         - Redis 4.0+ (5.0 recommended, using `volatile-*` [eviction policy](https://redis.io/topics/lru-cache) is required with default [Redis adapter](../guide/persistence_cache.md#redis))
         - Memcached 1.5 or higher (See [Memcached adapter](../guide/persistence_cache.md##memcached) for comparison with Redis)
     - Session: either own Redis instance with persistence turned on, or Database.
-    - Search: Solr 7 or Elasticsearch 7.7 (recommended over SQL-based Search engine, especially on cluster, as SQL does not provide the same feature set or performance as the other two).
+    - Search: Solr 7/8 or Elasticsearch 7.7 (recommended over SQL-based Search engine, especially on cluster, as SQL does not provide the same feature set or performance as the other two).
```
dabrt, line 59:
```suggestion
    - Search: Solr 7, Solr 8, or Elasticsearch 7.7 (recommended over SQL-based Search engine, especially on cluster, as SQL does not provide the same feature set or performance as the other two).
```
Has Solr 8 been tested with Ibexa Platform?

docs/guide/search/solr.md
```diff
@@ -4,15 +4,15 @@
...
+    If you have previously removed the bundle, add/update composer dependencies:
     ``` bash
-    composer require --no-update ezsystems/ezplatform-solr-search-engine:~2.0
+    composer require --no-update ezsystems/ezplatform-solr-search-engine:~3.0
     composer update
     ```
-
-    Make sure `EzPublishSolrSearchEngineBundle` is activated with the following line in the `app/AppKernel.php` file: `new EzSystems\EzPlatformSolrSearchEngineBundle\EzSystemsEzPlatformSolrSearchEngineBundle()`
+    
+    Symfony Flex should enable the bundle for you when installing the package.
```
dabrt, line 15:
```suggestion
    Symfony Flex should enable the bundle for you when installing the package.
```
"Should enable" or "will enable"? 

If "should", I would mention the circumstances that can prevent Symfony Flex from enabling the bundle, and if/how one can enable Solr manually.

docs/guide/search/solr.md
```diff
@@ -22,12 +22,11 @@ The example presents a configuration with a single core. For configuring Solr in
 
 !!! note "Solr versions"
 
-    `ezplatform-solr-search-engine` v1 still supports Solr 6.
-    Since v2 `ezplatform-solr-search-engine` works only with Solr 7.
+    For v3 `ezplatform-solr-search-engine` works with Solr 7 and 8, using an supported LTS versions of Solr is recommended.
```
dabrt, line 25:
```suggestion
    v3 of `ezplatform-solr-search-engine` works with Solr 7 and Solr 8.
    It is recommended that you use a current major version of Solr.
```

Lucene does not refer to its versions as "LTS". They use terms like major, previous major, or EOL. Alternatively, we could say "It is recommended that you do not use an EOL version of Solr."

docs/guide/search/solr.md
```diff
@@ -22,12 +22,11 @@ The example presents a configuration with a single core. For configuring Solr in
...
 
 ##### Solr 7
 
-Download and extract Solr. Solr Bundle 2 and higher supports Solr 7 *(currently tested and recommended with Solr 7.7)*:
+Download and extract Solr *(currently recommended with Solr 7.7LTS, so that is what example here is showing)*:
```
dabrt, line 29:
```suggestion
Download and extract Solr *(a recommended major version is Solr 7.7.3)*:
```
7.7.3 is the last stable version of the 7.x series. 

To avoid confusion, I would use either this, or another tested and approved version, but give the exact version number, as assigned by Apache

dabrt (changes requested):

docs/getting_started/requirements.md
andrerom: Unsure, it's part of the integration tests for some times but for manual tests you'll need to ask @ezsystems/qa-team

docs/guide/search/solr.md
andrerom: The thing behind the recommendation:
1. 7.7 is given bug fixes still, while 8.x receives new features and sometimes possible breaks in minor releases
2. Platform.sh [supports 7.7, and 8.6](https://docs.platform.sh/configuration/services/solr.html), however they probably will take some time to add 8.7 and 8.8 when it arrives, so we try to avoid this kind of friction.

But maybe we should rather be specific to recommend versions:
- We test on
- available on p.sh

That would mean either 7.7 or 8.6 right now, and then we update it once 8.8 is out and available on psh.

docs/guide/search/solr.md
andrerom: Major would refer to 7, so if we want to be specific patch release and say `7.7.3` I would maybe rather phrase it `*(7.7.3 + 8.6.3 tested and recommended at time of writing)*` But you end up having to update it quite often though, which is why I don't mention patch versions in doc usually but rather talk about which branches we support / recommend.

docs/guide/search/solr.md
dabrt: "That would mean 7.7 or 8.6 right now and then we update it once 8.8 is out and available on psh." - makes perfect sense to me

docs/guide/search/solr.md
dabrt: "at the time of writing" is assumed, we do not have to state that.

Can we say  *(the most recent release in 7.7 or 8.6 series is recommended)*

docs/guide/search/solr.md
andrerom: I didn't have the time to test and verify so that is why I as unsure. If anyone has install, remove solr, and try to require it to see what flex does. Or we omit the omit should, and take it as a bug if it turns out to be wrong 😉

docs/guide/search/solr.md
andrerom:
```suggestion
    For v3 `ezplatform-solr-search-engine` works with Solr 7 and 8, using 7.7 or 8.6 of Solr is currently recommended.
```

docs/guide/search/solr.md
andrerom:
```suggestion
Download and extract Solr *(the most recent release in 7.7 or 8.6 series is recommended)*:
```

docs/guide/search/solr.md
andrerom:
```suggestion
    Symfony Flex will enable the bundle for you when installing the package.
```

docs/guide/search/solr.md
```diff
@@ -22,12 +22,11 @@ The example presents a configuration with a single core. For configuring Solr in
...
 !!! note "Solr versions"
 
-    `ezplatform-solr-search-engine` v1 still supports Solr 6.
-    Since v2 `ezplatform-solr-search-engine` works only with Solr 7.
+    For v3 `ezplatform-solr-search-engine` works with Solr 7 and 8, using most recent version of Solr 7.7 or 8.6 is recommended.
 
 ##### Solr 7
 
-Download and extract Solr. Solr Bundle 2 and higher supports Solr 7 *(currently tested and recommended with Solr 7.7)*:
+Download and extract Solr *(here using 7.7.2 as an example)*:
```
dabrt, line 29:
```suggestion
Download and extract Solr, for example, in version 7.7.2:
```

dabrt (approved):

docs/guide/search/solr.md
```diff
@@ -4,15 +4,15 @@
...
+    If you have previously removed the bundle, add/update composer dependencies:
     ``` bash
-    composer require --no-update ezsystems/ezplatform-solr-search-engine:~2.0
+    composer require --no-update ezsystems/ezplatform-solr-search-engine:~3.0
     composer update
     ```
-
-    Make sure `EzPublishSolrSearchEngineBundle` is activated with the following line in the `app/AppKernel.php` file: `new EzSystems\EzPlatformSolrSearchEngineBundle\EzSystemsEzPlatformSolrSearchEngineBundle()`
+    
+    Symfony Flex will enable the bundle for you when installing the package.
```
andrerom, line 15: @dabrt FYI reverted part of this for 2.5 in 7cabc6801085a1ef16ac2a24e1c274648cf8e100, I forgot to mention the changes like this was meant for 3.x only.

PR #1208, opened 2020-10-29
GraphQL example using array of clauses and ascending sorting

arfaram:
This example shows how to define a sorting array and reverse the search results in the ascending way

| Question      | Answer
| ------------- | ---
| Versions      | 3.x

### DIFF
--- a/docs/api/graphql_queries.md
+++ b/docs/api/graphql_queries.md
@@ -325,7 +325,7 @@ You can sort query results using `sortBy`:
 }
 ```
 
-You can use an array of clauses as well. To reverse the item list you have to add `_desc` after the clause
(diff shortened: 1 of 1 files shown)


docs/api/graphql_queries.md
```diff
@@ -325,6 +325,12 @@ You can sort query results using `sortBy`:
 }
 ```
 
+You can use an array of clauses as well. To reverse the item list you have to add `_desc` after the clause
```
DominikaK, line 328:
```suggestion
You can use an array of clauses as well. To reverse the item list, add `_desc` after the clause:
```

DominikaK (approved):

PR #1210, opened 2020-10-30
Remove MFU configuration

DominikaK:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | n/a
| Versions      | 2.5, v3, master

Removes mention of MFU configuration via yaml files until https://github.com/ezsystems/ezplatform-admin-ui/pull/582 is fixed.


dabrt (commented):
How about removing the following pieces of information that refer to MFU while at it?
https://doc.ibexa.co/en/latest/guide/permissions/#multi-file-upload
https://doc.ibexa.co/en/latest/extending/extending_multifile_upload

DominikaK:
This is a different case, MFU is functional and those two pages should work, it's only the yaml configuration that is broken ATM

dabrt (approved):

PR #1212, opened 2020-10-30
Clarify how default scope settings can affect all site-accesses, including admin

lionelakpagni:
Make this part of the doc clearer. The default will override front and back views. 
I think it's important to have a code that  shows the example with desired siteaccess

| Question      | Answer
| ------------- | ---
| JIRA Ticket   | 
| Versions      | 

Make this part of the doc clearer. The default value will override front and back views.

### DIFF
--- a/docs/guide/siteaccess.md
+++ b/docs/guide/siteaccess.md
@@ -193,9 +193,9 @@ ezpublish:
             var_dir: var/site
 ```
 
-Be aware that the `default` scope overrides both back and front views.
-For example, the following configuration overrides both the front template for articles
-and the template used in the Back Office:
+Be aware that the `default` scope concerns both back and front views.
+For example, the following configuration defines both the front template for articles
+and the template used in the Back Office, unless other templates are configured for specific a SiteAccess or SiteAccess group:
 
 ``` yaml
 ezpublish:
--- a/docs/guide/templates.md
+++ b/docs/guide/templates.md
@@ -60,7 +60,7 @@ ezpublish:
 This is what individual keys in the configuration mean:
 
 - `ezpublish` and `system` are obligatory at the start of any configuration file which defines views.
-- `site_group` defines the scope for which the configuration will be used. `site_group` is a SiteAccess group used out of the box for all front pages.
+- `site_group` defines the scope for which the configuration will be used. `site_group` is a SiteAccess group used out of the box for all front-end SiteAccesses.
 See [Scope](siteaccess.md#scope) for other available keys.            
 - `user` and `layout` point to the main template file that is used in any situation where no other template is defined. All other templates extend this one.
 - `content_view` defines the view provider.


docs/guide/templates.md
```diff
@@ -61,6 +61,23 @@ This is what individual keys in the configuration mean:
 
 - `ezpublish` and `system` are obligatory at the start of any configuration file which defines views.
 - `default` defines the SiteAccess for which the configuration will be used. "default", as the name suggests, determines what views are used when no other configuration is chosen. You can also have separate keys defining views for other SiteAccesses.
+Be aware that the `default` value will override back and front views. It may be replaced by the desired siteaccess. See example:
+
+``` yaml
+# Sample configuration file overriding your default siteaccess
+ezpublish:
+    system:
+        your_default_siteaccess:
```
andrerom, line 70: If we want to show what config is wrong, then I'd write it like:

```suggestion
Be aware that the `default` value will be used by both frontend and backend. It may be replaced by the desired siteaccess. See example:

``` yaml
# Sample configuration file overriding a file in default scope, used by all siteaccesses including admin:
ezpublish:
    system:
        default:
```

If the point is to show a better approach then maybe rather:
```suggestion
Be aware that the `default` value will override back and front views. If you are looking to apply to only frontend, use a shared siteaccess group for that. See example:

``` yaml
# Sample configuration file for setting content view matching for a shared siteaccess group:
ezpublish:
    system:
        frontend_group:
```


Besides this, @DominikaK maybe this needs to go into a caution block or similar to be better visible in the middle of this text?

DominikaK: I'm actually not sure what this example is supposed to illustrate. If the intention is to say: "use a specific SA instead of default", we don't need a full example, because the "Be aware" note will be enough.
What we can (and probably should) do is replace the default key in the main example above (line 39) with another scope or a `<siteaccess>` placeholder

lionelakpagni: @DominikaK Yes, the example is for that illustration you mentioned, but the reason why this is relevant on top of the note is that most of developers are used to copy paste codes examples blocs without taking the time to read details' subtleties.
Also, keeping both examples shows a global scope of the overriding system.
I suggest to keep the second suggestion from @andrerom

DominikaK: OK, so to avoid copy-pasting errors, it's even more important to make it an example that will not break the Back Office.

How about changing the example to refer to `site_group` (then we'll need to change the description to "`site_group` defines the SiteAccess for which the configuration will be used...", etc.)

While an example of using `default` should land in https://doc.ibexa.co/en/master/guide/siteaccess/#scope, where we already have an example for `global`.

andrerom: I'm with @DominikaK here, the whole section above should be updated, even the bullet point from before is actually wrong here: ``default` defines the SiteAccess for which the configuration will be used. (...)`, as it kind of implies default is a site-access when it's a config scope

Would you be able to give it a go @DominikaK? As you seem to have a clear idea how you'd like to present this ;)

lionelakpagni: Yes @DominikaK , this way sounds good to me. Would you mind to suggest it as a change, then I will commit it, or if you have a better way to proceed on this.

DominikaK: @lionelakpagni I couldn't comment with suggestion to a different page, so please see my last commit.

docs/guide/siteaccess.md
```diff
@@ -193,6 +193,22 @@ ezpublish:
             var_dir: var/site
 ```
 
+Be aware that the `default` scope overrides both back and front views.
+For example, the following configuration overrides both the front template for articles
+and the template used in the Back Office:
```
andrerom, line 198:
```suggestion
Be aware that the `default` scope is  used by both backend and frontend siteaccesses.
For example, the following configuration sets article template globally, and is used across all siteaccesses unless also set in a higher priority scope:
```


Afaik the config reading order is default, group, siteacess and then global or whatever it is called. If so `default` does not override anything, but it is set as baseline config implicit for all siteaccesses

docs/guide/templates.md
```diff
@@ -60,7 +60,8 @@ ezpublish:
 This is what individual keys in the configuration mean:
 
 - `ezpublish` and `system` are obligatory at the start of any configuration file which defines views.
-- `default` defines the SiteAccess for which the configuration will be used. "default", as the name suggests, determines what views are used when no other configuration is chosen. You can also have separate keys defining views for other SiteAccesses.
+- `site_group` defines the scope for which the configuration will be used. `site_group` is a SiteAccess group used out of the box for all front pages.
```
andrerom, line 63:
```suggestion
- `site_group` defines the scope for which the configuration will be used. `site_group` is a SiteAccess group used out of the box for all frontend siteacceses.
```

Just more specific terminology here

andrerom (approved):

mateuszbieniek (approved):

DominikaK (approved):

PR #1234, opened 2020-11-24
Security checklist: Refer also to dev guidelines

glye:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | N/A
| Versions      | All current

Ref https://ibexa.slack.com/archives/C0BJ6NGBT/p1606213049046100


glye:
AR suggests also cross linking.

DominikaK:
> AR suggests also cross linking.

glye:
Yep, and it makes sense, there is some overlapping subject matter.

DominikaK (approved):

PR #1239, opened 2020-11-30
Updated note about changing app secret

konradoboza:
I added a small note about actions that should be performed after changing the Symfony app secret.

Note: on every merge-up, the Symfony documentation URL should be adapted depending on the targetted eZ Platform/Ibexa DXP release and Symfony version that it's built upon. 

So for eZ Platform 2.5, it should be **3.4**: https://symfony.com/doc/3.4/reference/configuration/framework.html#secret etc.

### DIFF
--- a/docs/getting_started/install_ez_platform.md
+++ b/docs/getting_started/install_ez_platform.md
@@ -124,7 +124,7 @@ After a moment the installer will ask you to provide a few parameters:
     The same goes for other secrets, like database password, Varnish invalidate token, JWT passphrase, etc.
     
     After changing the app secret, make sure that you clear the application cache and log out all the users.
-    For more details refer to [Symfony documentation](https://symfony.com/doc/2.8/reference/configuration/framework.html#secret).
+    For more information, see [Symfony documentation](https://symfony.com/doc/2.8/reference/configuration/framework.html#secret).
 
 !!! tip
 


docs/getting_started/install_ez_platform.md
```diff
@@ -122,6 +122,9 @@ After a moment the installer will ask you to provide a few parameters:
     Do not commit the secret to version control systems, or share it with anyone who does not strictly need it.
     If you have any suspicion that the secret may have been exposed, replace it with a new one.
     The same goes for other secrets, like database password, Varnish invalidate token, JWT passphrase, etc.
+    
+    Remember, that after changing the app secret you should regenerate application cache and log out all the users.
+    For more details refer to [Symfony documentation](https://symfony.com/doc/2.8/reference/configuration/framework.html#secret).
```
glye, line 127: Devil's ~advocate~ support request: "How do I regenerate application cache and log out all the users?"

For the first, we probably have some doc. For the other, do we even have this functionality, or would you have to manually delete all sessions?

konradoboza: For the first, maybe rephrasing this to "clear cache" would do the trick. For the second: probably yes, I cannot think of anything automatic for that scenario.

docs/getting_started/install_ez_platform.md
```diff
@@ -122,6 +122,9 @@ After a moment the installer will ask you to provide a few parameters:
     Do not commit the secret to version control systems, or share it with anyone who does not strictly need it.
     If you have any suspicion that the secret may have been exposed, replace it with a new one.
     The same goes for other secrets, like database password, Varnish invalidate token, JWT passphrase, etc.
+    
+    Remember, that after changing the app secret you should regenerate application cache and log out all the users.
```
dabrt, line 126:
```suggestion
    After changing the app secret, make sure that you clear the application cache and log out all the users.
```

docs/getting_started/install_ez_platform.md
dabrt, line 127:
```suggestion
    For more information, see [Symfony documentation](https://symfony.com/doc/2.8/reference/configuration/framework.html#secret).
```

dabrt (changes requested):

docs/getting_started/install_ez_platform.md
dabrt: Hi @konradoboza , do you want this change to be merged up to newer versions of the documentation, up to master?

konradoboza: As mentioned in the PR description, it should be merged into all supported branches taking Symfony doc URL change into account along the way.

glye: +1 to this one.

glye (approved):

andrerom (commented):
+1

But we don't really have a way to log out users.. But when changing it doesnt all sessions become invalid (logged out) anyway, so it should rather state that?

andrerom (approved):

konradoboza:
I noticed that the mentioned URL is already in the text above (`secret` word points to the same doc), ref: https://doc.ibexa.co/en/latest/getting_started/install_ez_platform/#change-installation-parameters.

glye:
> But we don't really have a way to log out users.. But when changing it doesnt all sessions become invalid (logged out) anyway, so it should rather state that?

The Symfony doc says _keep in mind that changing this value will invalidate all signed URIs and Remember Me cookies. That’s why, after changing this value, you should regenerate the application cache and log out all the application users_. I tested locally on v3.2, and my session was not invalidated, not even after a full cache clear.

dabrt (approved):

PR #1293, opened 2021-01-27
3.3 update

DominikaK:
Restoring https://github.com/ezsystems/developer-documentation/pull/1291 after mistaken merge.

### DIFF
--- a/docs/updating/updating_to_3.3.md
+++ b/docs/updating/updating_to_3.3.md
@@ -109,6 +109,10 @@ Add the following rewrite rule to your web server configuration:
 
 ## Update the database
 
+!!! caution
+
+    Before starting this step, back up your database.
+
 Apply the following database update script:
 
 === "Mysql"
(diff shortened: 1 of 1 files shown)


docs/updating/updating_to_3.3.md
```diff
@@ -0,0 +1,138 @@
...
+```
+git merge origin/v3.2-to-v3.3-upgrade
+```
+
+This will introduce changes from the [website skeleton](https://github.com/ibexa/website-skeleton/blob/main/composer.json)
+and result in conflicts in `composer.json`.
+
+Resolve the conflicts in the following way:
+
+- Make sure all `ezsystems/*` packages are removed.
```
mnocon, line 27: I've talked with @webhdx and it's not that easy.

User should remove most of the of `ezsystems/*` packages, but if there are any packages created by us that are not part of a standard installation (for example: translations packages, e-content package) that were added later explicitly by the user then they need to be kept.

DominikaK: `Make sure all automatically added ezsystems/* packages are removed. If you explicitly added any packages that are not part of the standard installation, retain them.` < sound OK?

mnocon: Sounds good to me!

PR #1301, opened 2021-02-05
Disable top commerce menu

DominikaK:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | n/a
| Versions      | v3, master

Replaces generic "remove tab" example with actual use case of removing the Commerce tab.

### DIFF
--- a/docs/extending/extending_menus.md
+++ b/docs/extending/extending_menus.md
@@ -143,7 +143,7 @@ $menu->addChild(
 
 ## Modifying menu items
 
-Reorder menu items, i.e. reverse the order:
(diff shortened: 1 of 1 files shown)


docs/extending/extending_menus.md
```diff
@@ -143,20 +143,21 @@ $menu->addChild(
 
 ## Modifying menu items
 
-Remove the *Media* menu item from the Content tab:
+Reorder menu items, i.e. reverse the order:
```
dabrt, line 146: Did you use i.e. as "for example" here?

dabrt (approved):

PR #1308, opened 2021-02-15
Remove install-solr script from installation guide

DominikaK:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | n/a
| Versions      | 3.x, master

The install-solr script is no longer present in the meta-repository, and was intended for development only anyways.


dabrt (approved):
You might consider listing the steps and linking them to the right sections in the target article, as I did with Elasticsearch. For consistency, and to how readers what steps are involved right away.

PR #1314, opened 2021-02-22
Extract FT tutorial code to separate files

DominikaK:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | n/a
| Versions      | 3.3, master

Note: including the `steps` folder is not an elegant solution, but necessary for a multi-step tutorial that improves on the same files.

### DIFF
--- a/docs/tutorials/generic_field_type/7_add_a_validation.md
+++ b/docs/tutorials/generic_field_type/7_add_a_validation.md
@@ -4,6 +4,7 @@ To provide basic validation that ensures both coordinates are provided, add asse
 
 ```php
 [[= include_file('code_samples/field_types/2dpoint_ft/src/FieldType/Point2D/Value.php', 6, 23) =]]
+// ...
 ```
 
 As a result, if a user tries to publish the Point 2D with just one value, they will receive an error message.
--- a/docs/tutorials/generic_field_type/4_introduce_a_template.md
+++ b/docs/tutorials/generic_field_type/4_introduce_a_template.md
@@ -20,7 +20,7 @@ Your basic template for Point 2D should look like this:
 
 ## Template mapping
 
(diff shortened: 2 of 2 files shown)


docs/tutorials/generic_field_type/7_add_a_validation.md
```diff
@@ -1,35 +1,9 @@
...
-    /**
-     * @var float|null
-     *
-     * @Assert\NotBlank()
-     */
-    private $y;
-    
- // ...   
-}
+[[= include_file('code_samples/field_types/2dpoint_ft/src/FieldType/Point2D/Value.php', 6, 23) =]]
```
dabrt, line 6: I can see a slight problem here - the object has no ending parenthesis - I know where it comes from, we're pulling part of the file.

Would the following include work?

```
[[= include_file('code_samples/field_types/2dpoint_ft/src/FieldType/Point2D/Value.php', 6, 23) =]]
[[= include_file('code_samples/field_types/2dpoint_ft/src/FieldType/Point2D/Value.php', 57) =]]
```

 (Also, the first line is missing, is it on purpose? If not, would the following include work?)
 [[= include_file('code_samples/field_types/2dpoint_ft/src/FieldType/Point2D/Value.php', 0, 1) =]]

dabrt (approved):

docs/tutorials/generic_field_type/7_add_a_validation.md
DominikaK, line 6:
```suggestion
[[= include_file('code_samples/field_types/2dpoint_ft/src/FieldType/Point2D/Value.php', 6, 23) =]]
// ...
```
We're not quoting the full file here (therefore, "add" in the introductory text), and the parenthesis goes after the rest of the code. But let's add an ellipsis.

PR #1316, opened 2021-02-24
EZP-32186: FormMapper for Datatype uses obscure syntax

DominikaK:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | https://issues.ibexa.co/browse/EZP-32186
| Versions      | 3.x, master

Note: This reuses code from the FT tutorial, instead of basing on the built-in FTs

### DIFF
--- a/docs/api/field_type_form_and_template.md
+++ b/docs/api/field_type_form_and_template.md
@@ -27,7 +27,7 @@ You have to add your form type to the content editing form, for example:
 ```
 
 Your type has to be called `value`.
-In the example above, the custom `Point2DType::class` is used, but you can use standard Symfony form type instead.
+In the example above, the custom `Point2DType::class` is used, but you can use a standard Symfony form type instead.
 
 It's good practice to encapsulate Fields with custom types as it allows easier templating.
 Type has to be compatible with your Field Type's `eZ\Publish\Core\FieldType\Value` implementation.
--- a/docs/api/field_type_type_and_value.md
+++ b/docs/api/field_type_type_and_value.md
@@ -140,9 +140,9 @@ This ensures that the initialization steps shared by all Field Types are execute
 
 It is recommended to use a simple associative array format for the settings schema returned by `eZ\Publish\SPI\FieldType\FieldType::getSettingsSchema()`, which follows these rules:
 
-- The key of the associative array identifies a setting (e.g. `format`)
+- The key of the associative array identifies a setting (for example, `format`)
 - Its value is an associative array describing the setting using:
-    - `type` to identify the setting type (e.g. `int` or `string`)
+    - `type` to identify the setting type (for example, `int` or `string`)
     - `default` containing the default setting value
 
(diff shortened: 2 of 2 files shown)


vidarl (approved):
This looks good to me

But did you double-check with someone in engineering if they find it valuable to use the formFactory in the example?
My opinion is stated in https://issues.ibexa.co/browse/EZP-32186, someone might not agree

docs/api/field_type_form_and_template.md
```diff
@@ -17,39 +17,17 @@ The `FieldValueFormMapperInterface::mapFieldValueForm` method accepts two argume
...
-}
+[[= include_file('code_samples/field_types/2dpoint_ft/src/FieldType/Point2D/Type.php', 6, 7) =]][[= include_file('code_samples/field_types/2dpoint_ft/src/FieldType/Point2D/Type.php', 10, 11) =]][[= include_file('code_samples/field_types/2dpoint_ft/src/FieldType/Point2D/Type.php', 12, 13) =]]
+// ...
+
+[[= include_file('code_samples/field_types/2dpoint_ft/src/FieldType/Point2D/Type.php', 33, 41) =]]
 ```
 
 Your type has to be called `value`.
-In the example above, `CheckboxFieldType::class` is used, but you can use standard Symfony form type instead.
+In the example above, the custom `Point2DType::class` is used, but you can use standard Symfony form type instead.
```
dabrt, line 30:
```suggestion
In the example above, the custom `Point2DType::class` is used, but you can use a standard Symfony form type instead.
```
Unless there is only one (seems like not), in wchich case the article would be "the".

docs/api/field_type_type_and_value.md
```diff
@@ -143,24 +140,15 @@ This ensures that the initialization steps shared by all Field Types are execute
 
 It is recommended to use a simple associative array format for the settings schema returned by `eZ\Publish\SPI\FieldType\FieldType::getSettingsSchema()`, which follows these rules:
 
-- The key of the associative array identifies a setting (e.g. `default`)
+- The key of the associative array identifies a setting (e.g. `format`)
```
dabrt, line 143:
```suggestion
- The key of the associative array identifies a setting (for example, `format`)
```

docs/api/field_type_type_and_value.md
```diff
@@ -143,24 +140,15 @@ This ensures that the initialization steps shared by all Field Types are execute
...
 - Its value is an associative array describing the setting using:
     - `type` to identify the setting type (e.g. `int` or `string`)
```
dabrt, line 145:
```suggestion
    - `type` to identify the setting type (for example, `int` or `string`)
```

dabrt (approved):

DominikaK:
After verification with engineering, the current code is correct, what we need to do is remove the confusing comment in code, as it refers to an internal DataTransformer (will be done separately).

PR #1337, opened 2021-03-18
Reorganize update section

DominikaK:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | n/a
| Versions      | all

Merges the 3.0 update with regular update guide.

- Both Update and Migration are now placed in one top-level navigation category.
- Misc steps at the end of the procedure merged into one "Finish the update" step.
- Replaced variable with product name spelled out. This change goes to all versions, where variables may be different/non existent and that frankly causes more problems than it does good.
- Renamed some pages for clarify and better searchability

(Note for reviewers:
The 4_x steps for updating 3.0 code were moved only, not modified.
The 5_ steps were also mostly moved with small text clarifications, but GH got confused and marked some pages as new.)

### DIFF
--- a/docs/updating/5_update_3.2.md
+++ b/docs/updating/5_update_3.2.md
@@ -15,3 +15,26 @@ mysql -u <username> -p <password> <database_name> < upgrade/db/mysql/ezplatform-
 ``` bash
 psql <database_name> < upgrade/db/postgresql/ezplatform-3.1.0-to-3.2.0.sql
 ```
+
+## Update entity managers
+
+Version v3.2.6 introduces new entity managers.
+To ensure that they work in multi-repository setups, you must update the GraphQL schema.
+You do this manually by following this procedure:
+
+1. Update your project to v3.2.6 and run the `php bin/console cache:clear` command to generate the [service container](../guide/service_container.md).
+
+1. Run the following command to discover the names of the new entity managers. 
+    Take note of the names that you discover:
+
+    `php bin/console debug:container --parameter=doctrine.entity_managers --format=json | grep ibexa_`
+
+1. For every entity manager prefixed with `ibexa_`, run the following command:
+
+    `php bin/console doctrine:schema:update --em=<ENTITY_MANAGER_NAME> --dump-sql`
+  
+1. Review the queries and ensure that there are no harmful changes that could affect your data.
+
+1. For every entity manager prefixed with `ibexa_`, run the following command to run queries on the database:
+
+    `php bin/console doctrine:schema:update --em=<ENTITY_MANAGER_NAME> --force`


mkdocs.yml
```diff
@@ -541,49 +541,39 @@ nav:
...
-        - '6. Dump assets': 'updating/6_dump_assets.md'
-        - '7. Commit, test and merge': 'updating/7_commit_test_merge.md'
+    - Update and migration:
+        - Updating Ibexa DXP:
+            - 'Updating to v3.3': 'updating/updating_to_3.3.md'
+            - 'Updating Ibexa DXP': 'updating/updating.md'
+            - '1. Check out a version': 'updating/1_check_out_version.md'
+            - '2. Resolve conflicts': 'updating/2_merge_composer.md'
+            - '3. Update the app': 'updating/3_update_app.md'
+            - '4. Upgrade the code':
```
mnocon, line 551: Is there a reason why some things are updated` (app, database) and some are upgraded (code)?

docs/updating/2_merge_composer.md
```diff
@@ -1,36 +1,59 @@
...
 
-## Manual merging
+## 2.1. Resolve conflicts in `composer.lock`
 
-!!! note "Sorted packages since 2.4 changes"
+Ignore the conflicts in `composer.lock`, because this file is regenerated when you execute `composer update` later.
+It is easiest to check out the version of the file from the tag and add it to the changes.
 
-    Because the 2.4 packages in `composer.json` are sorted, there will be more conflicts when updating to 2.4, but far fewer conflicts in the future. This is controlled by [sort-packages](https://getcomposer.org/doc/06-config.md#sort-packages) config in `composer.json`.
+If you get a lot of conflicts (for example, on the `doc` folder),
```
mnocon, line 8: It feels like this part deserves its own "chapter", it's not about composer.lock

mkdocs.yml
DominikaK: The idea was that update is smaller and more automatic, while upgrade is larger, requiring more hands-on work (as with upgrading custom code to v3), but I'm not married to the idea, let's see what the rest of the doc team has to say about it.

docs/updating/2_merge_composer.md
DominikaK: So maybe this under general "Resolve conflicts" heading, and specifics about `composer.json` in a separate section below, like they are now?

docs/updating/5_update_2.3.md
```diff
@@ -0,0 +1,43 @@
...
+
+Apply the following database update script:
+
+``` bash
+mysql -u <username> -p <password> <database_name> < vendor/ezsystems/ezpublish-kernel/data/update/mysql/dbupdate-7.2.0-to-7.3.0.sql
+```
+
+## Changes to timestamp
+
+A new timestamp column has been added in order to keep track of when items were trashed, this is exposed in the API but not yet in UI.
```
mnocon, line 15: > this is exposed in the API but not yet in UI.

not sure if this detail is important here (or if it's still valid).

docs/updating/6_finish_the_update.md
```diff
@@ -0,0 +1,57 @@
...
+yarn install
+yarn encore prod
+```
+
+## 6.3. Commit, test and merge
+
+Once all the conflicts have been resolved, and `composer.lock` updated, you can commit the merge.
+
+Note that you may or may not keep `composer.lock`, depending on your version management workflow.
+If you do not wish to keep it, run `git reset HEAD <file>` to remove it from the changes.
```
mnocon, line 37: It seems that using `<file>` is not needed here, we could probably use `composer.lock` directly?

docs/updating/6_finish_the_update.md
```diff
@@ -0,0 +1,57 @@
...
+git checkout master
+git merge <branch_name>
+```
+
+!!! note "Insecure password hashes"
+
+    To ensure that no users have unsupported, insecure password hashes, run the following command:
+    
+    ``` bash
+    php bin/console ezplatform:user:validate-password-hashes
```
mnocon, line 52: For Ibexa DXP this command will work, but with a deprecation notice - the best version would be `php bin/console ibexa:user:validate-password-hashes`

And if I'm not wrong it's available only for 3.1+ (https://github.com/ezsystems/ezplatform-kernel/pull/99), so does not apply for v1 and v2.

mnocon (commented):
It's hard to review that many files without being able to see how it will look live, but I think it will all make sense 😄

mnocon:
1) Zrzut ekranu 2021-03-30 o 13 38 12

> If your application consists of several packages that are placed in locations other than the src/ folder, apply the suggestions from the upgrade documentation to all the packages before you run composer update.

I don't know what the "upgrade documentation" is, is it the "Upgrade the app" part? I think a link here would be useful.

2) 
Zrzut ekranu 2021-03-30 o 13 39 43

> Perform these steps only if you are updating to version v3.0. If not, move on to the 5. Update the database step.

The "Update the database" link is a dead one.

docs/updating/6_finish_the_update.md
```diff
@@ -49,7 +49,7 @@ git merge <branch_name>
     To ensure that no users have unsupported, insecure password hashes, run the following command:
     
     ``` bash
-    php bin/console ezplatform:user:validate-password-hashes
+    php bin/console ibexa:user:validate-password-hashes
```
mnocon, line 52: If I understand correctly this page is meant for users using all possible versions:
```
# In v1 and v2:
php bin/console assetic:dump -e prod
# In v2 and v3:
yarn install
yarn encore prod
```

So maybe we should mention that for version 3.2 you should run:
`php bin/console ezplatform:user:validate-password-hashes`
For 3.3:
`php bin/console ibexa:user:validate-password-hashes`
And for v1 and v2 it's not available.

mnocon (approved):

juskora (changes requested):

juskora (approved):

PR #1348, opened 2021-04-08
Update user_management.md

glye:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | https://issues.ibexa.co/browse/EZP-32383
| Versions      | 2.5+

https://doc.ibexa.co/en/latest/guide/user_management/user_management/#repeating-passwords

Doc change to match the current state of the functionality, which is as intended according to the developer. Though we'd like to implement the feature, this would likely be in DXP 4.0+.

### DIFF
--- a/docs/guide/user_management/user_management.md
+++ b/docs/guide/user_management/user_management.md
@@ -59,7 +59,11 @@ The notification will be displayed in the Back Office after login and in the Use
 
 You can set a rule that the password cannot be reused.
 You set it for the User Content Type in the **User account (ezuser)** Field Type's settings.
-When this is set, the user cannot type in the same password as before, when it expires. It has to be changed to a new one. (Please note that this only checks the new password against the current one. A password that is older than this can be used again.)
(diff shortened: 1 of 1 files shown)


docs/guide/user_management/user_management.md
```diff
@@ -59,7 +59,7 @@ The notification will be displayed in the Back Office after login and in the Use
 
 You can set a rule that the password cannot be reused.
 You set it for the User Content Type in the **User account (ezuser)** Field Type's settings.
-When this is set, the user will not be able to set a password that had been in use before.
+When this is set, the user cannot type in the same password as before, when it expires. It has to be changed to a new one. (Please note that this only checks the new password against the current one. A password that is older than this can be used again.)
```
DominikaK, line 62:
```suggestion
When this is set, the user cannot type in the same password when it expires.
It has to be changed to a new one.

This only checks the new password against the current one.
A password that has been used before can be used again.
```
^ if I understood correctly the "older password" thing

DominikaK (approved):

PR #1351, opened 2021-04-12
Core packages in OSS

DominikaK:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | n/a
| Versions      | 3.3+

Based on feedback from the community, adds a note indicating which packages are available in the OSS edition.

### DIFF
--- a/docs/guide/bundles.md
+++ b/docs/guide/bundles.md
@@ -31,7 +31,7 @@ see the [How to Remove a Bundle](http://symfony.com/doc/5.0/bundles/remove.html)
 
 !!! tip
 
-    The core packages are included in Ibexa Open Source.
+    Ibexa Open Source is composed of the core packages.
 
 |Bundle|Description|
 |---------|-----------|


docs/guide/bundles.md
```diff
@@ -29,6 +29,10 @@ see the [How to Remove a Bundle](http://symfony.com/doc/5.0/bundles/remove.html)
 
 ## Core packages
 
+!!! tip
+
+    The core packages are included in Ibexa Open Source.
```
dabrt, line 34: Howabout using the active voice?
"Ibexa Open Source includes the core packages."?

dabrt (approved):

docs/guide/bundles.md
DominikaK: I've tried to make it clear that core packages are the *only* ones in OSS, without actually saying "only" :)

dabrt: In this case, I like "Ibexa Open Source is composed of the core packages" best, despite its passivevoiciness

DominikaK, line 34:
```suggestion
    Ibexa Open Source is composed of the core packages.
```

PR #1362, opened 2021-04-30
EZEE-3499: Update vcl configuration for Fastly

DominikaK:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | https://issues.ibexa.co/browse/EZEE-3499
| Versions      | all

Note for the future: add the same information to 1.13 and 3.3 after relevant tags are out.


mnocon (commented):
1) We can't just paste the snippet here, because people won't know where to insert it (between which lines), we need to indicate it somehow
2) The fix will be available in v1.13.7 once it's tagged, but it's not released yet (similar story for 3.3 - the fix is not included in 3.3.2 out of the box, will be available in 3.3.3)

docs/updating/5_update_2.5.md
```diff
@@ -161,3 +161,15 @@ You do this manually by following this procedure:
 1. For every entity manager prefixed with `ibexa_`, run the following command to run queries on the database:
 
     `php bin/console doctrine:schema:update --em=<ENTITY_MANAGER_NAME> --force`
+
+#### VCL configuration for Fastly
+
+If you use Fastly, update your VCL.
```
dabrt, line 167:
```suggestion
If you use Fastly, update your VCL configuration.
```

dabrt (approved):

PR #1365, opened 2021-05-06
Combine supported and recommended requirements

DominikaK:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | https://issues.ibexa.co/browse/IBX-313
| Versions      | all, with PHP 7.3 only for 1.13

### DIFF
--- a/docs/getting_started/requirements.md
+++ b/docs/getting_started/requirements.md
@@ -31,8 +31,7 @@ event MPM is recommended, if you need to use prefork you also need the `mod_php`
 
 ## PHP
 
-- 7.3
-- 7.4 (recommended)
+- 7.4
 
 ### PHP packages
 
@@ -79,7 +78,7 @@ event MPM is recommended, if you need to use prefork you also need the `mod_php`
 
 ## Asset manager
 
-- `Node.js` 10.22 or higher
+- `Node.js` 12 or higher
 - `yarn` 1.15.2 or higher
 
 ## Browser


DominikaK:
@glye: just making sure: this applies to all versions, back until 1.13?

glye:
Sorry, to clarify: We don't want to bump the _minimum requirements_ to 7.4, they should stay as they are. But we want to make it clear that support goes _up to_ 7.4, which is useful information since php 7.2 is EOL. This is what you meant, right @lserwatka?

I had a closer look at the pages, and it seems only 1.13 has a lower max than 7.4, it stops at 7.2. So only this page would need a change, then.

Should we run it through QA, or are we sure 1.13 is good with 7.4, @lserwatka?

DominikaK:
We actually list both 7.3 and 7.4 as "supported setups", we never name the minimum, crafty us ;) . So the question is, is 7.4 "recommended", or is it still 7.3?

Of course, another question is, does the "recommended" vs "supported" setups division on the page make sense in the longer run.
Wouldn't just one table with ranges (7.3-7.4 or whatever) be better?

glye:
LS said _"We can even state that we are always recommending latest supported by our product PHP version, so in this case 7.4. Version 8.0 is not supported yet."_
I like your idea of a supported range, plus the general statement that we recommend the most recent in that range.

1.13 is still an open question. Currently it lists php 7.0 and 7.2, but not 7.1. Could we safely extend this to the hole range of 7.0-7.4?

lserwatka:
Yes, we need a rage. We need to keep min supported version as stated in the composer.json. So don't change it.

As for 1.13 we need confirmation from @micszo

lserwatka:
We should always recommend latest supported by our product PHP version. I don't think this split makes sense anymore @DominikaK You can merge it and always recommend latest version.

DominikaK:
I've combined supported and recommended requirements into one set with ranges, but we'll need QA to verify if all the versions are correct.

docs/getting_started/requirements.md
```diff
@@ -6,119 +6,108 @@ The following server requirements cover both running the software on-premise and
...
-
-
-### Development and Experimental setups
-
-[[= product_name =]] can theoretically run and execute on many more setups than the ones listed as recommended and supported, including any [operating system supported by PHP](https://wiki.php.net/platforms), on a PHP 7.3 version or higher that pass the [Symfony requirements](http://symfony.com/doc/5.0/reference/requirements.html), using cache solutions technically supported by [Symfony Cache component](https://symfony.com/doc/5.0/components/cache/cache_pools.html), using databases supported by [Doctrine DBAL](https://www.doctrine-project.org/projects/doctrine-dbal/en/2.9/reference/configuration.html#driver), and using a binary file storage solution supported by [FlySystem](https://github.com/thephpleague/flysystem#adapters).
+The minimal setup requires PHP,  MySQL/MariaDB, Apache/Nginx, Node.js and `yarn`.
+Recommendation for production setups is to use Varnish/Fastly, Redis/Memcached, NFS/EFS/S3 and Solr/Elasticsearch in a [clustered setup](../guide/clustering.md).
 
-Examples of Development setups:
+Using the latest supported version or each product or component is always recommended.
```
glye, line 12:
```suggestion
Using the latest supported version of each product or component is always recommended.
```

glye: Do we need to clarify "latest supported **by us**"? I mean, php 8 is supported by php but not by us.

docs/getting_started/requirements.md
```diff
@@ -6,119 +6,108 @@ The following server requirements cover both running the software on-premise and
...
+## DBMS
+
+- MariaDB 10.3, 10.4 (optionally 10.2 - deprecated)
+- MySQL 8.0 (optionally 5.7 - deprecated)
+- PostgreSQL 10+
+
+## PHP
+
+- 7.3
+- 7.4
```
glye, line 35: Should we add a "(recommended)" here for 7.4?

docs/getting_started/requirements.md
```diff
@@ -6,119 +6,108 @@ The following server requirements cover both running the software on-premise and
...
 
-These setups have been undergone some additional manual testing and is known to work.
+- Linux ext4 / XFS
 
--   Mozilla® Firefox® most recent stable version
--   Google Chrome™ most recent stable version
+## Package manager
 
-### Supported browsers
+- Composer 2.0.8 or higher
```
glye, line 78:
```suggestion
- Composer 2.0.13 or higher
```
There's a security issue fixed in this version.

docs/getting_started/requirements.md
```diff
@@ -6,119 +6,108 @@ The following server requirements cover both running the software on-premise and
...
+
+## Asset manager
+
+- `Node.js` 10.22 or higher
+- `yarn` 1.15.2 or higher
+
+## Browser
+
+[[= product_name =]] is developed to work with *any* web browser that support modern standards, on *any* screen resolution suitable for web, running on *any* device. However for the Editorial and Administration User Interfaces you'll need; a minimum of 1366-by-768 screen resolution, a desktop or tablet device, and a recommended/supported browsers found below.
```
glye, line 87:
```suggestion
[[= product_name =]] is developed to work with *any* web browser that support modern standards, on *any* screen resolution suitable for web, running on *any* device. However for the Editorial and Administration User Interfaces you'll need; a minimum of 1366-by-768 screen resolution, a desktop or tablet device, and a recommended/supported browser among the ones found below.
```

glye (changes requested):
Some comments (some are optional), but looks good. Seems to be the right direction. Given all the pink vs. green here, is the page shorter overall? That's also good.

docs/getting_started/requirements.md
DominikaK, line 12:
```suggestion
Using the latest listed version of each product or component is always recommended.
```

DominikaK: Probably, let's make it completely clear

DominikaK:
>Given all the pink vs. green here, is the page shorter overall?

docs/getting_started/requirements.md
glye: This is covered by the "Using the latest listed version of each product or component is always recommended." I guess?

docs/getting_started/requirements.md
glye: I realise now that the "Using the latest listed version of each product or component is always recommended." makes this kind of redundant. But it doesn't hurt either, now that it's there. 🤷‍♂️

glye (approved):

docs/getting_started/requirements.md
```diff
@@ -6,119 +6,90 @@ The following server requirements cover both running the software on-premise and
...
+- ImageMagick
+- GD
 
-[[= product_name =]] is developed to work with *any* web browser that support modern standards, on *any* screen resolution suitable for web, running on *any* device. However for the Editorial and Administration User Interfaces you'll need; a minimum of 1366-by-768 screen resolution, a desktop or tablet device, and a recommended/supported browsers found below.
+## [Clustering](../guide/clustering.md)
 
-### Recommended browsers
+- Linux NFS or S3/EFS (for IO, aka binary files stored in content repository, not supported with legacy)
+- Redis 4.0+, 5.0 or higher (separate instances for session and cache, both using a `volatile-*` [eviction policy](https://redis.io/topics/lru-cache), session instance configured for persistence) or [Memcached](https://memcached.org/) 1.5 or higher
+- [Varnish](http://varnish-cache.org/) 6.0LTS with [varnish-modules](https://github.com/varnish/varnish-modules/blob/master/README.rst) or [Fastly](https://www.fastly.com/) using [the bundle provided with [[= product_name_exp =]]](../guide/http_cache.md#serving-varnish-through-fastly) (for HttpCache)
```
juskora, line 70:
```suggestion
- [Varnish](http://varnish-cache.org/) 6.0LTS with [varnish-modules](https://github.com/varnish/varnish-modules/blob/master/README.md) or [Fastly](https://www.fastly.com/) using [the bundle provided with [[= product_name_exp =]]](../guide/http_cache.md#serving-varnish-through-fastly) (for HttpCache)
```

docs/getting_started/requirements.md
```diff
@@ -6,119 +6,90 @@ The following server requirements cover both running the software on-premise and
...
+- Composer 2.0.13 or higher
+
+## Asset manager
+
+- `Node.js` 10.22 or higher
+- `yarn` 1.15.2 or higher
+
+## Browser
+
+[[= product_name =]] is developed to work with *any* web browser that support modern standards, on *any* screen resolution suitable for web, running on *any* device. However for the Editorial and Administration User Interfaces you'll need; a minimum of 1366-by-768 screen resolution, a desktop or tablet device, and a recommended/supported browser among the ones found below.
```
juskora, line 87:
```suggestion
[[= product_name =]] is developed to work with *any* web browsers that support modern standards, on *any* screen resolution suitable for web, running on *any* device. However for the Editorial and Administration User Interfaces you'll need; a minimum of 1366-by-768 screen resolution, a desktop or tablet device, and a recommended/supported browser among the ones found below.
```

micszo:
I did some testing locally on PHP 7.3 and 7.4 with version 1.13.
Below are the findings.

1) Using release v1.13.6 on PHP 7.3 there's an error at the end of the installation.
composer install ends with error:
```
> eZ\Bundle\EzPublishCoreBundle\Composer\ScriptHandler::clearCache


  [Symfony\Component\Debug\Exception\ContextErrorException]
  Warning: "continue" targeting switch is equivalent to "break". Did you mean to use "continue 2"?
```
Screenshot 2021-05-12 at 14 49 58
Frontend works.
Admin works.

2) Using release v1.13.6 on PHP 7.4 there's the same error at the end of the installation.
And both frontend and admin display error 500 `Notice: Trying to access array offset on value of type null`.

Screenshot 2021-05-12 at 15 08 25


3) Using stable branch 1.13 on PHP 7.3 composer update passes without errors.
Frontend works.
Admin works.

4) Using stable branch 1.13 on PHP 7.4 composer update passes without errors.
However both frontend and admin display the error 500 `Notice: Trying to access array offset on value of type null`.

Conclusion: we cannot go past PHP 7.3 with support for version 1.13.

glye:
Thanks, great findings. PHP 7.3 goes EOL at the end of this year, which is when v1.13 goes EOL too, so that's cool. But we need a fix for the installer.

DominikaK:
As master corresponds to 4.x, added two more changes:
- 4.x will only support PHP 7.4 (will keep 7.3 for lower branches)
- Node 12 is required.

juskora (approved):

PR #1366, opened 2021-05-07
IBX-155 Intro rewrite

dabrt:
| Question      | Answer |
| ------------- | --- |
| JIRA Ticket   | IBX-155 |
| Versions      | 3.2 and up |

### DIFF
--- a/docs/guide/personalization/basic_integration.md
+++ b/docs/guide/personalization/basic_integration.md
@@ -22,10 +22,12 @@ A code that includes an image may look like this:
 
 `<user_ID>` stands either for the user ID or session ID of the user who is currently 
 logged into your website (any URL-encoded string is allowed).
+
 `<content_type_ID>` stands for the [contentTypeId](../content_model.md#content-information) of the Content item or product that you want to track and recommend.
+
 `<content_ID>` stands for the [id](../content_model.md#content-information) of the Content item or product that you want to track and recommend.
 
-The following examples show how to integrate a CLICK event:
+The following examples show how you can integrate a CLICK event:
 
 PHP:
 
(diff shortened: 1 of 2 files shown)


docs/guide/personalization/personalization_quickstart.md
```diff
@@ -1,45 +1,32 @@
...
-It looks like this:
+The service primarily relies on event tracking. 
+For the events to be registered, every content item or product page must call 
+a special tracking URL.
+The simplest way to the tracking URL is placing a one pixel image on every page, 
+just like in the case of analytical tools or visitor counters.
+A code that includes an image may look like this:
 
-**`<img href="https://event.yoochoose.net/ebl/00000/click/john.smith/1/123" width="1" height="1">`**
+`<img href="https://event.yoochoose.net/ebl/00000/click/<user_ID>/1/<content_ID>" width="1" height="1">`
```
dkorostynski, line 23: @dabrt if we're changing 123 to <content_ID>, shouldn't we also change content type value? 
so it should be: 
/<content_type_ID>/<content_ID>" width="1" height="1">`

docs/guide/content_model.md
```diff
@@ -25,7 +25,7 @@ General information about a Content item is stored in a [`ContentInfo`](https://
 
 **`id`** - the unique ID of the Content object. These numbers are not recycled, so if an item is deleted, its ID will not be reused when a new one is created.
 
-**`contentTypeId`** - the unique ID of the Content Type on which the Content item is based.
+**`contentTypeId`** - the unique numerical identifier of the Content Type, on which the Content item is based.
```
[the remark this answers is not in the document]
dabrt: Just thinking out loud - whenever we have ID in its short form I get an impression that it is an actual name name of a variable. Could we use ID when it actially is a name of a variable, and identifier when we refer to a general notion of something that identifies an element?

docs/guide/personalization/basic_integration.md
```diff
@@ -0,0 +1,154 @@
+# Basic service integration
+
+To return recommendations, you must first [enable the Personalization service](enabling_personalization.md).
+Then, you must integrate the service with I[= product_name =]] by activating
```
DominikaK, line 4:
```suggestion
Then, you must integrate the service with [= product_name =]] by activating 
```

DominikaK (approved):

PR #1367, opened 2021-05-10
Update to v3.3.x

DominikaK:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | n/a
| Versions      | all

### DIFF
--- a/docs/updating/updating_to_3.3.md
+++ b/docs/updating/updating_to_3.3.md
@@ -187,6 +187,12 @@ composer run post-install-cmd
     
     To update to v3.3.3, remove `Kaliop\eZMigrationBundle\eZMigrationBundle::class => ['all' => true],`
     from `config/bundles.php` before running `composer require`.
+    
+    Then, in `composer.json`, set minimum stability to `stable`:
+    
+    ```
+    "minimum-stability": "stable",
+    ```
 
 To update from one v3.3 patch version to another (for example, from v3.3.2 to v3.3.3), run:
 


docs/updating/updating_to_3.3.md
```diff
@@ -167,7 +168,45 @@ Check the Location ID of the "Components" Content item and set it as a value of
...
+
+## Update to v3.3.x
+
+To update from one v3.3 patch version to another (for example, from v3.3.1 to v3.3.2), run:
+
+=== "[[= product_name_content =]]"
+
+    ``` bash
+    composer require ibexa/content:3.3.2 -W
+    composer recipes:install ibexa/content --force
```
mnocon, line 188: IMHO it's best to always add `-v` flag to increase verbosity:
`composer recipes:install ibexa/content --force -v`
(applies to other examples as well)

docs/updating/updating_to_3.3.md
```diff
@@ -167,7 +168,45 @@ Check the Location ID of the "Components" Content item and set it as a value of
...
+```
+
+## Update to v3.3.x
+
+To update from one v3.3 patch version to another (for example, from v3.3.1 to v3.3.2), run:
+
+=== "[[= product_name_content =]]"
+
+    ``` bash
+    composer require ibexa/content:3.3.2 -W
```
webhdx, line 187: I'd use `--with-all-dependencies` instead of `-W`. It's the same option but `-W` doesn't say much.

webhdx (approved):
Looks good to me.

Steveb-p (approved):

mnocon (approved):

juskora (approved):

docs/updating/updating_to_3.3.md
```diff
@@ -167,7 +168,49 @@ Check the Location ID of the "Components" Content item and set it as a value of
...
+
+    You can only update to the latest patch release of 3.3.x.
+
+To update from one v3.3 patch version to another (for example, from v3.3.1 to v3.3.2), run:
+
+=== "[[= product_name_content =]]"
+
+    ``` bash
+    composer require ibexa/content:3.3.2 --with-all-dependencies
+    composer recipes:install ibexa/content --force -v
```
Steveb-p, line 192: Since recipes might contain changes that are required for, for example, `cache:clear` command, we should call `composer require` with `--no-scripts` option to postpone them.

```suggestion
    composer require ibexa/content:3.3.2 --with-all-dependencies --no-scripts
    composer recipes:install ibexa/content --force -v
    composer run post-install-cmd
```

Discovered by @micszo

micszo: It does not do the trick, we need to find a different solution 🙂

webhdx:
@DominikaK  We need to add a remark that upgrading to 3.3.3 requires removing `Kaliop\eZMigrationBundle\eZMigrationBundle::class => ['all' => true],` from `config/bundles.php` before running `composer require ibexa/<edition> (...)`

This was discovered last week and this is the issue with how Flex handles recipes. We can't do much about it currently.

micszo:
Current status: 
right now we have the situation that doctrine/dbal is installed with stability dev during update.
This causes the stability of the installation to be set to dev.

micszo (approved):
I think we're good to go.

PR #1375, opened 2021-05-14
Namespace fix

dabrt:
| Question      | Answer |
| ------------- | --- |
| JIRA Ticket   | none |
| Versions      | master |

Namespace fix in master

### DIFF
--- a/docs/community_resources/package_structure.md
+++ b/docs/community_resources/package_structure.md
@@ -109,4 +109,4 @@ That namespace needs to be mapped to the `src/contracts` directory of a package.
 
 !!! note
 
-    Backward compatibility for interfaces and objects defined in the `Contracts` namespace is guarranteed.
+    Backward compatibility for interfaces and objects defined in the `Contracts` namespace is guaranteed.


docs/community_resources/package_structure.md
```diff
@@ -88,21 +88,25 @@ A package may introduce a namespace for contracts, to be consumed by first and t
...
 ```php
-namespace Ibexa\Platform\Contracts\Commerce\Shop;
+namespace Ibexa\Contracts\Commerce\Shop;
 ```
 
 That namespace needs to be mapped to the `src/contracts` directory of a package.
+
+!!! note
+
+    Backward compatibility for interfaces and objects defined in the `Contracts` namespace is guarranteed.
```
Steveb-p, line 112: Should we add something along the lines: "Compatibility for other code is maintained on a best effort basis"?

dabrt: Isn't it somethin that is already assumed for the whole product? That is why we maintain BC breakdown lists, mention BC promises etc.

adamwojs (approved):

docs/community_resources/package_structure.md
DominikaK, line 112:
```suggestion
    Backward compatibility for interfaces and objects defined in the `Contracts` namespace is guaranteed.
```

DominikaK (approved):

docs/community_resources/package_structure.md
alongosz: >> Should we add something along the lines: "Compatibility for other code is maintained on a best effort basis"?

> Isn't it something that is already assumed for the whole product? That is why we maintain BC breakdown lists, mention BC promises etc.

It is assumed but history shows not very clear for 3rd party Consumers. Worth to state it clear. I would also include here SemVer (sth like Backward compatibility according to [SemVer](https://semver.org/) rules or some similar sentence rephrased). Otherwise it might create an expectation of never ending BC.

Steveb-p: Symfony has a good description for it I guess?

https://symfony.com/doc/current/contributing/code/bc.html#:~:text=That's%20why%20we%20promise%20you,allowed%20to%20break%20backward%20compatibility.

alongosz: > Symfony has a good description for it I guess?
> 
> https://symfony.com/doc/current/contributing/code/bc.html#:~:text=That's%20why%20we%20promise%20you,allowed%20to%20break%20backward%20compatibility.

This is Symfony promise, not ours. Let's not suggest it's the same, because we have no control over it.

Steveb-p: That's not what I meant :smile: I just wanted to use their explanation as a base for ours.

dabrt: That is a great idea for a larger declaration of our philosophy. We know that a promise is needed, but we want to wait until the details crystallize, as far as what we can actually promise. This PR is only about reassuring our clients about interfaces and objects defined in the `Contracts` namespace.

PR #1381, opened 2021-05-19
RN link absolutization

dabrt:
| Question      | Answer |
| ------------- | --- |
| JIRA Ticket   | none |
| Versions      |  all |

### DIFF
--- a/docs/releases/ez_platform_v1.10.0.md
+++ b/docs/releases/ez_platform_v1.10.0.md
@@ -7,7 +7,7 @@ If you are looking for the Long Term Support (LTS) release, see[ https://ezplat
 
 Upgrade notes
 
-This release contains special steps to follow further described in [Updating eZ Platform](../updating/updating.md).
+This release contains special steps to follow further described in [Updating eZ Platform](https://doc.ibexa.co/en/1.13/updating/updating).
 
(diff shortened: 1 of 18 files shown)


Steveb-p:
May I ask why?

This will making working with local documentation build a nightmare, and if I'm not mistaken this will cause docs to keep switching between versions?

DominikaK:
In short: because all release notes are merged to all branches of the doc (it's a business decision), and links from RNs in newer versions are constantly broken, because the relevant page does not exist in that version. For example, Site Factory was released with 3.1, to links from https://doc.ibexa.co/en/2.5/releases/ez_platform_v3.1/#site-factory lead to a 404.

It's a ugly hack, but we couldn't figure out a better one. Other ideas welcome :)

DominikaK:
Side: considering the almost non-existent traffic to old RN pages, we could think of removing everything pre-1.13.

DominikaK (approved):

dabrt:
> Side: considering the almost non-existent traffic to old RN pages, we could think of removing everything pre-1.13.

We could remove them from 3.x and leave in 1.x and 2.x branches, or, even more aggressively, remove all rns in their respective master branches, which would leave us with a clean slate in 4.x

PR #1393, opened 2021-06-07
IBX-381: Image Editor quality setting

DominikaK:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | https://issues.ibexa.co/browse/IBX-381
| Versions      | 3.3+

### DIFF
--- a/docs/guide/image_editor.md
+++ b/docs/guide/image_editor.md
@@ -61,6 +61,20 @@ system:
                                     label: Custom
 ```
 
+### Image quality
+
+You can configure the quality of the images modified in the Image Editor with the following configuration.
+
+The setting accepts values between 0 and 1, which corresponds to the compression level, with 0 being the strongest compression.
+The default quality is 0.92:
+
+``` yaml
+system:
+    <scope>:
+        image_editor:
+            image_quality: 0.8
+```
+
 ### Additional information
 
 Each image can be accompanied by additional information that is not visible to the user.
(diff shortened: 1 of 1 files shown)


ViniTou (approved):

Steveb-p (approved):

docs/guide/image_editor.md
```diff
@@ -87,3 +87,16 @@ new FieldValue([
      ],
  ]),
 ```
+
+### Image quality
```
dabrt, line 91: The things I'd change is:
Move the image quality setting before the Additional information" section. Image quality is a readily available setting thet you do in the same place where you configure actions, while "Additional information" is an extension point, not directly related to how the editor works.

Also - what does quality mean? Is it compression rate of a JPEG image? Could we explain this? If so, does 1 correspond to 100% compression rate? Could we then say "...where .92 corresponds to 92% compression rate"?

Not overly attached to this, though.

dabrt (approved):

docs/guide/image_editor.md
DominikaK: Makes sense with the order of info, will move.
As for quality, I hope @ViniTou can elaborate?

ViniTou: I just created setting on our side, those questions should be addressed to @dew326.

But i would say its other way around, `0` is the strongest compression.

dew326: >A Number between 0 and 1 indicating the image quality to use for image formats that use lossy compression such as image/jpeg and image/webp.

https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/toDataURL - encoderOptions

dabrt (approved):

PR #1403, opened 2021-06-29
Http cache redesign

juskora:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | https://issues.ibexa.co/browse/IBX-400
| Versions      | 2.5+

Most of the content from https://github.com/ezsystems/ezplatform-http-cache/tree/master/docs/ is included in the documentation. 
Files: parameters.vcl and varnish5.vcl need to be relocated to be able to delete the docs folder.

### DIFF
--- a/docs/guide/cache/http_cache_config.md
+++ b/docs/guide/cache/http_cache_config.md
@@ -67,7 +67,7 @@ For the Page Builder, block cache by default respects `$content.ttl_cache$` and 
 However, if the given block value has a since or till date, 
 it is taken into account for the TTL calculation for both the block and the whole page.
 
-To overload this behavior, listen to [`BlockResponseEvents::BLOCK_RESPONSE`](../extending/extending_page.md#block-render-response),
+To overload this behavior, listen to [`BlockResponseEvents::BLOCK_RESPONSE`](../../extending/extending_page.md#block-render-response),
 and set priority to `-200` to adapt what Page Field Type does by default.
 
 For example, to disable cache for the block, use `$event->getResponse()->setPrivate()`.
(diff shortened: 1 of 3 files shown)


dabrt (commented):
At a first glance, the article seems very long. Can we split it into several articles under a common top level article?
Something like:
- HTTP Cache
-- HTTP Cache Configuration
-- Symfony reverse proxy and its replacements (this one might require further atomization into "tasks" or "master tasks"
-- Context-aware HTTP cache
-- Content-aware HTTP cache

Once split, I'd add links between articles to definitions of terms that have been presented earlier (like "purge" for example, which might appear in various places but has been explained early)

I'd add an explanation of TLAs and other terms that might occur throughout the article but have no definition, like VCL or FOS. (Expand them in each first occurrence in the resulting sub-articles.

mkdocs.yml
```diff
@@ -474,7 +474,13 @@ nav:
                 - 'Trash events': 'guide/repository/event_reference/trash_events.md'
                 - 'Other events': 'guide/repository/event_reference/other_events.md'
         - Cache:
-            - 'HTTP cache': 'guide/http_cache.md'
+            - 'HTTP cache': 
+              - 'HTTP cache': 'guide/cache/http_cache.md'
+              - 'HTTP cache configuration': 'guide/cache/http_cache_config.md'
+              - 'Symfony reverse proxy': 'guide/cache/symfony_reverse_proxy.md'
+              - 'Context-aware HTTP cache': 'guide/cache/context_aware_cache.md'
+              - 'Content-aware HTTP cache': 'guide/cache/content_aware_cache.md'
```
ViniTou, line 482: this is twice here.

ViniTou (commented):
I would just say that some filenames are kinda not related to whats inside as those docs from cache repo were kinda all mashed together.
Mostly in `symfony_reverse_proxy.md` where we have reverse proxy, fastly and captcha

DominikaK (changes requested):

DominikaK (approved):

ViniTou (approved):

PR #1405, opened 2021-06-30
Fix FT storage doc

DominikaK:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | n/a
| Versions      | all

Removes description of deprecated `eZ/Publish/Core/FieldType/StorageGateway.php`

Note: this was actually reported back in https://github.com/ezsystems/developer-documentation/pull/596/files?file-filters%5B%5D=.md&hide-deleted-files=true#r288964501, but got lost in the review.


Steveb-p:
This will require further clarification (from me, probably). Creating an external storage should also need tagging.

PR #1407, opened 2021-07-02
IBX-702: Added missing note about required normalizer tag in Migrations

Steveb-p:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | [IBX-702](https://issues.ibexa.co/browse/IBX-702)
| Versions      | 3.3

As discussed with @damianz5, we were actually missing an important piece of information regarding registration of an Action normalizer as a service. This PR remedies that.

### DIFF
--- a/docs/guide/data_migration.md
+++ b/docs/guide/data_migration.md
@@ -527,7 +527,7 @@ Then, tag the action denormalizer so it is recognized by the serializer used for
 ``` yaml
 services:
     Ibexa\Platform\Bundle\Migration\Serializer\Denormalizer\ContentType\Action\AssignContentTypeGroupActionDenormalizer:
-        # Autoconfigure set to false will prevent this normalizer from being registered in your application's serializer
+        # Make sure to set autoconfigure to false, otherwise this normalizer will be registered in your application's main serializer
         autoconfigure: false
         tags:
             - { name: 'ibexa.migrations.serializer.normalizer' }


damianz5 (approved):

Steveb-p:
Invoking the spirits of @adamwojs and @alongosz FYI and asking if maybe there's a way to do the autoconfiguration with auto-tagging of `serializer.normalizer` tag disabled :smile:

adamwojs (approved):

docs/guide/data_migration.md
```diff
@@ -514,6 +522,17 @@ final class AssignContentTypeGroupActionDenormalizer extends AbstractActionDenor
 
 ```
 
+Then the action denormalizer needs to be tagged to be recognized by serializer used for migrations.
```
DominikaK, line 525:
```suggestion
Then, tag the action denormalizer so it is recognized by the serializer used for migrations.
```

docs/guide/data_migration.md
```diff
@@ -514,6 +522,17 @@ final class AssignContentTypeGroupActionDenormalizer extends AbstractActionDenor
...
+
+``` yaml
+services:
+    Ibexa\Platform\Bundle\Migration\Serializer\Denormalizer\ContentType\Action\AssignContentTypeGroupActionDenormalizer:
+        # Autoconfigure set to false will prevent this normalizer from being registered in your application's serializer
```
DominikaK, line 530:
```suggestion
        # When autoconfigure is set to false, this normalizer is not registered in your application's serializer
```

DominikaK (changes requested):

docs/guide/data_migration.md
Steveb-p: True, but this is practically a warning - it's something the user **should not do** (setting `autoconfigure` to `true` in this service).

Usually it should not affect much, although it makes little sense to add Migration normalizers to the default application normalizer.

DominikaK: Ok, how about: `Make sure to set autoconfigure to false, otherwise...`?

Steveb-p: `Make sure to set autoconfigure to false, otherwise this normalizer will be registered in your application's main serializer`?

DominikaK: Yep, sounds good

DominikaK (approved):

Steveb-p:
@DominikaK I think that after Damian & Adam have done their reviews, the dev-side of things can be considered completed - if anyone had something to say about it / time to do the review, it was said.

PR #1412, opened 2021-07-07
Removed leftover `Platform` from package structure docs

Steveb-p:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | N/A
| Versions      | 3.3 :grey_question: / 4.0

Platforms are bad now, so I've removed the remaining platforms.

### DIFF
--- a/docs/community_resources/package_structure.md
+++ b/docs/community_resources/package_structure.md
@@ -32,20 +32,20 @@ The general package directory structure and corresponding PHP namespace mapping 
 ```
 .
 +-- src
-|   +-- bundle (`Ibexa\Platform\Bundle\<PackageName>`)
-|   +-- contracts (`Ibexa\Platform\Contracts\<PackageName>`)
-|   +-- lib (`Ibexa\Platform\<PackageName>`)
+|   +-- bundle (`Ibexa\Bundle\<PackageName>`)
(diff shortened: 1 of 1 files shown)


alongosz (approved):
> Platforms are bad now, so I've removed the remaining platforms.

Bad platform, go to the corner :rofl:

DominikaK (approved):

DominikaK:
@alongosz @Steveb-p, so is this 3.3 as well?

alongosz:
> @alongosz @Steveb-p, so is this 3.3 as well?

The answer to that question is complicated. I'm not 100% sure if we should change the 3.3 doc. In favor of that would be the fact that Platform namespace part is deprecated and doc should refer to the new "code". On the other hand due to BC this exist in "3.3-new" packages so might be confusing. Since the namespace part is deprecated, it falls under BC/FC[1] layer the same way as old eZ namespaces do. It means that new classes in 2.5 and 3.3 packages need to have non-Platform namespace

[1] forward compatibility layer, in 2.5 and 3.3

Steveb-p:
Maybe a note would be in order, that the `Platform` NS part will be removed in 4.0?

Otherwise this will look like an oversight on our part and this question will pop up again.

PR #1415, opened 2021-07-08
Legacy bridge support matrix removed

dabrt:
| Question      | Answer |
| ------------- | --- |
| JIRA Ticket   | none |
| Versions      | 2.5 and 1.13 |

Support of legacy bridge has ended in versions 2.5 and 1.13. Hence the removal of related information.


DominikaK (changes requested):
It's not the whole cloud that is unsupported, but only Legacy Bridge. And there are some more mentions of the Bridge higher up in the page.

DominikaK (approved):

PR #1418, opened 2021-07-12
Use Image instead of Image Asset for the tutorial

crevillo:
As part of my on boarding process I'm following tutorials in order to find things there. I found that if you follow strictly the `Ride` content type instructions, when adding an image to content of that Type you can see this kind of error

image

So, maybe a step to configure that is missed in the tutorial? I'm suggesting here to switch to `image` to keep it simple for now.


Steveb-p:
Looks like a frontend issue. @ezsystems/javascript-dev-team?

crevillo:
Looks to be a misleading error message only. I found the problem to be in my `public/var` folder. The web server didn't have permissions to write there. Once that is fixed everything goes as expected. I think we can close this one and maybe take this misleading message in account?

PR #1420, opened 2021-07-14
Misc cleanup

DominikaK:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | n/a
| Versions      | all

- Moves Service container page to APi section, and removes the uncomplete list of tags
- Removes a section in troubleshooting that referred to the old installation procedure

### DIFF
--- a/docs/api/service_container.md
+++ b/docs/api/service_container.md
@@ -9,19 +9,19 @@ It avoids the massive use of singletons, global variables or complicated factori
 The main issue with this pattern is how to resolve the dependencies for your services.
 This is where the service container comes into play. The role of a service container is to build and maintain your services and their dependencies.
 Each time you need a service, you may ask the service container for it.
-It either builds the service with the configuration you provided, or give you an existing instance if it is already available.
+It either builds the service with the configuration you provided, or gives you an existing instance if it is already available.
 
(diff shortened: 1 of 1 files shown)


docs/api/service_container.md
```diff
@@ -0,0 +1,32 @@
...
+
+!!! tip
+
+    To learn more about the service container, see the documentation for the [Symfony DependencyInjection component](http://symfony.com/doc/current/components/dependency_injection.html) documentation.
+
+## Service tags
+
+Service tags used by the Symfony service container are a way to dedicate services to a specific purpose. They are usually used for extension points.
+
+For example, if you want to register a [Twig extension](http://twig.sensiolabs.org/doc/advanced.html#creating-extensions) to add custom filters,
```
juskora, line 24: This link seems to be broken, I only managed to find this https://symfony.com/doc/current/templating/twig_extension.html but I'm not sure if it covers the topic at all.

juskora (changes requested):

docs/api/service_container.md
Steveb-p: It does, actually :+1:

DominikaK:
```suggestion
For example, if you want to register a [Twig extension](https://symfony.com/doc/current/templating/twig_extension.html) to add custom filters,
```

juskora (approved):

PR #1424, opened 2021-07-20
IBX-368: Image normalization command

DominikaK:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | https://issues.ibexa.co/browse/IBX-368
| Versions      | 2.5+

### DIFF
--- a/docs/guide/images.md
+++ b/docs/guide/images.md
@@ -196,7 +196,7 @@ Additionally you can provide two parameters:
 
 ## Normalizing images
 
-If you use image files with unprintable UTF-8 characters, you may come across a problem with images not displaying.
+If you use image files with unprintable UTF-8 characters in file names, you may come across a problem with images not displaying.
 In that case, run the `ezplatform:images:normalize-path` command to normalize them:
 
 ``` bash


mateuszbieniek (approved):

docs/guide/images.md
```diff
@@ -194,6 +194,15 @@ Additionally you can provide two parameters:
 
     This command publishes a new version of each Content item it modifies.
 
+## Normalizing images
+
+If you use image files with unprintable UTF-8 characters, you may come across a problem with images not displaying.
```
dabrt, line 199: Do these unprintable UTF-8 characters occur in file names or file paths?
Can we say:
"If you use image files with unprintable UTF-8 characters in file names, you might come across a problem with images not displaying."

dabrt (approved):

docs/guide/images.md
DominikaK: File names (when uploading images from external source), paths are handled by the platform. I'd assumed it's self-explanatory that we mean filesnames here, isn't it?

dabrt: I know I am nitpicking and I know we do not support vector files, but there could be a CDRm SVG, or any other vector file with with unprontable chars _inside_. This is why I wanted to stress the fact that we're talking about file names.
"If you use images with unprintable UTF-8 characters in file names..."

DominikaK: Ah, you're right, I didn't think about SVG-like files.
```suggestion
If you use image files with unprintable UTF-8 characters in file names, you may come across a problem with images not displaying.
```

PR #1431, opened 2021-08-04
Add warning about SSL with different SA domains

DominikaK:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | https://issues.ibexa.co/browse/IBX-375
| Versions      | 2.5+

### DIFF
--- a/docs/guide/multisite/multisite_configuration.md
+++ b/docs/guide/multisite/multisite_configuration.md
@@ -158,6 +158,12 @@ ezplatform:
                     - another_domain.org
 ```
 
+!!! caution "SiteAccess with separate admin domain"
+
+    If an admin SiteAccess in your installation uses a different domain than the front SiteAccesses,
+    be sure to use SSL (https protocol).
+    Otherwise, you cannot preview content in Page Builder from the Back Office.
+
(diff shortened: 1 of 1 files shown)


webhdx (approved):

dabrt (approved):

docs/guide/multisite/multisite_configuration.md
```diff
@@ -158,6 +158,12 @@ ezplatform:
                     - another_domain.org
 ```
 
+!!! caution "SiteAccess with separate admin domain"
+
+    If an admin SiteAccess in your installation uses a different domain than the front SiteAccesses,
+    be sure to use SSL (https protocol).
```
vidarl, line 164: I don't see why the fix for https://github.com/ezsystems/developer-documentation/pull/1431 has any realtion to https vs http..? This statement is simply false, or am I missing anything @webhdx (https://github.com/webhdx)

What is really important is that trusted_proxies are configured correctly if you have varnish, fastly or any HTTPS termination in front of the webserver..

webhdx: I don't exactly remember what was to reason to require SSL but it may be related to the frontend code.

vidarl: Okay. I tested both with HTTP and with HTTPS and I got both working.

PR #1439, opened 2021-08-16
Improve code highlight

DominikaK:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | n/a
| Versions      | all

Customizes syntax highlighting, as the built-in one wasn't distinguishing between some elements clearly, making PHP code poorly readable.
Also, bumps dependencies to latest versions.

**Reviewers**: please check whether code blocks are moderately readable for most languages and whether nothing unexpected breaks in the site.


dabrt (commented):
Nothing breaks, at least I wasn't able to find broken code snippets.
I am not completely sold on replacing green with red-ish and blue with green-ish, but I'll get used to it.
Text colors in snippets seem duller now, I preferred them brighter.
At the same time, methods in php are now bright red, which startled me a bit (is anything wrong with this method, I thought)
obraz
Can we ask UX team for an opinion on colors?

PR #1466, opened 2021-10-07
IBX-1187: Link to example vhost file

DominikaK:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | https://issues.ibexa.co/browse/IBX-1187
| Versions      | 3.3, master


DominikaK:
@mnocon Hm, but is the file itself acceptable? Then we can copy it to the doc repo.

PR #1473, opened 2021-10-19
Notify support about installation upgrade

glye:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | N/A
| Versions      | All supported

Support portals are often not updated when clients upgrade. That means support portal users are notified about releases that do not apply to them, and worse: They are **not** not about releases that **do** apply to them, including important security advisories. Upgrade docs should remind them about this.

This should also be added in relevant part(s) in the "Migrating to Ibexa DXP" doc. It is not relevant in "Update to v3.3.x". It's only relevant for major upgrades, like v3.2 to v3.3, or v3.3 to v4.

(Incidentally, I'm not sure if the use of "upgrade" and "update" is consistent. I used "update" since this is used directly above, but I personally prefer "upgrade" for installing a newer version. To me an "update" is a refresh that may or may not result in an actual change, but an "upgrade" is always a change to something newer or better.)

### DIFF
--- a/docs/updating/updating_to_3.3.md
+++ b/docs/updating/updating_to_3.3.md
@@ -181,6 +181,10 @@ php bin/console ibexa:graphql:generate-schema
 composer run post-install-cmd
 ```
 
+## Notify Support
+
+Please tell Support that you have updated your installation. They will update your support portal to match the new version. This ensures that you will receive notification about new maintenance releases and security advisories for the correct version. You can contact Support at support@ibexa.co.
+
 ## Update to v3.3.x
 
 !!! note


DominikaK:
@glye is email the recommended way to notify Support about it? Ot is there some way to do it via the Support portal?
Also, I guess this applies to all update procedures (also e.g. 1.13 -> 2.5)? Then I'll add it to all steps in https://github.com/ezsystems/developer-documentation/pull/1467, as that PR rewrites the whole update section.

glye:
Thanks, closing.

PR #1486, opened 2021-11-09
Example of using focal point

DominikaK:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | n/a
| Versions      | 3.3, master

Cf. https://github.com/ezsystems/user-documentation/pull/112

### DIFF
--- a/docs/guide/content_rendering/embed_and_list_content/render_images.md
+++ b/docs/guide/content_rendering/embed_and_list_content/render_images.md
@@ -71,7 +71,7 @@ To use it, select the variation when rendering the image:
 ## Use focal point
 
 In the [image editor](../../image_editor.md) you can define a focal point for an image.
-The focal point doesn't have an instant effect when you use the default templates.
+The focal point does not have an instant effect when you use the default templates.
 However, you can use it to select the part of the image the view focuses on when the image is cropped.
 
 The following example shows how to use an image contained in an Image Field as a focussed background.
@@ -88,7 +88,7 @@ First, in the main template, render the Image Field with a custom template:
 }) }}
 ```
 
-Then, create the custom Field template in `template/fields/image.html.twig`,
+Then, create the custom Field template in `templates/fields/image.html.twig`,
 [overriding the default `ezimage_field` template block](../render_content/render_content.md#field-templates):
 
 ``` html+twig


docs/guide/content_rendering/embed_and_list_content/render_images.md
```diff
@@ -67,3 +67,46 @@ To use it, select the variation when rendering the image:
...
+[overriding the default `ezimage_field` template block](../render_content/render_content.md#field-templates):
+
+``` html+twig
+{% block ezimage_field %}
+    {% if field.value.additionalData.focalPointX is defined and field.value.additionalData.focalPointY is defined %}
+        {% set position_x = (field.value.additionalData.focalPointX / field.value.width) * 100 %}
+        {% set position_y = (field.value.additionalData.focalPointY / field.value.height) * 100 %}
+    {% else %}
+        {% set position_x = 0 %}
+        {% set position_y = 0 %}
```
dew326, line 97: I think better would be the center of the image instead of the top left corner as default. So "50".

docs/guide/content_rendering/embed_and_list_content/render_images.md
```diff
@@ -67,3 +67,46 @@ To use it, select the variation when rendering the image:
...
 }) }}
 ```
+
+## Use focal point
+
+In the [image editor](../../image_editor.md) you can define a focal point for an image.
+The focal point doesn't have an instant effect when you use the default templates.
+However, you can use it to select the part of the image the view focuses on when the image is cropped.
+
+The following example shows how to use an image contained in an Image Field as a focussed background.
```
dew326, line 77: I would also mention that there are different approaches to focal point and JS libraries as well.

dew326 (changes requested):

dew326 (approved):

docs/guide/content_rendering/embed_and_list_content/render_images.md
```diff
@@ -67,3 +67,50 @@ To use it, select the variation when rendering the image:
...
+
+First, in the main template, render the Image Field with a custom template:
+
+``` html+twig
+{{ ez_render_field(content, 'image', {
+    'template': 'fields/image.html.twig'
+}) }}
+```
+
+Then, create the custom Field template in `template/fields/image.html.twig`,
```
juskora, line 91: is the path correct? Should it be template/fields... or templates/fields...?

docs/guide/content_rendering/embed_and_list_content/render_images.md
```diff
@@ -67,3 +67,50 @@ To use it, select the variation when rendering the image:
     }
 }) }}
 ```
+
+## Use focal point
+
+In the [image editor](../../image_editor.md) you can define a focal point for an image.
+The focal point doesn't have an instant effect when you use the default templates.
```
juskora, line 74:
```suggestion
The focal point does not have an instant effect when you use the default templates.
```

juskora (changes requested):

docs/guide/content_rendering/embed_and_list_content/render_images.md
DominikaK:
```suggestion
Then, create the custom Field template in `templates/fields/image.html.twig`,
```

juskora (approved):

juskora (approved):

PR #1487, opened 2021-11-09
Migration custom matchers

DominikaK:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | https://issues.ibexa.co/browse/IBX-1303
| Versions      | 3.3, master

**Note**: for review, please look exclusively at https://github.com/ezsystems/developer-documentation/commit/70603c51fea341cbe932539d0efb08c837962c83. The other commit only moves text sections around.

### DIFF
--- a/docs/guide/data_migration/add_data_migration_matcher.md
+++ b/docs/guide/data_migration/add_data_migration_matcher.md
@@ -4,7 +4,7 @@
 
 In addition to the built-in matchers, you can create custom matchers for content.
 
-The following example create a matcher for Section identifiers.
+The following example creates a matcher for Section identifiers.
 
 ## Create normalizer
 
--- a/docs/guide/data_migration/data_migration.md
+++ b/docs/guide/data_migration/data_migration.md
@@ -5,7 +5,7 @@ They can be useful for project templates. Migrations are able to store shared da
 or incrementally upgrade older projects to your new standard, if needed.
 They are a developer-friendly tool that allows you to share data without writing code.
 
-You can migrate your Repository data, that is Content items, as well as Content Types, languages, Object states, Sections, etc.,
+You can migrate your Repository data, that is Content items, as well as Content Types, languages, Object states, Sections, and so on,
 between installations by using the migration command.
 
 !!! tip "Public PHP API"


Steveb-p:
## Normalization

...is the more important part. This is what is actually used when migrating, so needs to be stressed out.
One would be perfectly fine just creating a normalizer for a `Criterion` (:exclamation: soon `FilteringCriterion` only :exclamation:).
Reason for this is that the `Criterion` that is created as part of denormalization will be passed directly into `SearchService` (:exclamation: soon `ContentService::find` :exclamation:).

To understand this, it's best to understand the serialization process in [Symfony Serializer](https://symfony.com/doc/current/components/serializer.html):
image
which is used extensively throughout the Ibexa Migrations.

## Generator

...is less important, because it is only used when exporting the data (creating migrations). It may be created, but is not necessary for migrations to work with new `match` type.

Basically Normalizer can live without Generator, while Generator cannot exist without a Normalizer (to convert the Criterion used for exporting into it's array representation).

code_samples/data_migration/config/custom_services.yaml
```diff
@@ -7,3 +7,11 @@ services:
     App\Migrations\Action\AssignSectionExecutor:
         tags:
             - { name: 'ibexa.migrations.executor.action.content', key: !php/const App\Migrations\Action\AssignSection::TYPE }
+
+    App\Migrations\Matcher\SectionIdentifierNormalizer:
+        tags:
+            - { name: 'ibexa.migrations.serializer.normalizer', priority: 10 }
```
Steveb-p, line 13: I think that all of our built-in normalizers are set to a negative value in `priority`, so user defined normalizers always go first unless specifically needed (and adjusted via `priority`). So it's cool to have the information that this can be used, but it's should not be needed.

If that's not the case then it's a bug.

docs/guide/data_migration/add_data_migration_matcher.md
```diff
@@ -0,0 +1,47 @@
...
+Register the normalizer as a service:
+
+``` yaml
+[[= include_file('code_samples/data_migration/config/custom_services.yaml', 10, 13) =]]
+```
+
+!!! note "Normalizer order"
+
+    User-defined normalizers are always executed before the built-in ones.
+    However, you can additionally set the priority of your normalizers.
```
Steveb-p, line 32: Normalization service priorities can be checked using `php bin/console debug:container --tag ibexa.migrations.serializer.normalizer`

docs/guide/data_migration/exporting_and_importing_data.md
```diff
@@ -0,0 +1,280 @@
...
+            translatable: true
+            category: ''
+            defaultValue: 'New article'
+            fieldSettings: {  }
+            validatorConfiguration:
+                StringLengthValidator:
+                    maxStringLength: 255
+                    minStringLength: null
+        # - ...
+    actions: {  }
```
Steveb-p, line 127: Equivalent to:
```suggestion
```

docs/guide/data_migration/add_data_migration_matcher.md
```diff
@@ -0,0 +1,53 @@
+# Add data migration matcher
+
+[Matchers in data migrations](exporting_and_importing_data.md#match-property) enable you to select which data from the Repository to export.
+
+In addition to the built-in matchers, you can create custom matchers for content.
+
+The following example create a matcher for Section identifiers.
```
Steveb-p, line 7: Matchers will be `FilteringCriterion`s soon (this is because migration process should not actually rely on SearchEngine). So I guess we should stress out that the Normalizer that will be created needs to `denormalize` into an instance of one.

See https://github.com/ibexa/migrations/pull/279

code_samples/data_migration/src/Migrations/Matcher/SectionIdentifierNormalizer.php
```diff
@@ -0,0 +1,31 @@
...
+    public function __construct()
+    {
+        parent::__construct('section_identifier');
+    }
+
+    /**
+     * @param array<mixed> $data
+     * @param array<mixed> $context
+     */
+    protected function createCriterion(array $data, string $type, ?string $format, array $context): Criterion
```
Steveb-p, line 20: Once https://github.com/ibexa/migrations/pull/279 is merged this will become:

```suggestion
    protected function createCriterion(array $data, string $type, ?string $format, array $context): FilteringCriterion
```

Note: this also applies to `use` statement at the beginning of the file - `FilteringCriterion` is in an entirely different namespace.

docs/guide/data_migration/exporting_and_importing_data.md
```diff
@@ -0,0 +1,280 @@
...
+        urlAliasSchema: ''
+        nameSchema: '<short_title|title>'
+        container: true
+        defaultAlwaysAvailable: false
+        defaultSortField: 1
+        defaultSortOrder: 1
+        translations:
+            eng-GB:
+                name: Article
+                description: ''
```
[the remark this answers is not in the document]
mateuszdebinski: I checked for CT, probably the content will not have this problem either

mateuszdebinski (approved):

docs/guide/data_migration/data_migration.md
```diff
@@ -0,0 +1,20 @@
+# Data migration
+
+You can use migrations in projects that require the same data to be present across multiple instances.
+They can be useful for project templates. Migrations are able to store shared data, so they can be applied for each new project you start,
+or incrementally upgrade older projects to your new standard, if needed.
+They are a developer-friendly tool that allows you to share data without writing code.
+
+You can migrate your Repository data, that is Content items, as well as Content Types, languages, Object states, Sections, etc.,
```
juskora, line 8:
```suggestion
You can migrate your Repository data, that is Content items, as well as Content Types, languages, Object states, Sections, and so on,
```

juskora (approved):

PR #1489, opened 2021-11-15
IBX-1410: Added UserId, UserEmail & UserLogin Criterions handling

DominikaK:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | https://issues.ibexa.co/browse/IBX-1410
| Versions      | 3.3, master

Note: Merge to 3.3 after next tag

### DIFF
--- a/docs/guide/search/criteria_reference/useremail_criterion.md
+++ b/docs/guide/search/criteria_reference/useremail_criterion.md
@@ -8,6 +8,10 @@ searches for content based on the email assigned to the User account.
 - `value` - string(s) representing the User email(s).
 - `operator` - operator constant (IN, EQ, LIKE).
 
+## Limitations
+
+Solr search engine and Elasticsearch support IN and EQ operators only.
+
 ## Example
 
 ``` php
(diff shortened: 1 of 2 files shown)


Steveb-p (approved):

adamwojs (changes requested):
You can get bonus point for explanation why only EQ and IN operation are supported

adamwojs (approved):

adamwojs:
> @adamwojs, does the why have practical consequences for someone using these Criteria?

One of the practical consequences is documented in https://github.com/ezsystems/developer-documentation/pull/1489/commits/8eb2282270c71cdeeced655ad04cfc5c1cee1e4b 

You can merge it without bonus points :stuck_out_tongue_winking_eye:

Steveb-p:
@DominikaK this makes it impossible for someone to make a partial search (i.e. try to find ones email by a few first letters, or anything like that).
You have to know the full email / login to be able to find Content associated with one's email / login.

This is done for privacy (and partially security) reasons, because Search engine can bypass permissions easily - I'm not sure it's insecure out of the box, because I'm not an expert.

DominikaK:
I clearly wasn't clear enough. I meant: does the user need to know **why** LIKE isn't supported, or is it enough for them to know **that** it isn't. But it seems we're good, as this is the way isn't done in the rest of the reference

juskora (approved):

PR #1492, opened 2021-11-16
GraphQL location handling

DominikaK:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | https://issues.ibexa.co/browse/EZP-32182
| Versions      | master

Note: the mutation part of the doc is not updated, due to https://issues.ibexa.co/browse/IBX-869

### DIFF
--- a/docs/api/graphql.md
+++ b/docs/api/graphql.md
@@ -65,8 +65,8 @@ You can also use your own [custom schema](graphql_customization.md#custom-schema
 GraphQL is SiteAccess-aware, but can have only one schema per installation.
 This means you cannot use GraphQL with multiple repositories.
 
-When you request a URL from a SiteAccess different that the current one,
-the API generates it for the item's SiteAccess, with an absolute URL if necessary.
+When you request a URL from a SiteAccess that is different than the current one,
+the API generates it for the Content item's SiteAccess, with an absolute URL if necessary.
 
 ## Authentication
 
--- a/docs/api/graphql_customization.md
+++ b/docs/api/graphql_customization.md
@@ -6,7 +6,8 @@ You can customize the GraphQL schema that is generated from your repository.
 
 You can use it if your application requires custom GraphQL resources, for instance for Doctrine entities.
 
-To do so, create an `config/graphql/Query.types.yaml` file. It will be used as the GraphQL query root.
+To do so, create an `config/graphql/Query.types.yaml` file. 
+The file will be used as the GraphQL query root.
 
 In that file, add new fields that use any custom type or custom logic you require, based
 on [overblog/GraphQLBundle](https://github.com/overblog/GraphQLBundle).
@@ -49,7 +50,7 @@ Query:
 ### Custom mutations
 
 Custom mutations are created in the same way as custom query configuration.
-An `config/graphql/Mutation.types.yaml` file will be used as the source for mutation definitions in your schema.
+A `config/graphql/Mutation.types.yaml` file will be used as the source for mutation definitions in your schema.
 
 ``` yaml
 Mutation:
--- a/docs/api/graphql_queries.md
+++ b/docs/api/graphql_queries.md
@@ -6,7 +6,7 @@ You can query a single Content item or a list of Content items using fields defi
 
 ### Get a Content item
 
-To get a specific Content item by its content, Location ID, or URL alias, use its relevant singular field,
+To get a specific Content item by its content ID, Location ID, or URL alias, use its relevant singular field,
 for example `article`, `folder`, `image`, etc.:
 
 ```
@@ -245,7 +245,7 @@ When you use `_location`, the API returns:
 
 - the Location specified in the `locationId` or `urlAlias` argument
 - the Location based on the current SiteAccess
-- the main Location.
+- the main Location
 
 ```
 {


bdunogier:
This one confuses me: I don't think it covers the actual location handling part of the feature, as described in https://issues.ibexa.co/browse/EZP-32182. Isn't it rather about other changes, such as renaming `id` to `contentId`, or the path changes ?

docs/api/graphql_queries.md
```diff
@@ -43,6 +43,30 @@ Response:
 
 You can request any Fields of the Content item. In the example above, these are `title` and `author`.
 
+You can also query the generic `item` object. The `item` object references either a Content item or a Location.
```
bdunogier, line 46: That is actually a bit incorrect: an item is mostly a content, and location data can be requested for it. If it was loaded using a locationId or an urlAlias, it will refer to that specific location. If it was loaded using a content id, the location will be guessed automatically.

The same thing is true for any specific type (article, folder...)

docs/api/graphql_queries.md
```diff
@@ -183,18 +207,50 @@ Response:
 
 ## Querying Locations
 
-!!! tip
+You can get the Location object from any Content item object by querying for `_location` or `_allLocations`.
+When you use `_location`, the API picks the Location to return based on the current SiteAccess, or returns the main one.
```
bdunogier, line 211: Or (if my memory serves me right) the location that was specified if either `locationId` or `urlAlias` were used as arguments.

docs/api/graphql_queries.md
```diff
@@ -183,18 +207,50 @@ Response:
...
 
-    To make use of enhanced Location handling, you can add the beta 3.0 version of [ezplatform-graphql](https://github.com/ezsystems/ezplatform-graphql/tree/3.0) to your project.
+```
+{
+  "data": {
+    "content": {
+      "folder": {
+        "_allLocations": [
+          {
+            "id": 132
```
bdunogier, line 234: I'd use pathString or alias instead of id, it looks better as an example.

docs/api/graphql_queries.md
```diff
@@ -183,18 +207,50 @@ Response:
...
+    }
+  }
+}
+```
 
-    See [overview of the upcoming changes](https://github.com/ezsystems/ezplatform-graphql/pull/90).
+## Getting children of a Location
 
-To query a Location and its children, use the repository schema:
+To get a Location's children, query the `children` property of an `item` or `content` object:
```
bdunogier, line 248: It would probably be worth mentioning that query fields are still preferred over that method :)

We may also want to remind that the location that is used to retrieve children is either automatic (if a content id was specified, or in case of a list of items), or manual (if a location id or url alias was used to find the item).

bdunogier (commented):
I had left unpublished comments, so I'll publish those right away, and then have a new look at the PR. Sorry about that.

docs/api/graphql_queries.md
```diff
@@ -216,18 +240,54 @@ Response:
 
 ## Querying Locations
 
-!!! tip
+You can get the Location object from any item by querying for `_location` or `_allLocations`.
+When you use `_location`, the API returns:
```
bdunogier, line 244: We may also want to document the `_url` field, that returns an URL alias for the content item. It behaves like `_location`, and guesses the "best" location based on a set of rules (it is one of the core features of that change).

docs/api/graphql_queries.md
```diff
@@ -216,18 +240,54 @@ Response:
...
+        ]
+      }
+    }
+  }
+}
+```
+
+## Getting children of a Location
+
+To get a [Location's](#querying-locations) children, query the `children` property of an `item` or `content` object:
```
bdunogier, line 285: I am not a big fan of documenting that approach as a first-class citizen for getting children. The API it exposes differs from the `item` one (since it is based on "location" and then "content", not "item"), and it will make queries weird.

In addition, the "location" depends on quite a few factors (siteaccess, to begin with, but also the choice of the location if there are several in the current site's tree root).

For these reasons, I'd strongly suggest to use query fields rather than the "raw" children API, as these will use proper queries and can take into account the dynamic location handling introduced by that change.

docs/api/graphql.md
```diff
@@ -66,11 +60,14 @@ The `_repository` field also enables you to query e.g. Object states configured
...
 You can also use your own [custom schema](graphql_customization.md#custom-schema).
 
-### Multiple repositories
+### SiteAccesses and multiple repositories
 
 GraphQL is SiteAccess-aware, but can have only one schema per installation.
 This means you cannot use GraphQL with multiple repositories.
 
+When you request a URL from a SiteAccess different that the current one,
+the API generates it for the item's SiteAccess, with an absolute URL if necessary.
```
dabrt, line 69:
```suggestion
When you request a URL from a SiteAccess that is different than the current one,
the API generates it for the Content item's SiteAccess, with an absolute URL if necessary.
```

docs/api/graphql_customization.md
```diff
@@ -6,7 +6,7 @@ You can customize the GraphQL schema that is generated from your repository.
 
 You can use it if your application requires custom GraphQL resources, for instance for Doctrine entities.
 
-To do so, create an `app/config/graphql/Query.types.yaml` file. It will be used as the GraphQL query root.
+To do so, create an `config/graphql/Query.types.yaml` file. It will be used as the GraphQL query root.
```
dabrt, line 9:
```suggestion
To do so, create an `config/graphql/Query.types.yaml` file. 
The file will be used as the GraphQL query root.
```

docs/api/graphql_customization.md
```diff
@@ -49,7 +49,7 @@ Query:
 ### Custom mutations
 
 Custom mutations are created in the same way as custom query configuration.
-An `app/config/graphql/Mutation.types.yaml` file will be used as the source for mutation definitions in your schema.
+An `config/graphql/Mutation.types.yaml` file will be used as the source for mutation definitions in your schema.
```
dabrt, line 52:
```suggestion
A `config/graphql/Mutation.types.yaml` file will be used as the source for mutation definitions in your schema.
```

docs/api/graphql_queries.md
```diff
@@ -6,13 +6,13 @@ You can query a single Content item or a list of Content items using fields defi
 
 ### Get a Content item
 
-To get a specific Content item by its ID, use its relevant singular field,
+To get a specific Content item by its content, Location ID, or URL alias, use its relevant singular field,
```
dabrt, line 9:
```suggestion
To get a specific Content item by its content ID, Location ID, or URL alias, use its relevant singular field,
```

docs/api/graphql_queries.md
```diff
@@ -216,18 +240,84 @@ Response:
...
 ## Querying Locations
 
-!!! tip
+You can get the Location object from any item by querying for `_location` or `_allLocations`.
+When you use `_location`, the API returns:
 
-    To make use of enhanced Location handling, you can add the beta 3.0 version of [ezplatform-graphql](https://github.com/ezsystems/ezplatform-graphql/tree/3.0) to your project.
+- the Location specified in the `locationId` or `urlAlias` argument
+- the Location based on the current SiteAccess
+- the main Location.
```
dabrt, line 248:
```suggestion
- the main Location
```

dabrt (changes requested):

dabrt (approved):

PR #1493, opened 2021-11-17
Fix Aggregation IntegerStatsAggregation example

arfaram:
Add the right Aggregation field class

Provide some getter example

| Question      | Answer
| ------------- | ---
| JIRA Ticket   | 
| Versions      | 3.3+

### DIFF
--- a/docs/guide/search/aggregation_reference/integerstats_aggregation.md
+++ b/docs/guide/search/aggregation_reference/integerstats_aggregation.md
@@ -19,5 +19,6 @@ and provides statistical information for the values including:
 
 ``` php
 $query = new Query();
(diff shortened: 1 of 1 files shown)


DominikaK:
@arfaram, does the added note apply only to this aggregation, or to all Stats aggregations?

arfaram:
> @arfaram, does the added note apply only to this aggregation, or to all Stats aggregations?

@DominikaK Indeed, float and raw stats aggregation have the same statistical information

DominikaK:
How about including the method names in the bullet points in lines 6-10 of this page (and the two other stats-pages), to keep the info in one place?

adamwojs (approved):

DominikaK:
Replaced by https://github.com/ezsystems/developer-documentation/pull/1518

PR #1552, opened 2022-03-10
Custom product Content Type clarification

DominikaK:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | n/a
| Versions      | 3.3

### DIFF
--- a/docs/guide/catalog/product.md
+++ b/docs/guide/catalog/product.md
@@ -44,15 +44,15 @@ silversolutions_eshop:
 ```
 
 The custom Content Type must mirror the structure of the built-in `ses_product` Content Type,
-including having Fields with the same identifiers,
+by having Fields with the same identifiers,
 but you can add other Fields to it.
 
 !!! tip
 
-    To ensure all the Fields are set up correctly, you can copy the built-in Product Content Type
+    To ensure that all the Fields are set up correctly, you can copy the `ses_product` Content Type
     and add your custom Field to the copy.
 
-Additionally add the following parameters for your custom Content Type's identifier:
+Additionally, add the following parameters for your custom Content Type's identifier:
 
 ``` yaml
 parameters:


reithor (approved):

docs/guide/catalog/product.md
```diff
@@ -43,12 +43,23 @@ silversolutions_eshop:
             - my_custom_product_type
```
dabrt, line 43: Should there be a minus before "my_custom_product_type"? 
`yaml` styling interprets it as substraction and colors the line red while we just said that we want to add this identifier to the config file.

docs/guide/catalog/product.md
```diff
@@ -43,12 +43,23 @@ silversolutions_eshop:
...
 ```
 
-Additionally, to enable buying the product in the shop,
-add the following parameter for your custom Content Type's identifier:
+The custom Content Type must mirror the structure of the built-in `ses_product` Content Type,
+including having Fields with the same identifiers,
```
dabrt, line 47: Is it "the structure that includes the Fields with the same identifiers", or is it "by including/having the Fields with the same IDs", or would it be just "The custom Content Type must include all the Fields with the same identifiers as the built-in `ses_product` Content Type, ..."?

docs/guide/catalog/product.md
```diff
@@ -43,12 +43,23 @@ silversolutions_eshop:
...
+but you can add other Fields to it.
+
+!!! tip
+
+    To ensure all the Fields are set up correctly, you can copy the built-in Product Content Type
```
dabrt, line 52:
```suggestion
    To ensure that all the Fields are set up correctly, you can copy the `ses_product` Content Type
```

docs/guide/catalog/product.md
```diff
@@ -43,12 +43,23 @@ silversolutions_eshop:
...
+    and add your custom Field to the copy.
+
+Additionally add the following parameters for your custom Content Type's identifier:
```
dabrt, line 55:
```suggestion
Additionally, add the following parameters for your custom Content Type's identifier:
```

dabrt (approved):

docs/guide/catalog/product.md
DominikaK: That's part of YAML's array syntax: https://www.w3schools.io/file/yaml-arrays/, see for example: https://github.com/ezsystems/ezcommerce-shop/blob/7169ce3dfbafca67f696b2009209d83414207bb8/src/Silversolutions/Bundle/EshopBundle/Resources/config/product_settings.yaml#L3

docs/guide/catalog/product.md
DominikaK, line 47:
```suggestion
by having Fields with the same identifiers,
```

PR #1564, opened 2022-03-23
Reorder steps for 3.3 update

DominikaK:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | 
| Versions      | all


barw4 (approved):

barw4:
As a minor note, we could link the latest v3.3 releases somewhere, https://github.com/ibexa/oss/releases?q=v3.3.&expanded=true. That way we wouldn't have to change documentation every time new patch version is up,

DominikaK:
> That way we wouldn't have to change documentation every time a new patch version is up
We still have to update the variables with every version so that the commands are up to date

PR #1569, opened 2022-03-25
IBX-2558: Data migration examples

DominikaK:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | https://issues.ibexa.co/browse/IBX-2558
| Versions      | 4.0, master

Example of product creation with attribute values to be added later.

Also extracts earlier examples to separate files and places the Data migration one level up in TOC.

### DIFF
--- a/docs/guide/data_migration/importing_data.md
+++ b/docs/guide/data_migration/importing_data.md
@@ -1,19 +1,20 @@
 # Importing data
 
-To import data from YAML migration files into Repository, run the `ibexa:migrations:migrate` command.
+To import data from YAML migration files into Repository, you run the `ibexa:migrations:migrate` command.
 
-You can use the `ibexa:migrations:import` command to automatically place migration files in the correct folder.
+The `ibexa:migrations:import` command automatically places migration files in the correct folder.
 
 Alternatively, you can place the files manually in the `src/Migrations/Ibexa/migrations` folder
-or in [a custom folder that you configure](migration_management.md#migration-folders).
-The command takes the file name within this folder as parameter.
-If file is not specified, all files within this directory are used.
+or in [a custom folder that you configure](migration_management.md#migration-folders), 
+and specify the file name within this folder as parameter.
+If you do not specify the file, all files within this directory are used.
 
 ``` bash
 php bin/console ibexa:migrations:migrate --file=my_data_export.yaml
 ```
 
-Ibexa Migrations store execution metadata in `ibexa_migrations` database table. This allows incremental upgrades:
+Migrations store execution metadata in the `ibexa_migrations` database table. 
+This allows incremental upgrades:
 the `ibexa:migration:migrate` command ignores files that it had previously executed.
 
 
@@ -31,7 +32,7 @@ The required metadata keys are: `identifier`, `mainTranslation`, `contentTypeGro
 [[= include_file('code_samples/data_migration/examples/create_blog_post_ct.yaml') =]]
 ```
 
-### Content
+### Content items
 
 The following example shows how to create two Content items: a folder and an article inside it.
 
@@ -40,7 +41,7 @@ as well as `parentLocationId`.
 
 To use the Location ID of the folder, which is created automatically by the system,
 you can use a [reference](migration_management.md#references).
-In this case you assign the `parent_folder_location_id` reference name to the Location ID
+In this case you assign the `parent_folder_location_id` reference name to the Location ID,
 and then use it when creating the article.
 
 ``` yaml hl_lines="15 24"
@@ -52,7 +53,7 @@ and then use it when creating the article.
 The following example shows how to create a Role.
 A Role requires the `identifier` metadata key.
 
-For each Policy assigned to the Role you select the module and function, with optional Limitations.
+For each Policy assigned to the Role, you select the module and function, with optional Limitations.
 
 ``` yaml
 [[= include_file('code_samples/data_migration/examples/create_role.yaml') =]]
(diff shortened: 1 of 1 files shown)


code_samples/data_migration/examples/create_attribute_group.yaml
```diff
@@ -0,0 +1,24 @@
...
+-   type: attribute
+    mode: create
+    identifier: color
+    attribute_type_identifier: selection
+    attribute_group_identifier: hat
+    names:
+        eng-GB: Color
+    options:
+        choices:
+            - { value: red, label: { "eng-GB": "Red"}}
```
Steveb-p, line 22: Just an internal note, but this might not be obvious:

```suggestion
            -    value: red
                 label: 
                     "eng-GB": "Red"
```
is also a valid notation (which may provide better readability in some cases, especially with long data).

code_samples/data_migration/examples/create_currency.yaml
```diff
@@ -0,0 +1,4 @@
+-   type: currency
+    mode: create
+    code: TST
+    subunits: 3
```
Steveb-p, line 4:
```suggestion
    subunits: 3
    enabled: true # default, optional
```

docs/guide/data_migration/importing_data.md
```diff
@@ -0,0 +1,165 @@
+# Importing data
+
+To import Repository data from YAML files, run the `ibexa:migrations:migrate` command.
```
Steveb-p, line 3: Strictly speaking this is correct, although a different command exists (`ibexa:migrations:import` that allows "importing" files from elsewhere in the system into Migrations directory). Describing both as import might be confusing for the user.

Therefore, to more clearly show what data is manipulated, I'd suggest:
```suggestion
To import data from YAML migration files into Repository, run the `ibexa:migrations:migrate` command.
```

This comment might also affect the paragraph below, as ideally one would use the aforementioned command to import migration files, as it handles picking the right directory for him/her.

docs/guide/data_migration/importing_data.md
```diff
@@ -0,0 +1,165 @@
...
+
+The following example shows how to create a currency:
+
+``` yaml
+[[= include_file('code_samples/data_migration/examples/create_currency.yaml') =]]
+```
+
+## Criteria
+
+When using `update` or `delete` modes, you can use criteria to identify the objects to operate on.
```
Steveb-p, line 121: A word of warning: those criteria only work for objects introduced in Product Catalog as of today. Other migrations have their criteria written explicitly, since for them there is no underlying framework that would allow for searching.

Search engine does exist, but has been proven that it's use for migrations might cause data integrity issues (due to how Solr work, especially). Because of this, Content-based migrations do make use of Content filtering, and Product Catalog migrations use filtering that has been baked into all new objects.

code_samples/data_migration/examples/create_attribute_group.yaml
```diff
@@ -0,0 +1,32 @@
...
+        choices:
+            -    value: red
+	                 label: 
+	                     "eng-GB": "Red"
+            -    value: white
+	                 label: 
+	                     "eng-GB": "White"
+            -    value: black
+	                 label: 
+	                     "eng-GB": "Black"
```
Steveb-p, line 32:
```suggestion
            -    value: red
	         label: 
	             "eng-GB": "Red"
            -    value: white
	         label: 
	             "eng-GB": "White"
            -    value: black
	         label: 
	             "eng-GB": "Black"
```

Steveb-p (changes requested):

Steveb-p (approved):

dabrt (approved):

docs/guide/data_migration/importing_data.md
```diff
@@ -0,0 +1,171 @@
+# Importing data
+
+To import data from YAML migration files into Repository, run the `ibexa:migrations:migrate` command.
+
+You can use the `ibexa:migrations:import` command to automatically place migration files in the correct folder.
```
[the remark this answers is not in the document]
Steveb-p: `import` command does indeed **place** the files in the migrations folder.

Command cannot be run without parameters (it will throw an error if tried), but `--from-stdin` option can be used to instruct it to read the file from the standard input instead (which means some other script can generate the file and pipe it into the command: `generate.sh > php bin/console ibexa:migrations:import --from-stdin` (haven't tested the exact instructions, but you get the idea)).

And you understand the command correctly. It's supposed to be used to pick YAML migration files from elsewhere and place them in the right directory. It's primary reason for existence is that this directory can be configured to be elsewhere (if not right now, it is supposed to be configurable).

docs/guide/data_migration/importing_data.md
```diff
@@ -0,0 +1,171 @@
...
+
+``` yaml
+[[= include_file('code_samples/data_migration/examples/create_attribute_group.yaml') =]]
+```
+
+#### Product type
+
+The following example shows how to create a product type.
+
+The main part of the migration file is the same as when creating a regular Content item.
```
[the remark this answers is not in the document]
Steveb-p: In Product Catalog, they exist as `ProductType` class. They should follow the Content convention though.

Internally they do match with Content Type in the current iteration. That is, every Product Type has a matching Content Type. At least for now.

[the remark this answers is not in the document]
DominikaK: Throughout the doc, I used "product", "product type": https://doc.ibexa.co/en/master/guide/catalog/catalog/#products. Agree with consistency, but I'd rather go the other way round (de-capitalize Content Types). But that's a story for another PR.

PR #1575, opened 2022-04-04
Added instructions how to create a migration step

Steveb-p:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | N/A
| Versions      | 3.3, 4.0

Code used to create migration step: `ImATeapot` :sweat_smile:

### DIFF
--- a/code_samples/data_migration/config/custom_services.yaml
+++ b/code_samples/data_migration/config/custom_services.yaml
@@ -19,7 +19,7 @@ services:
     App\Migrations\Step\ReplaceNameStepNormalizer:
         tags:
             - 'ibexa.migrations.serializer.step_normalizer'
-            - { name: 'ibexa.migrations.serializer.normalizer', priority: 9000 }
+            - 'ibexa.migrations.serializer.normalizer'
 
     App\Migrations\Step\ReplaceNameStepExecutor:
         tags:
--- a/code_samples/data_migration/src/Migrations/Step/ReplaceNameStepExecutor.php
+++ b/code_samples/data_migration/src/Migrations/Step/ReplaceNameStepExecutor.php
@@ -38,8 +38,9 @@ final class ReplaceNameStepExecutor extends AbstractStepExecutor
                     continue;
                 }
 
-                if ($field->value == 'Company Name') {
-                    $struct->setField($field->fieldDefIdentifier, new Value($step->getReplacement()));
+                if (str_contains($field->value, 'Company Name')) {
+                    $newValue = str_replace('Company Name', $step->getReplacement(), $field->value);
+                    $struct->setField($field->fieldDefIdentifier, new Value($newValue));
                 }
             }
 
--- a/code_samples/data_migration/src/Migrations/Step/ReplaceNameStepNormalizer.php
+++ b/code_samples/data_migration/src/Migrations/Step/ReplaceNameStepNormalizer.php
@@ -40,7 +40,7 @@ final class ReplaceNameStepNormalizer extends AbstractStepNormalizer
 
     public function getType(): string
     {
-        return 'name';
+        return 'company_name';
     }
 
     public function getMode(): string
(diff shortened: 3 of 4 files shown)


code_samples/data_migration/src/Migrations/Step/ImATeapotStepExecutor.php
```diff
@@ -0,0 +1,62 @@
...
+
+/**
+ * @copyright Copyright (C) Ibexa AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+declare(strict_types=1);
+
+namespace App\Migrations\ImATeapot;
+
+use eZ\Publish\Core\FieldType\TextLine\Value;
```
Steveb-p, line 11:
```suggestion
use Ibexa\Core\FieldType\TextLine\Value;
```

code_samples/data_migration/config/custom_services.yaml
```diff
@@ -15,3 +15,12 @@ services:
     App\Migrations\Matcher\SectionIdentifierGenerator:
         tags:
             - { name: 'ibexa.migrations.generator.criterion_generator.content' }
+
+    App\Migrations\Step\ReplaceNameStepNormalizer:
+        tags:
+            - 'ibexa.migrations.serializer.step_normalizer'
+            - { name: 'ibexa.migrations.serializer.normalizer', priority: 9000 }
```
Steveb-p, line 22: This was a joke, priority can be 0. All our normalizers have negative priority, so user defined go first unless they also have lower priority.

```suggestion
            - 'ibexa.migrations.serializer.normalizer'
```

code_samples/data_migration/src/Migrations/Step/ReplaceNameStepNormalizer.php
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    public function getHandledClassType(): string
+    {
+        return ReplaceNameStep::class;
+    }
+
+    public function getType(): string
+    {
+        return 'name';
```
Steveb-p, line 43:
```suggestion
        return 'company_name';
```
would make it less likely to collide with a different migration type (it's unlikely as it is, but...)

docs/guide/data_migration/create_migration_step.md
```diff
@@ -0,0 +1,55 @@
...
+Tag the executor with `ibexa.migrations.step_executor` tag.
+
+```yaml
+[[= include_file('code_samples/data_migration/config/custom_services.yaml', 23, 27) =]]
+```
+
+Then you can create a migration file that represents this step in your application:
+
+```yaml
+-   type: name
```
Steveb-p, line 52:
```suggestion
-   type: company_name
```

code_samples/data_migration/src/Migrations/Step/ReplaceNameStepExecutor.php
```diff
@@ -0,0 +1,60 @@
...
+                    continue;
+                }
+
+                if ($field->fieldDefIdentifier === 'identifier') {
+                    continue;
+                }
+
+                if ($field->value == 'Company Name') {
+                    $struct->setField($field->fieldDefIdentifier, new Value($step->getReplacement()));
+                }
```
Steveb-p, line 43:
```suggestion
                if ($field->value === 'Company Name') {
                    $struct->setField($field->fieldDefIdentifier, new Value($step->getReplacement()));
                }
```
We could also check & replace if value contains such a string, not only equals:
```suggestion
                if (str_contains($field->value, 'Company Name')) {
                    $newValue = str_replace('Company Name', $step->getReplacement(), $field->value);
                    $struct->setField($field->fieldDefIdentifier, new Value($newValue));
                }
```

(`str_contains` is a PHP8 only function, but it is contained in Symfony's PHP8 polyfill)

PR #1596, opened 2022-04-20
Elastic version bump

dabrt:
| Question      | Answer |
| ------------- | --- |
| JIRA Ticket   | IBX-1699 |
| Versions      | 3.3 and up |

### DIFF
--- a/docs/getting_started/requirements.md
+++ b/docs/getting_started/requirements.md
@@ -56,7 +56,7 @@ event MPM is recommended, if you need to use prefork you also need the `mod_php`
 ## Search
 
 - Solr 7.7LTS or Solr 8
-- Elasticsearch 7.7, using Oracle Java/Open JDK 8 or higher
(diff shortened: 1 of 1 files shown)


barw4 (approved):

glye (commented):
JDK should also be upgraded, to 11 imho, see https://developers.ibexa.co/security-advisories/cve-2021-44228-log4j-vulnerability

I'm not sure this qualifies as a requirement, but the Log4j bug that affected Elasticsearch was only exploitable with JDK 8. 9 is unsupported. 11 is a supported LTS release, it would be good to recommend this. 17 is a newer LTS, but I hear we're not quite compatible with that, so 11 is best. https://www.oracle.com/java/technologies/java-se-support-roadmap.html

glye:
Follow up PR on Java: https://github.com/ezsystems/developer-documentation/pull/1601

PR #1599, opened 2022-04-25
IBX-2640 - GraphQL filtering by attributes and content fields

dabrt:
| Question      | Answer |
| ------------- | --- |
| JIRA Ticket   | IBX-2640 |
| Versions      | 4.1 and up |

Add filter by CFs and attributes description

### DIFF
--- a/docs/api/graphql_queries.md
+++ b/docs/api/graphql_queries.md
@@ -582,7 +582,6 @@ To filter products based on content fields:
 
 To filter products based on attributes:
 
-
 ```
 {
   products {


docs/api/graphql_queries.md
```diff
@@ -553,6 +553,79 @@ Response:
...
+          }
+        }
+      }
+    }
+  }
+}
+
+```
+ 
+If the attribute type cannot be found in the schema, the response is:
```
konradoboza, line 607: Nitpicking: given we return two attributes, I would personally rephrase it to be more specific, something like:
```suggestion
If the attribute type cannot be found (`measure` in this case) in the schema, the response is:
```

konradoboza (approved):

docs/api/graphql_queries.md
```diff
@@ -553,6 +553,79 @@ Response:
...
+          }
+        }
+      }
+    }
+  }
+}
+```
+
+To filter products based on attributes:
+
```
juskora, line 584:
```suggestion
```

juskora (approved):

PR #1606, opened 2022-05-02
Add instructions on encoding database password

glye:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | N/A
| Versions      | Seems limited to Symfony 5 (not sure), e.g. DXP v3.3 and v4

Based on this community question: https://ezcommunity.slack.com/archives/C032PD5N9UY/p1651091978286979
we should document that database passwords has certain limitations. I'm following this recommendation: https://github.com/symfony/symfony/issues/35568#issuecomment-792657130
and adding the alternative of simply using chars that require no encoding.

I am not sure if using headers inside the tip box works as intended. Also not sure how much entropy discussion is worthwhile here, we want to keep it short but understandable.

### DIFF
--- a/docs/getting_started/troubleshooting.md
+++ b/docs/getting_started/troubleshooting.md
@@ -4,17 +4,24 @@ This page lists potential problems that you may encounter while installing, conf
 
 ## Encoding database password
 
-The password entered in `DATABASE_URL` must either be URL encoded, or not contain any special characters that would require URL encoding.
+The password entered in `DATABASE_URL` during installation must either be URL encoded,
+or not contain any special characters that would require URL encoding.
 
-#### Alternative 1: URL encoding
+### URL encoding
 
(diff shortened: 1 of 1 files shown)


carlosmauri:
I'm pretty sure this will help/prevent others to spent time debugging the issue.
Thanks for documenting it @glye !

DominikaK:
@glye before proofing: do you think we could move the bulk of the text to https://doc.ibexa.co/en/master/getting_started/troubleshooting, and leave only a mention and link on this page? That would prevent overfilling the install guide with too many "special concerns".

glye:
@DominikaK Sure, that would be fine.

adamwojs (approved):

DominikaK (approved):

carlosmauri:
Thanks for the work on this guys 👏

PR #1607, opened 2022-05-03
Fix version specific notes

reithor:
Fix version specific notes for releases  **<=v2.1.1** and **<=v2.4.1**

### DIFF
--- a/docs/updating/from_1.x_2.x/update_app_to_2.5.md
+++ b/docs/updating/from_1.x_2.x/update_app_to_2.5.md
@@ -13,16 +13,16 @@ latest_tag: '2.5.27'
 
 [[% include 'snippets/update/merge_composer.md' %]]
 
-## 3. Update the app
+## 3. Version Specific changes
 
-First, perform version-specific steps depending on which version you are updating from.
+When updating a release **earlier than v2.5.0**, some version-specific changes are required.
 
-### A. v2.2
+### A. Update from <=v2.1.1
     
-When upgrading an Enterprise installation to v2.2, you need to disable `EzSystemsPlatformEEAssetsBundle` by removing
+When updating from a release **<=v2.1.1** you need to disable `EzSystemsPlatformEEAssetsBundle` by removing
 `new EzSystems\PlatformEEAssetsBundle\EzSystemsPlatformEEAssetsBundle(),` from `app/AppKernel.php`.
 
-### B. v2.5
+### B. Update from <=v2.4.2
 
 Since v2.5 eZ Platform uses [Webpack Encore]([[= symfony_doc =]]/frontend.html#webpack-encore) for asset management.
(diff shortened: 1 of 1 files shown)


docs/updating/from_1.x_2.x/update_app_to_2.5.md
```diff
@@ -13,16 +13,16 @@ latest_tag: '2.5.27'
...
 [[% include 'snippets/update/merge_composer.md' %]]
 
-## 3. Update the app
+## 3. Version Specific changes
 
-First, perform version-specific steps depending on which version you are updating from.
+When updating a release **earlier than v2.5.0**, some version-specific changes are required.
 
-### A. v2.2
+### A. Update from <=v2.1.1
```
adriendupuis, line 20: I would prefer
```suggestion
### A. Update from earlier than v2.2
```
If not, maybe the use of  `≤` instead of `<=`
```suggestion
### A. Update from ≤v2.1.1
```

reithor: - I used "<="  because I have seen it here https://doc.ibexa.co/en/2.5/updating/from_1.x_2.x/update_app_to_2.5/#c-prepare-for-pulling-changes
- "earlier than" is less specific than "<=v2.1.1"

but I am definitely fine with "from earlier than v2.2" (or v2.2.0)

docs/updating/from_1.x_2.x/update_app_to_2.5.md
```diff
@@ -13,16 +13,16 @@ latest_tag: '2.5.27'
 
 [[% include 'snippets/update/merge_composer.md' %]]
 
-## 3. Update the app
+## 3. Version Specific changes
 
-First, perform version-specific steps depending on which version you are updating from.
+When updating a release **earlier than v2.5.0**, some version-specific changes are required.
```
DominikaK, line 18: @reithor, in general, can you explain the reason for this change? We use the same structure for all update pages (see https://doc.ibexa.co/en/master/updating/from_3.3/update_from_3.3/#v332 for example), so changing it here would require reformulating the headings everywhere in the update section for consistency.
If I understand correctly and the issue is: "we are updating **to** v2.2, not **from** v2.2", I'd prefer rephrasing this introductory sentence:
```suggestion
First, perform version-specific steps depending on which version you are updating to.
```

reithor: @DominikaK : changing the structure is not neccessary - I just thought it woul be clearer.
The whole point of the PR is that :
"A. v2.2" --> should be "A. from earlier than v2.2"
"B. v2.5" --> should be "B. from earlier than v2.5"

reithor:
replaced by https://github.com/ezsystems/developer-documentation/pull/1609

PR #1611, opened 2022-05-05
Fix APP_ENV for vhost setup

reithor:
Symfony is using APP_ENV instead of SYMFONY_ENV since Symfony Flex. (applies for >=v3.3)
See also https://github.com/ezsystems/developer-documentation/blob/master/code_samples/install/vhost_template/vhost.template


glye (approved):
Good. Only question is if there may be other doc pages with the same issue? But I didn't find any.

glye (approved):

dabrt (approved):

PR #1624, opened 2022-05-20
Autolinks plugin

DominikaK:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | 
| Versions      | all

Enable [mkdocs-autolinks-plugin](https://github.com/midnightprioriem/mkdocs-autolinks-plugin) to make linking between pages easier and moving content about the file structure less painful.

Both old and new link formats will work, so switching to new format can happen gradually.

Note that this plugin works only for _internal_ links. It generates correct links from markdown in html. Links from external sources still need to contain a full path.

## Potential issues

The plugin works with images, but not when they have a title in the syntax: `Alt`. This is reported but unresolved: https://github.com/midnightprioriem/mkdocs-autolinks-plugin/issues/8

To do:
- [x] Test with deeply nested links (ref. https://github.com/midnightprioriem/mkdocs-autolinks-plugin/issues/6)
- [x] Test with images
- [x] Consider potential traps of enabling this and switching all links to new format
- [x] Rename duplicated files
- [x] Remove test commit?


DominikaK:
Note to self: looks like it breaks our automatic link checker, question is, will we need it? Can we resolve it by running mkdocs in [strict mode ](https://www.mkdocs.org/user-guide/configuration/#strict)?

MagdalenaZuba (approved):

PR #1625, opened 2022-05-20
Redirect plugin

DominikaK:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | 
| Versions      | all

Enable using [mkdoc-redirects](https://github.com/mkdocs/mkdocs-redirects) plugin.

This would allow us to move redirect handling from the RtD admin to the repo, and avoid forgetting to update redirects after a PR renames or moves files.

## Potential issues

ATM doesn't handle redirects covering whole directories, or using regex, though this has already been reported: https://github.com/mkdocs/mkdocs-redirects/issues/29

Redirects work for internal headings on RtD, but lot locally, when running `mkdocs serve` - not consistently? to verify again

Redirects _do not_ seem to work with internal links (see https://github.com/ezsystems/developer-documentation/pull/1625/commits/18b3014312abf40b8b041fca5af0436b1ff8906f#diff-e34770cf50f7bea1374bcce92218746c31bb71abe6ce4ff4a2b77152cefbc196R3-R5) - `reference.md` is not converted into proper link, but `reference/` is redirected correctly (but that should not be an issue, as it is improvement anyway over setting redirects in RtD).

To do:
- [x] Remove TEST commit
- [x] Test thoroughly
- [x] Test if redirects work for internal headings (page.md#heading)
- [x] Consider extracting plugin configuration into a separate file to avoid overcrowding mkdocs.yml

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -1,3 +1,5 @@
+INHERIT: plugins.yml
+
 site_name: Developer Documentation
 repo_url: https://github.com/ezsystems/developer-documentation
 site_url: https://doc.ibexa.co/en/latest/
@@ -560,15 +562,6 @@ theme:
         primary: 'blue grey'
         accent: 'deep orange'
     logo: 'images/ibexa-dxp-logo.svg'
-
-plugins:
-    - search
-    - macros:
-        j2_block_start_string: '[[%'
-        j2_block_end_string: '%]]'
-        j2_variable_start_string: '[[='
-        j2_variable_end_string: '=]]'
-    - redirects
 
 extra:
   append_bootstrap:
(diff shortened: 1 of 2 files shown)


docs/guide/admin_panel.md
```diff
@@ -1,5 +1,9 @@
 # Administration management
 
+[TEST LINK](../references.md)
+
+[TEST LINK](../reference.md)
+
```
MagdalenaZuba, line 6:
```suggestion
```

MagdalenaZuba (approved):

docs/guide/admin_panel.md
DominikaK: That's part of the test commit, will be removed anyway before merging.

PR #1634, opened 2022-06-02
IBX-2883: Added RepeatableStep and Expressions in step fields

DominikaK:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | https://issues.ibexa.co/browse/IBX-2883
| Versions      | 3.3+

### DIFF
--- a/docs/guide/data_migration/importing_data.md
+++ b/docs/guide/data_migration/importing_data.md
@@ -96,7 +96,7 @@ This step generates Field values with fake personal names.
 
 You can use [Symfony expression syntax](https://symfony.com/doc/current/components/expression_language/syntax.html) in data migrations.
 It is especially useful in [repeatable steps](#repeatable-steps),
-where you can use it to generate varied content if migration steps.
+where you can use it to generate varied content in migration steps.
 
 The expression syntax uses the following structure: `###<IDENTIFIER> <EXPRESSION> <IDENTIFIER>###`
 


docs/guide/data_migration/importing_data.md
```diff
@@ -43,6 +43,53 @@ The following modes are available for specific objects:
 | `segment`            | &#10004; | &#10004; | &#10004; |
 | `setting`            | &#10004; | &#10004; | &#10004; |
 
+### Repeatable steps
+
+You can run the same migration step multiple times by using the special `repeatable` migration type.
```
Steveb-p, line 48:
```suggestion
You can run a set of one or more similar migration steps multiple times by using the special `repeatable` migration type.
```

docs/guide/data_migration/importing_data.md
```diff
@@ -43,6 +43,53 @@ The following modes are available for specific objects:
...
+You can vary the operations using the iteration counter.
+
+For example, to create five Folders, with names ranging from "Folder 0" to "Folder 4", you can run the following migration using the iteration counter `i`:
+
+``` yaml hl_lines="16"
+[[= include_file('code_samples/data_migration/examples/repeatable_step.yaml', 0, 16) =]]
+```
+
+To vary the content name, the migration above uses Symfony expression syntax.
+The expression is enclosed in `###` and any repeated string, in this case, `SSS`.
```
Steveb-p, line 70: I think we need to find some better words to explain this :confused: 

At the very least, we should show a "template" of sorts: `###<IDENTIFIER> <EXPRESSION> <IDENTIFIER>###`.

We can look to [PHP heredoc](https://www.php.net/manual/en/language.types.string.php#language.types.string.syntax.heredoc) documentation for inspiration. Syntax is similar there.

Additionaly, after the changes that introduced the RepeatableStep, expression syntax can be also used in each & every other migration. Repeatable step is an instance where it's almost necessary to use it, since rarely we can (or make sense) to execute the same exact migration twice, without any differences.

juskora (approved):

docs/guide/data_migration/importing_data.md
```diff
@@ -43,6 +43,64 @@ The following modes are available for specific objects:
...
+[[= include_file('code_samples/data_migration/examples/repeatable_step.yaml', 16, 19) =]]
+```
+
+This step generates Field values with fake personal names.
+
+### Expression syntax
+
+You can use [Symfony expression syntax](https://symfony.com/doc/current/components/expression_language/syntax.html) in data migrations.
+It is especially useful in [repeatable steps](#repeatable-steps),
+where you can use it to generate varied content if migration steps.
```
DominikaK, line 99:
```suggestion
where you can use it to generate varied content in migration steps.
```

Steveb-p (approved):

PR #1641, opened 2022-06-07
Added Security Headers subsection

glye:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | https://issues.ibexa.co/browse/EZP-32043
| Versions      | all

Added doc for security headers. Could be split out into a separate page if we want. Particularly if we want to add a PHP code example, but I think it would be better to add a product feature for configurable headers instead.

### DIFF
--- a/docs/guide/security_checklist.md
+++ b/docs/guide/security_checklist.md
@@ -140,7 +140,7 @@ Headers must be adapted to the site in question, and in most cases it is site ow
 The headers can be set either by the web server, or by a proxy like Varnish. 
 You can also set headers in PHP code by making a Symfony `RequestListener` for the `kernel.response` event and adding the header to the response object headers list.
 
-You will likely need to vary the security headers based on the SiteAccess in question and the site implementation details, such as frontend code and libraries used.
+You will likely need to vary the security headers based on the SiteAccess in question and site implementation details, such as frontend code and libraries used.
 
 - `Strict-Transport-Security` - ensures that all requests are sent over HTTPS, with no fallback to HTTP. 
 All production sites should use HTTPS and this header unless they have very particular needs. 


docs/guide/security_checklist.md
```diff
@@ -133,6 +133,19 @@ Once you have properly configured secure user roles and permissions, to avoid ex
...
+- `Content-Security-Policy`: Blocks cross site scripting (XSS) attacks by setting an allowlist (whitelist) of resources to be loaded for a given page. You can set separate lists for scripts, images, fonts, and more. You can use `Content-Security-Policy-Report-Only` for experimentation and testing before activating the actual policy.
+- `Referrer-Policy`: Limit what information is sent from the previous page/site when navigating to a new page/site. It has several directives for fine-tuning the referrer information.
+- `Permissions-Policy`: Limits what features the browser can use, like fullscreen, notifications, location, camera, microphone, etc. If for instance someone has succeeded in injecting their javascript into your site, this header prevents them from using those features to attack your users.
+
```
juskora, line 148:
```suggestion
- `Strict-Transport-Security`- ensures all requests are over HTTPS, with no fallback to HTTP. All production sites should use HTTPS and this header, unless they have very particular needs. During development it is less important, at least as long as the site is on an internal, protected network.
- `X-Frame-Options` - ensures the site will not be embedded in a frame, by a compliant browser. Set to `SAMEORIGIN` if you need to allow embedding by your own site, or `DENY` to block framing completely.
- `X-Content-Type-Options` - prevents the browser from second-guessing the mime-type of delivered content. Not important if users cannot upload content and you trust your editors, though the safer option is to use it. Make sure the Content-Type header is also correctly set, including for the top-level document, to avoid issues with HTML documents being downloaded when they should be rendered.
- `Content-Security-Policy` - blocks cross site scripting (XSS) attacks by setting an allowlist (whitelist) of resources to be loaded for a given page. You can set separate lists for scripts, images, fonts, and more. You can use `Content-Security-Policy-Report-Only` for experimentation and testing before activating the actual policy.
- `Referrer-Policy` - limits what information is sent from the previous page or site when navigating to a new page or site. It has several directives for fine-tuning the referrer information.
- `Permissions-Policy` - limits what features the browser can use, such as fullscreen, notifications, location, camera, microphone. If for instance someone has succeeded in injecting their JavaScript into your site, this header prevents them from using those features to attack your users.

```

juskora (changes requested):

docs/guide/security_checklist.md
glye: @juskora Right, sure. Then the first entry on line 142 should have one more space: `'- ' -> ' - '`

docs/guide/security_checklist.md
```diff
@@ -133,6 +133,19 @@ Once you have properly configured secure user roles and permissions, to avoid ex
             - { path: ^/search, roles: ROLE_USER}
     ```
 
+### Security headers
+
+There are a number of security related HTTP response headers you can use to improve your security. This is primarily the responsibility of the site owner, since the headers need to be adapted to the site in question. The headers can be set by the web server, or by a proxy like Varnish. It can also be set in PHP code, by making a Symfony `RequestListener` for the `kernel.response` event and adding the header to the response object headers list.
+
+You will likely need to vary the security headers based on the site access in question, and on the details of the site implementation, like frontend code and libraries used.
+
+- `Strict-Transport-Security`- ensures all requests are over HTTPS, with no fallback to HTTP. All production sites should use HTTPS and this header, unless they have very particular needs. During development it is less important, at least as long as the site is on an internal, protected network.
```
glye, line 142:
```suggestion
- `Strict-Transport-Security` - ensures all requests are over HTTPS, with no fallback to HTTP. All production sites should use HTTPS and this header, unless they have very particular needs. During development it is less important, at least as long as the site is on an internal, protected network.
```

juskora (approved):

glye:
Thanks for the reviews and improvements. I can merge if you like, but I haven't merged in doc before as it's really not my domain.

juskora:
@glye I'll merge it today :)

PR #1644, opened 2022-06-08
IBX-2504 - Multiple forms caution

dabrt:
| Question      | Answer |
| ------------- | --- |
| JIRA Ticket   | IBX-2504 |
| Versions      | all |

Add a caution that having multiple  instances of the same form on the page results in submitting data from  all forms at the same time


DominikaK:
We also need a mention of this outside tutorials, somewhere here: https://doc.ibexa.co/en/master/guide/form_builder/forms/

PR #1652, opened 2022-06-15
Fix errors in custom tag page

DominikaK:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | n/a
| Versions      | 3.3+ (verify for 3.3)

Fixes non-functional code samples, wrong screenshot and some leftovers from earlier versions.


MagdalenaZuba (approved):

MagdalenaZuba:
Can you take another look at updated screenshot of custom tag link, previously it was just a bracket but in a new one we can see the background with red underlining that, in my opinion, look unprofessional. I would either trim it or removed the underlining if you want to keep the background.

PR #1615, opened 2022-05-11
IBX-1430: Fix REST API reference

adriendupuis:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | [IBX-1430](https://issues.ibexa.co/browse/IBX-1430)
| Versions      | 4.1, 4.0

Fix several indentation errors that hide some parts or create non-existing routes.

- `/calendar/calendar/event` → `/calendar/event`
- `/commerce/basket/{basketId}/{basketId}/line` → `/commerce/basket/{basketId}/line`
- `/content/assets/assets/images` → `/content/assets/images`
- `/content/locations/{path}/{locationId}` → `/content/locations/{path}` (the two forms were existing and are now merged; the first one is not necessary as the location ID is always part of the location path.)
- `/content/locations/{path}/children` and `/content/locations/{path}/urlaliases` were absent from HTML output.
- `/content/objects`'s POST request example was badly indented.
- `/content/objectstategroups/objectstates` → `/content/objectstategroups/{objectStateGroupId}/objectstates`
- `/content/types/{contentTypeId}/{fieldDefinitionId}` → `/content/types/{contentTypeId}/fielddefinitions/{fieldDefinitionId}`
- remove non-existing `/content/views/{identifier}` (`/views/{viewId}` equivalent is not implemented)
- `/product/catalog/attributes/{id}`'s request body example was absent from HTML output.
- `/product/catalog/attributes/translation/{languageCode}` → `/product/catalog/attributes/translation/{id}/{languageCode}`
- `/product/catalog/attribute_groups`'s 403 descriptions were incomplete or erroneous.
- `/user/users`:
  - Fix query parameters descriptions when using `HEAD` method.
  - Move `?roleId=/user/roles/1` example to /user/groups because, on a clean install, the anonymous role `1` is not assigned to the Anonymous User but to its Anonymous User Group, and, [`/user/users?roleId` only retrieves users with the given role directly assigned to them](https://github.com/ibexa/rest/blob/main/src/lib/Server/Controller/User.php#L527).

PR #1655, opened 2022-06-15
IBX-1430: Fix REST API reference for 4.x

adriendupuis:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | [IBX-1430](https://issues.ibexa.co/browse/IBX-1430)
| Versions      | 4.1, 4.0

Last fixes forgotten in #1615
- `/user/users`: fix `HEAD` method examples
- `/user/groups`: fix `roleId` example


docs/api/rest_api_reference/input/ez-user-groups.raml
```diff
@@ -3,7 +3,7 @@ get:
     description: Loads User Groups for either an an ID or a remote ID or a Role.
     queryParameters:
         roleId:
-            description: Lists User Groups assigned to the given Role (e.g. GET /user/users?roleId=/user/roles/1).
+            description: Lists User Groups assigned to the given Role (e.g. GET /user/groups?roleId=/user/roles/1).
```
adriendupuis, line 6: /user/users and /user/groups don't handle the roleId parameter the same way.
- /user/users parses it as an href: https://github.com/ibexa/rest/blob/main/src/lib/Server/Controller/User.php#L465
- /user/groups uses it directly as an ID: https://github.com/ibexa/rest/blob/main/src/lib/Server/Controller/User.php#L572

```suggestion
            description: Lists User Groups assigned to the given Role (e.g. GET /user/groups?roleId=1).
```

PR #1660, opened 2022-06-27
improve-file_type_blacklist-overwrite

reithor:
We usually do not want to change the default value - just overwrite the value for Site Accesses when needed.

### DIFF
--- a/docs/content_management/images/images.md
+++ b/docs/content_management/images/images.md
@@ -232,7 +232,7 @@ First, enable adding SVG files to content by removing them from the blacklist
 of allowed MIME types.
 
 To do it, overwrite `ibexa.site_access.config.default.io.file_storage.file_type_blacklist` defined in `Core/Resources/config/default_settings.yml` so that `svg` is removed from the blacklist.
-You can do it per SiteAccess or SiteAccess group by using [SiteAccess-aware configuration](multisite/siteaccess_aware_configuration.md).
+You can do it per SiteAccess or SiteAccess group by using [SiteAccess-aware configuration](siteaccess_aware_configuration.md).
 
 Then, add a download route to the `config/routes.yaml` file:
 


docs/guide/images.md
```diff
@@ -227,8 +227,7 @@ a custom extension that lets you display and download files in your templates.
 First, enable adding SVG files to content by removing them from the blacklist 
 of allowed MIME types.
 
-To do it, comment out the relevant line under `ibexa.site_access.config.default.io.file_storage.file_type_blacklist`
-in `Core/Resources/config/default_settings.yml`.
+To do it, overwrite `ibexa.site_access.config.default.io.file_storage.file_type_blacklist` defined in `Core/Resources/config/default_settings.yml` so that `svg` is removed from the Blacklist. This can be done per Site Access using [SiteAccess-aware configuration](multisite/siteaccess_aware_configuration.md).
```
DominikaK, line 230:
```suggestion
To do it, overwrite `ibexa.site_access.config.default.io.file_storage.file_type_blacklist` defined in `Core/Resources/config/default_settings.yml` so that `svg` is removed from the blacklist.
You can do it per Site Access by using [SiteAccess-aware configuration](multisite/siteaccess_aware_configuration.md).
```

DominikaK (changes requested):

DominikaK (approved):

reithor:
Hi @DominikaK :
the change is also valid for https://doc.ibexa.co/en/2.5/guide/images/#handling-svg-images (slightly different - but file_type_blacklist also was introduced for 2.5)
Should I create a PR for this ?

DominikaK:
@reithor yes, if the change would be different for 2.5, please create a separate PR.

adriendupuis (approved):

DominikaK:
@reithor actually, a version for 2.5 will not be necessary, because it is part end of maintenance.
However, can you please rebase this PR so we can merge?

reithor:
@DominikaK PR is rebased.

`[SiteAccess-aware configuration](multisite/siteaccess_aware_configuration.md)`

I am not sure about internal links - shouldn't it be `[SiteAccess-aware configuration](siteaccess_aware_configuration.md)`  instead ?

docs/content_management/images/images.md
```diff
@@ -231,8 +231,8 @@ a custom extension that lets you display and download files in your templates.
 First, enable adding SVG files to content by removing them from the blacklist 
 of allowed MIME types.
 
-To do it, comment out the relevant line under `ibexa.site_access.config.default.io.file_storage.file_type_blacklist`
-in `Core/Resources/config/default_settings.yml`.
+To do it, overwrite `ibexa.site_access.config.default.io.file_storage.file_type_blacklist` defined in `Core/Resources/config/default_settings.yml` so that `svg` is removed from the blacklist.
+You can do it per SiteAccess or SiteAccess group by using [SiteAccess-aware configuration](multisite/siteaccess_aware_configuration.md).
```
DominikaK, line 235:
```suggestion
You can do it per SiteAccess or SiteAccess group by using [SiteAccess-aware configuration](siteaccess_aware_configuration.md).
```
@reithor we are using a plugin that allows linking by file name without specifying complete relative path. The link current may be working locally for you if you do not have that plugin installed, but that is why the check is failing in the PR.

PR #1688, opened 2022-07-17
drag and drop - Provide function definition

arfaram:
This will keep coherence in the doc like the example in:  https://doc.ibexa.co/en/latest/extending/creating_custom_dropdowns/#initialize . Without the function definition, it will result in `Cannot read properties of undefined (reading 'Draggable')`

| Question      | Answer
| ------------- | ---
| Versions      | v4.x

### DIFF
--- a/docs/extending/drag_and_drop.md
+++ b/docs/extending/drag_and_drop.md
@@ -22,14 +22,14 @@ First, prepare the HTML code structure and place it in a Twig template. See the 
 To initialize a drag and drop interface, add a JavaScript Code that comes with the template following the convention:
 
 ```javascript
-(function (global, document) {
-    const draggable = new global.ibexa.core.Draggable({
-        itemsContainer: document.querySelector('.items-container-drag'),
+(function (global, doc, ibexa) {
+    const draggable = new ibexa.core.Draggable({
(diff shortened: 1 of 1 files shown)


docs/extending/drag_and_drop.md
```diff
@@ -22,12 +22,14 @@ First, prepare the HTML code structure and place it in a Twig template. See the
...
-    const draggable = new ibexa.core.Draggable({
-        itemsContainer: doc.querySelector('.items-container-drag'),
+(function (global, document) {
+    const draggable = new global.ibexa.core.Draggable({
+        itemsContainer: document.querySelector('.items-container-drag'),
         selectorItem: '.item-drag',
         selectorPlaceholder: '.item-placeholder-drag',
     });
     draggable.init();
+})(window, window.document);
```
tischsoic, line 32: To follow convention from our source-code
```suggestion
(function (global, doc, ibexa) {
    const draggable = new ibexa.core.Draggable({
        itemsContainer: doc.querySelector('.items-container-drag'),
        selectorItem: '.item-drag',
        selectorPlaceholder: '.item-placeholder-drag',
    });
    draggable.init();
})(window, window.document, window.ibexa);
```

PR #1689, opened 2022-07-17
Fix not rendered icon and add example to override specific contenttype

arfaram:
* Add thumbnail using dummyimage.com otherwise the provided jpg example will not display any thumbnail 
* Add example for specific contenttype strategy override

| Question      | Answer
| ------------- | ---
| Versions      | v4.x

### DIFF
--- a/docs/extending/extending_thumbnails.md
+++ b/docs/extending/extending_thumbnails.md
@@ -87,9 +87,9 @@ Priority `100` will allow this strategy to be used first on a clean installation
 
 At this point you can go to the Back Office and check the results.
 
-!!! note "thumbnail mechanism "
+!!! note "Thumbnail mechanism "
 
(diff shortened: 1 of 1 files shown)


docs/extending/extending_thumbnails.md
```diff
@@ -87,6 +87,11 @@ Priority `100` will allow this strategy to be used first on a clean installation
 
 At this point you can go to the Back Office and check the results.
 
+!!! note "thumbnail mechanism "
+
+    Above strategy will override all genertaed thumbnails. You might specify a specific contenttype. See example [here](https://github.com/ibexa/user/blob/main/src/lib/Strategy/DefaultThumbnailStrategy.php)
+
```
juskora, line 93:
```suggestion
!!! note "Thumbnail mechanism "

    This strategy overrides all generated thumbnails. You can specify a specific Content Type. See the example [here](https://github.com/ibexa/user/blob/main/src/lib/Strategy/DefaultThumbnailStrategy.php)

```

juskora (changes requested):

juskora (approved):

PR #1704, opened 2022-07-29
IBX-3465: Added Nginx vhost template

glye:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | https://issues.ibexa.co/browse/IBX-3465
| Versions      | All supported

Like https://github.com/ezsystems/developer-documentation/pull/1466 but for Nginx.
Sample vhost copied from https://github.com/ibexa/docker/blob/main/templates/nginx/vhost.template and Ibexified.

NB: It refers to files in ez_params.d, meaning in https://github.com/ibexa/docker/tree/main/templates/nginx/ez_params.d
Should we copy them over, too?

Alternatively, should we just link directly to the vhost file in the docker repo and not copy anything over?

### DIFF
--- a/docs/getting_started/install_ibexa_dxp.md
+++ b/docs/getting_started/install_ibexa_dxp.md
@@ -347,50 +347,47 @@ for information on how to do it on different systems.
 
 Prepare a [virtual host configuration](https://en.wikipedia.org/wiki/Virtual_hosting) for your site.
 
-#### Apache
+=== "Apache"
 
-You can copy [the example vhost file](https://raw.githubusercontent.com/ibexa/post-install/main/resources/templates/apache2/vhost.template)
-to `/etc/apache2/sites-available` as a `.conf` file and modify it to fit your project.
+    You can copy [the example vhost file](https://raw.githubusercontent.com/ibexa/post-install/main/resources/templates/apache2/vhost.template)
+    to `/etc/apache2/sites-available` as a `.conf` file and modify it to fit your project.
 
-Specify `/<your installation directory>/public` as the `DocumentRoot` and `Directory`, or ensure `BASEDIR` is set in the environment.
-Uncomment the line that starts with `#if [APP_ENV]` and set the value to `prod` or `dev`, depending on the environment that you are configuring,
-or ensure `APP_ENV` is set in the environment.
+    Specify `/<your installation directory>/public` as the `DocumentRoot` and `Directory`, or ensure `BASEDIR` is set in the environment.
+    Uncomment the line that starts with `#if [APP_ENV]` and set the value to `prod` or `dev`, depending on the environment that you are configuring,
+    or ensure `APP_ENV` is set in the environment.
 
-```
-SetEnvIf Request_URI ".*" APP_ENV=prod
-```
+    ```
+    SetEnvIf Request_URI ".*" APP_ENV=prod
+    ```
 
-When the virtual host file is ready, enable the virtual host and disable the default:
+    When the virtual host file is ready, enable the virtual host and disable the default:
 
-``` bash
-a2ensite ibexa
-a2dissite 000-default.conf
-```
+    ``` bash
+    a2ensite ibexa
+    a2dissite 000-default.conf
+    ```
 
-Finally, restart the Apache server.
-The command may vary depending on your Linux distribution.
-For example, on Ubuntu use:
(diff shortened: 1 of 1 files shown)


docs/getting_started/install_ez_platform.md
```diff
@@ -317,6 +317,10 @@ Prepare a [virtual host configuration](https://httpd.apache.org/docs/2.4/vhosts/
 You can copy [the example vhost file](https://github.com/ezsystems/developer-documentation/tree/master/code_samples/install/vhost_template/vhost.template)
 to `/etc/apache2/sites-available` as a `.conf` file and modify it to fit your project.
 
+!!! note "Nginx"
+
+    If you are using Nginx instead of Apache, please use  [this example vhost file](https://github.com/ezsystems/developer-documentation/tree/master/code_samples/install/vhost_template/nginx-vhost.template) and modify it to fit your project. Also modify the instructions below accordingly.
```
DominikaK, line 322:
```suggestion
    If you are using Nginx instead of Apache, use [this example vhost file](https://github.com/ezsystems/developer-documentation/tree/master/code_samples/install/vhost_template/nginx-vhost.template) and modify it to fit your project. Also modify the instructions below accordingly.
```

DominikaK (commented):
The rationale for copying the files here was that the docker repo is unsupported, so I guess copying both files makes most sense.

glye:
In that case, should we copy over the 6 extra params files, as well?

DominikaK:
Copying over is imperfect, ideally we should have a separate repo with this (I guess there could be other "resources" like this for the whole system), and keeping it in doc risks that the files will not be updated.

glye:
What say ye, @mnocon and @micszo? It does sound nice to me to maintain default vhosts and include them in testing, so they will be updated.

micszo:
What about ibexa/post-install? Why not there?

glye:
So adding "apache" and "nginx" dirs with vhost examples under "resources" in ibexa/post-install? Sounds good to me. Ref. https://github.com/ibexa/post-install/tree/main/resources

What do the package maintainers there say about that? @Nattfarinn @webhdx @adamwojs

glye:
Post-release ping :)

> What do the package maintainers there say about that? @Nattfarinn @webhdx @adamwojs

webhdx:
It makes sense to move them to ibexa/post-install. We should also make a command for generating vhosts. I remember we had a bash script back in a day: https://github.com/ezsystems/ezplatform/blob/master/bin/vhost.sh

glye:
Thanks. I copied the templates in https://github.com/ibexa/post-install/pull/50
If that passes, I'll upgrade this PR to make use of them.

glye:
It's necromancy time! 🦹‍♂️💀
@DominikaK, since https://github.com/ibexa/post-install/pull/50 is merged, post-install now contains both apache and nginx vhosts in https://github.com/ibexa/post-install/tree/main/resources/templates and they are supposed to be maintained there.

So it makes sense to
1. remove my nginx-vhost.template from this PR
2. maybe also remove the existing apache vhost.template (or not, optional)
3. update my changes to install_ibexa_dxp.md such that both the apache and nginx sections refer to the files in post-install

Agree?

DominikaK:
@glye: Yes, Almost yes and Yes :)

I'm all for removing the template file from this repo, but it's also referred to in https://github.com/ezsystems/developer-documentation/blob/master/docs/infrastructure_and_maintenance/environments.md and https://github.com/ezsystems/developer-documentation/blob/master/docs/infrastructure_and_maintenance/security/security_checklist.md. Can you update links in those two as well?

glye:
Bunch of changes:

- Split install doc into Apache and nginx sections.
- Standardised on the "nginx" all lower case spelling which is used above in the doc, and seems to be official.
- Removed both Apache and nginx samples from doc, and refer both to those in post-install.
- Clarify instructions, you don't have to hardcode doc root and app env.
- Also pointed Environments and Security Checklist pages to post-install vhost.
- Replaced ref to line number with a text string, so it doesn't break on changes.

DominikaK (approved):

DominikaK:
I've added one commit moving the two sections, Apache and nginx, into tabs for better viewing.

glye:
@DominikaK Right, that sounds good. 👍

PR #1765, opened 2022-09-23
Fix CACHE_POOL injection in 3.3

reithor:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | N/A
| Versions      | 3.3
| Edition       | All

config loading was moved from `config/env/` to `ezplatform-core` with v3.3. Could be also possible to just write :
"automatically configured in `ezsystems/ezplatform-core` bundle" instead of mentioning the exact file name.

### DIFF
--- a/docs/guide/persistence_cache.md
+++ b/docs/guide/persistence_cache.md
@@ -160,7 +160,7 @@ Out of the box in `config/packages/cache_pool/cache.redis.yaml` you'll find a de
 
 !!! note "Ibexa Cloud"
 
-    For Ibexa Cloud/Platform.sh: This is automatically configured in `vendor/platformsh/symfonyflex-bridge/platformsh-flex-env.php` if you have enabled Redis as `rediscache` Platform.sh service.
+    For Ibexa Cloud/Platform.sh: This is automatically configured in `vendor/ezsystems/ezplatform-core/src/EzPlatformCoreBundle/bundle/DependencyInjection/EzPlatformCoreExtension.php` if you have enabled Redis as `rediscache` Platform.sh service.
 
 For anything else, you can enable it with environment variables.
 For instance, if you set the following environment variables `export CACHE_POOL="cache.redis" CACHE_DSN="secret@example.com:1234/13"`, it will result in config like this:


reithor:
change for 4.2 is similar - just using `vendor/ibexa/core/src/bundle/Core/DependencyInjection/IbexaCoreExtension.php`

DominikaK (approved):

adriendupuis:
There is a suspect inconsistency between Redis and Memcached. They both have the same note but do not target the same file:
- ["This is automatically configured in `vendor/platformsh/symfonyflex-bridge/platformsh-flex-env.php` if you have enabled Redis as `rediscache` Platform.sh service."](https://github.com/ezsystems/developer-documentation/blob/98105959db6e8904dea6c0088d7a34af11e5df30/docs/guide/persistence_cache.md?plain=1#L163)
- ["This is automatically configured in `vendor/ezsystems/ezplatform-core/src/EzPlatformCoreBundle/bundle/DependencyInjection/EzPlatformCoreExtension.php` if you have enabled Memcached as `cache` Platform.sh service."](https://github.com/ezsystems/developer-documentation/blob/98105959db6e8904dea6c0088d7a34af11e5df30/docs/guide/persistence_cache.md?plain=1#L241)
At the time I write this, this also the case on more recent branches like https://doc.ibexa.co/en/master/infrastructure_and_maintenance/cache/persistence_cache/#adjusting-configuration VS https://doc.ibexa.co/en/master/infrastructure_and_maintenance/cache/persistence_cache/#adjusting-configuration_1

adriendupuis:
Fixed for Redis too, both using https://github.com/ezsystems/ezplatform-core/blob/v2.3.30/src/EzPlatformCoreBundle/bundle/DependencyInjection/EzPlatformCoreExtension.php#L73 and https://github.com/ezsystems/ezplatform-core/blob/v2.3.30/src/EzPlatformCoreBundle/bundle/DependencyInjection/EzPlatformCoreExtension.php#L125

adriendupuis (approved):

PR #1781, opened 2022-10-10
IBX-3856 - Add note on checkbox behavior

dabrt:
| Question      | Answer | 
| ------------- | --- |
| JIRA Ticket   | IBX-3856 |
| Versions      | 3.3 and up |

CheckBox Field Type is not going to be considered as empty in any case - it can be either true or false.

### DIFF
--- a/docs/content_management/field_types/field_type_reference/checkboxfield.md
+++ b/docs/content_management/field_types/field_type_reference/checkboxfield.md
@@ -46,4 +46,4 @@ $checkboxValue = new Checkbox\Value( true );
 
 ##### String representation
 
-As this Field Type is not a string but a boolean, it will return "1" (true) or "0" (false) in cases where it is cast to string, and it is never considered as empty.
+As this Field Type is not a string but a boolean, it will return "1" (true) or "0" (false) in cases where it is cast to string, and it is never considered empty.


alongosz (approved):

docs/content_management/field_types/field_type_reference/checkboxfield.md
```diff
@@ -46,4 +46,4 @@ $checkboxValue = new Checkbox\Value( true );
 
 ##### String representation
 
-As this Field Type is not a string but a boolean, it will return "1" (true) or "0" (false) in cases where it is cast to string.
+As this Field Type is not a string but a boolean, it will return "1" (true) or "0" (false) in cases where it is cast to string, and it is never considered as empty.
```
konradoboza, line 49: Shouldn't it be
```suggestion
As this Field Type is not a string but a boolean, it will return "1" (true) or "0" (false) in cases where it is cast to string, and it is never considered empty.
```
?

konradoboza (approved):

mnocon (approved):
Thanks!

juskora (approved):

ViniTou (approved):

PR #1801, opened 2022-11-08
Limit DDL rights

glye:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | N/A
| Versions      | all

Proposal by @alongosz to further protect installations against future vulnerabilities.

#### Checklist

- [x] Text renders correctly
- [x] ~Code samples are working~
- [x] ~PHP code samples have been fixed with PHP CS fixer~
- [x] ~Added link to this PR in relevant JIRA ticket or code PR~

### DIFF
--- a/docs/infrastructure_and_maintenance/security/security_checklist.md
+++ b/docs/infrastructure_and_maintenance/security/security_checklist.md
@@ -121,7 +121,7 @@ Once you have properly configured secure user roles and permissions, to avoid ex
 
 - Avoid exposing servers on the open internet when not strictly required.
 - Ensure any servers, services, ports and virtual hosts that were opened for testing purposes are locked down before going live.
-- Secure the database with a good password, keys, firewall, etc.
+- Secure the database with a good password, keys, firewall, etc. Ensure that the database user used by the web app only has access to do the operations needed by Ibexa DXP. The Data Definition Language (DDL) commands (create, alter, drop, truncate, comment) are not needed for running Ibexa DXP, only for installing and upgrading it. If the web app user does not have these rights, then that reduces the damage that can be done if there is a security breach.
 - Consider whether certain interfaces must be left available on the open internet. Roles protect your content on all interfaces, but you may prefer to reduce your attack surface. For example:
     - The `/search` and `/graphql` endpoints
     - The REST API endpoints


alongosz (approved):

docs/infrastructure_and_maintenance/security/security_checklist.md
```diff
@@ -121,7 +121,7 @@ Once you have properly configured secure user roles and permissions, to avoid ex
 
 - Avoid exposing servers on the open internet when not strictly required.
 - Ensure any servers, services, ports and virtual hosts that were opened for testing purposes are locked down before going live.
-- Secure the database with a good password, keys, firewall, etc.
+- Secure the database with a good password, keys, firewall, etc. Ensure that the database user used by the web app only has access to do the operations needed by Ibexa DXP. The Data Definition Language (DDL) commands (create, alter, drop, truncate, comment) are not needed for running Ibexa DXP, only for installing and upgrading it. If the web app user does not have these rights, then that reduces the damage that can be done if there is a security breach.
```
Steveb-p, line 124: Schema definition of a database is not a vulnerable part of an application. Data is.

From my standpoint, once database access is acquired by a malicious actor, a breach has already occured. Removing the ability to change data definition doesn't prevent data from being leaked - at most it will cause data loss, but that can also be achieved by performing data deletion in a similar manner (with cascading options, if necessary).

At the same time, preventing application from performing DDL operations limits our upgrade capabilities, and some functions (like database-based lock mechanisms and doctrine migrations) automatically attempt to create their tables if they do not yet exist.

:-1: from me. The added complexity isn't worth the cost.

glye: Good points. Agree of course that data leaks (read operations) is the most commonly discussed db vulnerability, because of the importance of data. Also that deletes and also edits are possible without ddl. But claiming that the schema isn't a vulnerable part - that sounds to me like a dangerous assumption. Apps assume the schema is correct. Who knows in what interesting ways an app can be made to break with skillful schema manipulation? QA doesn't test for that except maybe in upgrade situations, and I guess devs don't think about it much either. For instance, what could be achieved by changing the data type of a column to a less restrictive type? Does our storage layer have valid checks for everything? Maybe it's not a real reason to worry, but we don't know for sure.

Upgrade capabilities etc: Interesting. But it seems to me that such code would in any case need to check whether it has the right to do what it wants to do with the schema, because some scrupulous admins may have already made such permission limitations anyway. Our code shouldn't crash / throw exceptions if that happens, as long as our doc doesn't explicitly say not to make such strict limitations.

In fact it raises the point of whether our upgrade scripts should verify whether both the web user and the console user has the permissions they need to have, so that they can fail in a controlled way on lack of permissions, rather than crash later when attempting to do illegal operations, leaving half-upgraded installations. Sure, it would make the upgrade process rather more complex.

I'm not decided, would like to hear @alongosz's take.

Steveb-p: > Upgrade capabilities etc: Interesting. But it seems to me that such code would in any case need to check whether it has the right to do what it wants to do with the schema, because some scrupulous admins may have already made such permission limitations anyway. Our code shouldn't crash / throw exceptions if that happens, as long as our doc doesn't explicitly say not to make such strict limitations.

Our code does not perform permission checks on database (why would it, it expects you to execute in the right user context). Also there is no common way of performing these on our supported databases. And finally - we do expect code to throw an exception in these cases, because that's exactly the state application is in: we just failed to update schema, even though we were expected to.

What can be done to prevent partial upgrades is to wrap the whole thing in a transaction, which helps **if the database supports DDL transactions** (Postgres does).

alongosz: Yes, of course you can still `DELETE` in our case. There are however systems which don't produce `DELETE` statements at any point, they just mark something as deleted and therefore that permission is not needed. I worked on a system which had permissions to just execute stored procedures and nothing else. DB was a black box which internally performed some operations.

This is not a common thing in PHP, nor Symfony, but probably we should be able to use different credentials for the installation and upgrade process and different ones for app. It's now possible, but inconvenient.

Even tough this is not a silver bullet, it's still worth to mention as a good practice. It doesn't create any extra cost for us.

glye: > What can be done to prevent partial upgrades is to wrap the whole thing in a transaction, which helps if the database supports DDL transactions (Postgres does).

And/or do permission checks beforehand. It would be preferable for performance and stability to fail early when we can during upgrades, even where DDL transactions are supported. And even if _complete_ permission checks are not possible, it's still good to fail early where we can, from the POV of the admin doing the upgrade. But whether it's worth the effort _for us_, especially if it must be done differently for each dbms as you say, that's another question. Perfection vs. hard priorities.

> preventing application from performing DDL operations limits our upgrade capabilities

All it takes is for the upgrade doc to mention that if you've blocked the app user from ddl per this recommendation, then you need to unblock x and y actions because of z change in the product. Granted, someone has to remember that, and QA may even have to test it. So there is a cost in time/complexity. I see the possible "friction" that making such a recommendation can cause, while also seeing the benefit in reducing unknown unknowns in our attack surface. It's not a hill I will die on.

(I can imagine that certain big and security minded enterprise customers would drool at the level of DB security Andrew describes above 😄 )

DominikaK: Gentlemen, do we have a decision of what should be included in the doc for this?

glye: We do not. So, should we bring in more people? Involve PM? Run a poll among engineers? Mud wrestling?
I'm open for suggestions, and not married to this PR :)

bdunogier: Mud wrestling sounds good 🐷 

I’d tend to lean towards @glye’s suggestion. It isn’t a _huge_ protection, but recommending that unneeded permissions aren’t granted is still good. As for upgrades, they’re never executed in a web context, and we could quite easily (I think) come up with an administrative access to the database for such operations.

Maybe there’s a risk with doctrine migrations that would manipulate the schema for custom ORM though...

Steveb-p: I'd agree with an additional access, if admins can handle the complexity. 

I'm slightly concerned about the additional complexity that might bring issues into our product. Specifically, dev-users would need either a copy of an app with changed connection, or two connections added to the same app (which might cause issues with our connection copy). Or additional roles would need to be granted during the upgrade/migration, and subsequently dropped.

If you feel that this is worthwile, then by all means go for it. I'd just be prepared to answer questions how our partners are actually expected to do it properly.

alongosz: > I'd agree with an additional access, if admins can handle the complexity.
> 
> I'm slightly concerned about the additional complexity that might bring issues into our product. Specifically, dev-users would need either a copy of an app with changed connection, or two connections added to the same app (which might cause issues with our connection copy, the `ibexa.persistent_connection` service). Or additional roles would need to be granted during the upgrade/migration, and subsequently dropped.
> 
> If you feel that this is worthwile, then by all means go for it. I'd just be prepared to answer questions how our partners are actually expected to do it properly.

@Steveb-p this is production recommendation, dev environment should not be concerned with this.

DominikaK (approved):

docs/infrastructure_and_maintenance/security/security_checklist.md
Steveb-p: I haven't said anything about dev/prod distinction though...? :thinking:

PR #1826, opened 2022-12-07
IBX-3139: Requirements update

MagdalenaZuba:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | https://issues.ibexa.co/browse/IBX-3139
| Versions      | 4.3

#### Checklist

- [ ] Text renders correctly

### DIFF
--- a/docs/getting_started/requirements.md
+++ b/docs/getting_started/requirements.md
@@ -39,7 +39,7 @@ event MPM is recommended, if you need to use prefork you also need the `mod_php`
 
 - 8.1
 - 8.0
-- 7.4 (PHP 7.4 has reached its End of Life. Unless you have extended support from vendors like Debian or Zend, you should use PHP 8.)
+- 7.4 (PHP 7.4 has reached its End of Life. Unless you have extended support from vendors like Debian or Zend, you should use PHP 8.1)
 
 ### PHP packages
 


docs/getting_started/requirements.md
```diff
@@ -94,8 +94,8 @@ Optionally if you intend to edit [PNG, SVG, GIF or WEBP files in the Image Edito
 
 ## Asset manager
 
-- `Node.js` 14
-- `yarn` 1.15.2 or higher
+- `Node.js` 14+
+- `yarn` 1.15.2+
```
alongosz, line 98: @ezsystems/javascript-dev-team any objections here?

alongosz (approved):

docs/getting_started/requirements.md
```diff
@@ -10,26 +10,26 @@ The following server requirements cover both running the software on-premise and
...
 
 - Debian 10.x "buster" or Debian 11.x "bullseye"
-- Ubuntu 20.04 "Focal Fossa"
+- Ubuntu 20.04 "Focal Fossa" or higher
 - RHEL / CentOS 8.1+
 
 ## Web server
 
-- Nginx 1.18 or higher
+- Nginx 1.18+
```
glye, line 26: Should we clarify in cases like this that we mean "any 1.x version higher or equal to 1.18, but not 2.x"? The same goes for "MariaDB 10.3+" and others. I mean, even if Nginx 2 and MariaDB 11 don't exist yet, suddenly one day they are released, and some customers will demand they work from day one, given text like this. Or do you guys feel it's clear?

MagdalenaZuba: I feel it's clear, but I think it is worth mentioning somewhere.

MagdalenaZuba: https://github.com/ezsystems/developer-documentation/pull/1826/files#diff-113e92f1aa1bfe8d987f618755d55f4e62bb5158db53286243b9b58746533b85R16

docs/getting_started/requirements.md
```diff
@@ -10,26 +10,28 @@ The following server requirements cover both running the software on-premise and
...
-Using the latest listed version of each product or component is always recommended.
+!!! note "Recommended versions"
+
+    Using the latest listed version of each product or component is always recommended. Review all the recommended versions carefully. If you see a "+" next to the product version it means that we recommend this version or higher within the same major release for example "Nginx 1.18+" means any 1.x version higher or equal to 1.18, but not 2.x.
 
 ## Operating system
 
 - Debian 10.x "buster" or Debian 11.x "bullseye"
-- Ubuntu 20.04 "Focal Fossa"
+- Ubuntu 20.04 "Focal Fossa" or higher
```
micszo, line 23: In case of operating systems "or higher" is a loaded statement. Because an OS includes in itself many versions of various components.
In the particular case of Ubuntu, 20.04 "Focal Fossa" is the previous LTS (still in standard support, as is 18.04), 22.04 "Jammy Jellyfish" is the current LTS. But it hasn't been tested with our product yet.
In short, IMO we shouldn't add this until the lates LTS was tested (in a separate sprint ticket).
ref. https://www.releases.ubuntu.com/

docs/getting_started/requirements.md
```diff
@@ -10,26 +10,28 @@ The following server requirements cover both running the software on-premise and
...
     For running on [Ibexa Cloud](https://www.ibexa.co/products/ibexa-cloud), where recommended configuration and support is provided out of the box, see separate [Ibexa Cloud section](#ibexa-cloud-requirements-and-setup) for further reading on its requirements.
 
-The minimal setup requires PHP,  MySQL/MariaDB, Apache/Nginx, Node.js and `yarn`.
+The minimal setup requires PHP, MySQL/MariaDB, Apache/Nginx, Node.js and `yarn`.
 Recommendation for production setups is to use Varnish/Fastly, Redis/Memcached, NFS/EFS/S3 and Solr/Elasticsearch in a [clustered setup](clustering.md).
 
-Using the latest listed version of each product or component is always recommended.
+!!! note "Recommended versions"
+
+    Using the latest listed version of each product or component is always recommended. Review all the recommended versions carefully. If you see a "+" next to the product version it means that we recommend this version or higher within the same major release for example "Nginx 1.18+" means any 1.x version higher or equal to 1.18, but not 2.x.
```
glye, line 18:
```suggestion
    Using the latest listed version of each product or component is always recommended. Review all the recommended versions carefully. If you see a "+" next to the product version it means that we recommend this version or higher within the same major release. For example "Nginx 1.18+" means any 1.x version higher or equal to 1.18, but not 2.x.
```
Added a period, for breathing space :D

docs/getting_started/requirements.md
glye: Resolved, but see separate punctuation suggestion :)

glye (approved):

docs/getting_started/requirements.md
```diff
@@ -94,8 +96,8 @@ Optionally if you intend to edit [PNG, SVG, GIF or WEBP files in the Image Edito
 
 ## Asset manager
 
-- `Node.js` 14
-- `yarn` 1.15.2 or higher
+- `Node.js` 14+
```
micszo, line 99: I think the plus sign next to Node could be tricky. 
People could interpret it to mean 16, 18 etc.
Summon @dew326

MagdalenaZuba: @micszo We have a big note at the top of the page that says: If you see a “+” next to the product version, it means that we recommend this version or higher within the same major release. For example, “Nginx 1.18+” means any 1.x version higher or equal to 1.18, but not 2.x.

micszo: Yeah, I've seen it. 😉 But I would be calmer if someone from JS team confirmed it's clear enough for Node.

MagdalenaZuba: @dew326 even so, we cannot remove node 14 for v4 for legal reasons, as @alongosz explained to me

micszo (approved):

docs/getting_started/requirements.md
glye: It would be really great to have a dependencies page where you select your product and then it lists all the scheduled deprecations and EOLs we know about, in an easy to read overview. Something like https://www.php.net/supported-versions.php only covering more deps. Would require some development :)

For now, sprinkling notes here and there is much better than nothing. +1 to this. And while we're at it, we should also note that PHP 7.4 is EOL, _unless_ you have extended support from vendors like Debian or Zend.

glye (changes requested):
Can you please add something like this to PHP 7.4?

"NB: PHP 7.4 is EOL. Unless you have extended support from vendors like Debian or Zend, you should use PHP 8."

docs/getting_started/requirements.md
MagdalenaZuba: Let's go with additional notes here and there as we are mid v4, we will think about redesigning and better showcasing the dependencies for LTS or v5.

MagdalenaZuba: For Node.js, I would add a note that it is EOL after 2023-04-30. I don't want Requirements page turning into a calendar with dates that we need to keep track of and update.

glye (approved):

docs/getting_started/requirements.md
```diff
@@ -23,21 +25,21 @@ Using the latest listed version of each product or component is always recommend
...
+- MariaDB 10.3+
 - MySQL 8.0
 - PostgreSQL 10+
 
 ## PHP
 
 - 8.1
 - 8.0
-- 7.4
+- 7.4 (PHP 7.4 is EOL. Unless you have extended support from vendors like Debian or Zend, you should use PHP 8.)
```
DominikaK, line 42:
```suggestion
- 7.4 (PHP 7.4 has reached its End of Life. Unless you have extended support from vendors like Debian or Zend, you should use PHP 8.)
```

DominikaK (approved):

dew326 (approved):

docs/getting_started/requirements.md
```diff
@@ -23,21 +25,21 @@ Using the latest listed version of each product or component is always recommend
...
+- MariaDB 10.3+
 - MySQL 8.0
 - PostgreSQL 10+
 
 ## PHP
 
 - 8.1
 - 8.0
-- 7.4
+- 7.4 (PHP 7.4 has reached its End of Life. Unless you have extended support from vendors like Debian or Zend, you should use PHP 8.)
```
adamwojs, line 42:
```suggestion
- 7.4 (PHP 7.4 has reached its End of Life. Unless you have extended support from vendors like Debian or Zend, you should use PHP 8.1)
```

adamwojs (approved):

docs/getting_started/requirements.md
glye: 8.2 was just released, but I guess we're not supporting it just yet :) https://www.php.net/releases/8.2/en.php

adamwojs: I don't expect any issues but we should do "official" certification as usually // cc @micszo

micszo: We already have plans in the team. 🙂

PR #1837, opened 2022-12-16
Change gitignore settings

dabrt:
| Question      | Answer |
| ------------- | --- |
| JIRA Ticket   | N/A |
| Versions      | all |

Change gitignore settings to include phpstorm files in .idea folder

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,10 +2,4 @@
 **/.DS_Store
 __pycache__/*
 /site/
-.idea/**/tasks.xml
-.idea/**/workspace.xml
-.idea/**/vcs.xml
-.idea/**/*.iml
-.idea/**/modules.xml
-.idea/**/php.xml
-.idea/**/.gitignore
+**/.idea/


dabrt:
@DominikaK One rabbi says "ignore it all", the other says "modules and vcs files should be preserved".
I added all the files that popped up in my project, as you suggested.

dabrt:
I also did a quick check on repositories in our organization by using the `idea org:ezsystems org:ibexa filename:.gitignore` search phrase and it seems that in most cases out devs tend to ignore the whole `.idea` folder, without listing detailed filenames, except for maybe a couple of our repositories, which specifically rule out *.iml files.
Funnily enough, this is contrary to the instructions that come from [intelliJ](https://intellij-support.jetbrains.com/hc/en-us/articles/206544839), where they state that iml files should be uploaded to VCS.

DominikaK:
Given that our doc is not a PHP project and we only use Storm opportunistically as a text editor, I think we'd be good with ignoring the whole `idea` folder.

PR #1843, opened 2022-12-28
REST API ref: Merge /user/groups/{id}/ into /user/groups/{path}/

adriendupuis:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | 
| Versions      | 4.3 ([ibexa/rest 4.3](https://github.com/ibexa/rest/blob/v4.3.0/src/bundle/Resources/config/routing.yml#L1051)), 4.2 ([ibexa/rest 4.2](https://github.com/ibexa/rest/blob/v4.2.4/src/bundle/Resources/config/routing.yml#L1051)),<br/> 3.3 ([ezsystems/ezplatform-rest 1.3](https://github.com/ezsystems/ezplatform-rest/blob/1.3/src/bundle/Resources/config/routing.yml#L1059)), 2.5 ([ezsystems/ezpublish-kernel 7.5](https://github.com/ezsystems/ezpublish-kernel/blob/7.5/eZ/Bundle/EzPublishRestBundle/Resources/config/routing.yml#L1051))

* Merge `/user/groups/{id}/users` into `/user/groups/{path}/users`: [`GET`](https://github.com/ibexa/rest/blob/v4.3.0/src/bundle/Resources/config/routing.yml#L1067) and [`POST`](https://github.com/ibexa/rest/blob/v4.3.0/src/bundle/Resources/config/routing.yml#L1075) of the same route were separated, both respective controller actions ([`loadUsersFromGroup`](https://github.com/ibexa/rest/blob/v4.3.0/src/lib/Server/Controller/User.php#L813) and [`createUser`](https://github.com/ibexa/rest/blob/v4.3.0/src/lib/Server/Controller/User.php#L272)) treat the argument as a path, and extract an ID from it.
* Merge `/user/groups/{id}/subgroups` into `/user/groups/{path}/subgroups`: [`GET`](https://github.com/ibexa/rest/blob/v4.3.0/src/bundle/Resources/config/routing.yml#L1051) and [`POST`](https://github.com/ibexa/rest/blob/v4.3.0/src/bundle/Resources/config/routing.yml#L1059) of the same route, [`loadSubUserGroups`](https://github.com/ibexa/rest/blob/v4.3.0/src/lib/Server/Controller/User.php#L715) and [`createUserGroup `](https://github.com/ibexa/rest/blob/v4.3.0/src/lib/Server/Controller/User.php#L228) treat the argument as a path. 

### TODO
- [ ] Generate HTML just before merge (after would have been easier for cherry picking)
- [x] Rebase to hide ez_original.raml edits

#### Checklist

- [x] Text renders correctly


docs/api/rest_api/rest_api_reference/input/ez-user-groups.raml
```diff
@@ -127,6 +127,36 @@ get:
             404:
                 description: Error - the User Group does not exist.
     /users:
+        get:
+            displayName: Load Users of Group
+            description: Loads the Users of the Group with the given ID.
```
adriendupuis, line 132: The controller is extracting the ID from the path argument by splitting this path on slashes and keeping the last item ([` extractLocationIdFromPath`](https://github.com/ibexa/rest/blob/v4.3.0/src/lib/Server/Controller/User.php#L1043)). So, it works with just the ID.

The description is not wrong but doesn't cover the whole possibilities. If having only a path, a developer shouldn't extract the ID from this path before use but directly use this path. Anyway, just having access to the path seems like a rare case to me, and an ID simpler to use. 

I propose to not edit this description in this PR.

But this is a more general problem than just below `/user/groups/{path}/`;
Other routes having a {path} argument are in fact extracting the ID from it with their documentation not saying it.
From my PoV, the worst could be [`/content/locations/{path}` description](https://github.com/ezsystems/developer-documentation/blob/6df96089d7247df80266d0edca00d41073791676/docs/api/rest_api/rest_api_reference/input/ez-content-location.raml#L27) only documenting the path while [its controller will extract the ID from this path](https://github.com/ibexa/rest/blob/v4.3.0/src/lib/Server/Controller/Location.php#L392). The 3 following example are equivalent but only the complex one is proposed:
* `/content/locations/1/2/61` (full path)
* `/content/locations/2/61` (incomplete path)
* `/content/locations/61` (ID)

Steveb-p (approved):

adriendupuis:
> We shouldn't really encourage using "path" as the last argument, even if it works this way. Instead we should always declare that we expect an ID there.

@Steveb-p I'm agree. So we could merge route entries the reverse way, both under /user/groups/{id}/ instead.

PR #1850, opened 2023-01-11
Update installation instruction for PHP 7.4

DominikaK:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | n/a
| Versions      | all

Note: for 3.3 merge after 3.3.30 and update to replace the current note (switching php 8.0 with 7.4).

#### Checklist

- [x] Text renders correctly


MagdalenaZuba (approved):

webhdx (changes requested):
According to @vidarl's (investigation)[https://issues.ibexa.co/browse/IBX-4773] the same steps should be used when installing on PHP 8.0.

webhdx (approved):

konradoboza (approved):

PR #1857, opened 2023-01-13
Added JSON `/views` example

Steveb-p:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | N/A
| Versions      | 3.3, 4.x

Added JSON sample for `/views` route, similar to the XML one.

### Additional changes

Markdown library and Twig extension has been added to facilitate writing longer descriptions for routes, including paragraphs and links specifically.

#### Checklist

- [ ] Text renders correctly


docs/api/rest_api_reference/input/examples/views/POST/ViewInput.json.example
```diff
@@ -0,0 +1,16 @@
...
+        "ContentQuery": {
+            "Filter": {
+                "ContentTypeIdentifierCriterion": "image",
+                "SectionIdentifierCriterion": "media"
+            },
+            "limit": 10,
+            "offset": 0,
+            "SortClauses": {
+                "ContentName": "ascending"
+            }
```
adriendupuis, line 13: Can we have the `contentTypeFacetBuilder` like [in the XML example](https://github.com/ezsystems/developer-documentation/blob/add-views-search-sample/docs/api/rest_api_reference/input/examples/views/POST/ViewInput.xml.example#L14)?

adriendupuis, line 13: Can we have the `contentTypeFacetBuilder` like [in the XML example](https://github.com/ezsystems/developer-documentation/blob/add-views-search-sample/docs/api/rest_api_reference/input/examples/views/POST/ViewInput.xml.example#L14)?

adriendupuis (changes requested):

docs/api/rest_api_reference/input/examples/views/POST/ViewInput.json.example
MagdalenaZuba: @Steveb-p

Steveb-p: So, the thing is, Facet Builders were supposed to be removed in 4.x. While I can construct a query in JSON that creates those facet builder objects internally, I'm not sure if they will work properly.

When reading through the code related to search engine (`Ibexa\Contracts\Core\Repository\SearchService::findContent`) facet builders appear to be unused and therefore silently discarded.

Needs confirmation from @alongosz / @Nattfarinn / @adamwojs: The first two placed the deprecation notice, while Adam is knowledgeable regarding how facet builders / aggregations should behave in context of REST API, or at least could point us in the right direction.

alongosz: > So, the thing is, Facet Builders were supposed to be removed in 4.x. While I can construct a query in JSON that creates those facet builder objects internally, I'm not sure if they will work properly.
> 
> When reading through the code related to search engine (`Ibexa\Contracts\Core\Repository\SearchService::findContent`) facet builders appear to be unused and therefore silently discarded.
> 
> Needs confirmation from @alongosz / @Nattfarinn / @adamwojs: The first two placed the deprecation notice, while Adam is knowledgeable regarding how facet builders / aggregations should behave in context of REST API, or at least could point us in the right direction.

Aggregations should be used instead of Facet builders.

Steveb-p: @alongosz That is plain. I'll provide concrete questions:
1. Do facet builders still work in context of REST API?
2. Do aggregation work in context of REST API (as I haven't found direct reference to their property in `ViewInput` class yet)?
3. Do you know the structure of XML/JSON that should be used for REST API to make it work (I'm looking for tests that might cover this case and / or directly looking into code at this very moment, but maybe you have some samples on hand)?

alongosz: > @alongosz That is plain. I'll provide concrete questions:
> 
> 1. Do facet builders still work in context of REST API?
> 2. Do aggregation work in context of REST API (as I haven't found direct reference to their property in `ViewInput` class yet)?
> 3. Do you know the structure of XML/JSON that should be used for REST API to make it work (I'm looking for tests that might cover this case and / or directly looking into code at this very moment, but maybe you have some samples on hand)?

It doesn't matter whether it works or not, it shouldn't be advocated in the doc.

To fully answer these questions 1 and 3 I need to spend significant amount of time, so you'll need to wait once I have such time slot available. Question 2 is for Adam, I have no knowledge of Aggregations.

Steveb-p: @adriendupuis added samples with `Aggregations`, and added a reference to our Search Criteria reference page - to do this I also added markdown dependency on the raml builder tool (which I think would be a nice improvement over trying to squeeze pure HTML into method description, which is also an option).

A few things I've discovered when testing those, so I'll be pushing PRs into rest package, and cross-link them here.

adriendupuis (approved):

MagdalenaZuba (approved):

tools/raml2html/composer.json
```diff
@@ -20,7 +20,9 @@
         "raml-org/raml-php-parser": "dev-master",
         "twig/twig": "^2.0",
         "symfony/filesystem": "^4.2",
-        "ramsey/uuid": "^3.9"
+        "ramsey/uuid": "^3.9",
+        "twig/markdown-extra": "^3.5",
+        "league/commonmark": "^2.3"
```
adriendupuis, line 25:
```
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - Root composer.json requires league/commonmark ^2.3 -> satisfiable by league/commonmark[2.3.0, ..., 2.3.8].
    - league/commonmark[2.3.0, ..., 2.3.8] require php ^7.4 || ^8.0 -> your php version (7.2.34) does not satisfy that requirement.
```

So, this addition is not compatible with `"php": "^7.2",` above.

adriendupuis: Quick fix in c34be1469801e0ce0035448577074fbbe991a4fa for PHP 7.4 w/ few warnings at runtime.

A better fix for PHP 8.1+ will come in #1871

Steveb-p: @adriendupuis we will need to take the raml parser and everything else related to it under our organization, because that library is in practice abandoned.

PR #1860, opened 2023-01-16
IBX-4064: Search page

adriendupuis:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | [IBX-4064](https://issues.ibexa.co/browse/IBX-4064), [IBX-5007](https://issues.ibexa.co/browse/IBX-5007)
| Versions      | All

* `See all results` link at bottom of search dropdown menu
    - If there is less than 10 results, all results are in the drop down menu so the link is not shown.
    - If there is more than 10 results, 10 are shown on the drop down and the link allow to access the search result page where other result can be consulted. (examples: "TaxonomyEntryIdAggregation" 6 results, "SesExternalData" 4 results)
    - If there is 10 results, the link is shown even if useless due to the drop down not knowing the real result count  
* Page to see all those results. It's limited to 1000 results/100 pages by Algolia default settings. After few unsuccessful attempts to increase [this](https://www.algolia.com/doc/api-reference/api-parameters/paginationLimitedTo/), it was decided to not invest more time in it as nobody reaches the 100th page anyway. Page is empty when there is no searched text.

To wrap the search page into the doc layout, I use https://www.mkdocs.org/user-guide/writing-your-docs/#meta-data

Search page preview: https://ez-systems-developer-documentation--1860.com.readthedocs.build/en/1860/search_results/?sq=Content&p=5

Search dropdown previews:

It can be previewed anywhere on this [PR build](https://ez-systems-developer-documentation--1860.com.readthedocs.build/en/1860/). But, notice that the “See all results” link is always wrong has the PR ID is not seen as a valid version/branch and is replaced by `master`. You can put the version back to PR ID 1860 to be on https://ez-systems-developer-documentation--1860.com.readthedocs.build/en/1860/search_results/ with the right arguments behind.

search_Content
The "See all results" link will bring to a page like https://ez-systems-developer-documentation--1860.com.readthedocs.build/en/1860/search_results/?sq=Content&p=1

search_TaxonomyEntryIdAggr
No "See all results" link as all the results are available in the drop down suggestions.

search_customize_cart

#### Checklist

- [ ] Text renders correctly

### DIFF
--- a/docs/js/instantsearch.js
+++ b/docs/js/instantsearch.js
@@ -9,6 +9,7 @@
         version = 'master';
     }
     const hitsContainer = '#hits';
+    const statsContainer = '#stats';
     const paginationContainer = '#pagination';
     const search = instantsearch({
         indexName: 'ezplatform',
@@ -23,9 +24,11 @@
         searchFunction(helper) {
             if (helper.state.query) {
                 helper.search();
+                $(statsContainer).css('visibility', 'visible');
                 $(paginationContainer).show();
             } else {
                 $(hitsContainer).empty();
+                $(statsContainer).css('visibility', 'hidden');
                 $(paginationContainer).hide();
             }
         },
@@ -76,7 +79,7 @@
             hitsPerPage: 10,
         }),
         instantsearch.widgets.stats({
-            container: '#stats',
+            container: statsContainer,
             templates: {
                 text: `<h1>
                 Search results ({{#helpers.formatNumber}}{{nbHits}}{{/helpers.formatNumber}})


docs/js/custom.js
```diff
@@ -97,26 +97,44 @@ $(document).ready(function() {
                     }
                 }
             });
+            let link = $('.ds-dropdown-menu a.search-page-link');
+            if (!link.length) {
+                link = $('.ds-dropdown-menu').append('<a class="search-page-link" href="" style="position: absolute; bottom: 0px; z-index: 1000;">See all results</a>');
+            }
+            let href = '/en/' + branchName + '/search_results/#q=' + encodeURI($(/*this.inputSelector*/'#search_input').val()) + '&p=1';
+            link.attr('href', href).show();
+            if (hits.length < 10/*hitsPerPage*/) {
```
adriendupuis, line 106: hits.length == hitsPerPage means that there are hitsPerPage or more results; hits.length > hitsPerPage can't happen. So we can only be sure that the link must be hidden if there is less than hitsPerPage results.

docs/js/instantsearch.js
```diff
@@ -0,0 +1,170 @@
...
+                        children: [...groupChildren, newItem],
+                    };
+                });
+
+                return Object.values(outputItemsMap);
+            },
+            templates: {
+                item: (hit) => {
+                    const { groupNameKey, children } = hit;
+                    const groupItem = children[0].item;
```
adriendupuis, line 98: This is wrong, not all hits on a page are in the same group.

adriendupuis: Fixed in 91f98906b36d5f5f1976ea95c8842f7648b935de where there is room for refactoring.

docs/js/custom.js
```diff
@@ -104,20 +104,21 @@ $(document).ready(function() {
...
+    let search = docsearch({
         container: '#docsearch',
         appId: '2DNYOU6YJZ',
         apiKey: '21ce3e522455e18e7ee16cf7d66edb4b',
         indexName: 'ezplatform',
         inputSelector: '#search_input',
         transformData: function(hits) {
-            let removedPattern = '¶';
+            const hitsPerPage = 10;
+            const removedPattern = '¶';
```
adriendupuis, line 121: This is now done at indexation time

docs/js/custom.js
```diff
@@ -128,24 +129,51 @@ $(document).ready(function() {
...
+
+            link.attr('href', href).show();
+
+            if (hits.length < hitsPerPage) {
+                link.hide();
+            }
         },
         algoliaOptions: {
             facetFilters: ['lang:en', 'version:' + branchName],
             hitsPerPage: 10,
```
adriendupuis, line 150: Should use the same constant than `transformData` above

docs/js/custom.js
```diff
@@ -141,7 +141,7 @@ $(document).ready(function() {
                 window.location = $('.ds-dropdown-menu a.search-page-link').attr('href');
             }
         },
-        debug: false,
+        debug: true,
```
adriendupuis, line 144: :facepalm:

mnocon (approved):
Looks great! I've checked both on Safari and Firefox that it's rendered corretly, did you have a chance to check it with Chrome?

Two (optional) remarks, we can discuss them and maybe come back to them as follow ups:
1) It looks like the number of pages is capped at 100 - and with 10 items per page you can "only" get the first 1000 results.

Some queries return more than 1k results, meaning that some of them are not accessible, for example:
https://ez-systems-developer-documentation--1860.com.readthedocs.build/en/1860/search_results/?sq=Content&p=1

You cannot see what the result 1900 is, even when browsing to the last page.

What do you think about displaying `1000+` when there are more than 1k results?

2) Having a full search page is great for experimentation and really tailoring the query.
When playing with this I found myself using the `-` keyword a lot to remove things which I'm not looking for:

`migration step` gives different results than `migration step -update`.
https://ez-systems-developer-documentation--1860.com.readthedocs.build/en/1860/search_results/?sq=migration+step&p=1

https://ez-systems-developer-documentation--1860.com.readthedocs.build/en/1860/search_results/?sq=migration+step+-update&p=1

I thought about including an info box somewhere in the search page informing users about this possibility - it's described in the Algolia doc: https://www.algolia.com/doc/api-reference/api-parameters/advancedSyntax/#how-to-use

It's a small feature, but useful for power users (developers). Or do you think developers already know this?

docs/images/ez-icons.svg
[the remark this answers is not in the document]
adriendupuis: I'll ping @GrabowskiM on this one.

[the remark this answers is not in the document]
adriendupuis: @mnocon About result page limit, I tried to use https://www.algolia.com/doc/api-reference/api-parameters/paginationLimitedTo/ in our crawler to increase from 1000 to 2000 without success. The property seem recognized by the Editor but after reindexing I saw no change.
paginationLimitedTo
I propose to give up as readers probably won't get past the 10th page.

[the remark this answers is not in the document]
adriendupuis: > It's a small feature, but useful for power users (developers). Or do you think developers already know this?

I found no place to add this nor the right text to explain this. I feel like we should explain it somewhere as it's a powerful feature, or someone could exclude some results without knowing it while searching for a command line with options or whatever starting with a minus/hyphen. Google doesn't document it while it handles it. This is something we could keep for a second round.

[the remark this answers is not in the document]
mnocon: Sure, let's keep both of them for the second round

dabrt (approved):

PR #1642, opened 2022-06-08
IBX-1430: Complete 4.4 REST API reference

adriendupuis:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | [IBX-1430](https://issues.ibexa.co/browse/IBX-1430)
| Versions      | 4.4

Update/Complete REST API reference
- Add missing methods to existing routes
- Add missing routes
- ~~Update namespace (`xmlns`) from `ez.no` to `ibexa.co`~~ (kept for another latter PR)
- Harmonize deprecation and removal flags
- ~~Add a `test` command to compare the built reference against real routing configuration~~ (moved to #1901)
- Fix `build` command's templates to quit skipping some methods

Related and dependencies:
- Follows #1615 and #1843 where there were only fixes.
- ~~Needs ibexa/scheduler#76 to properly document scheduler API~~

Changelog:
- Add [`GET /content/types/{contentTypeId}/fieldDefinitions`](https://github.com/ibexa/rest/blob/v4.3.0/src/bundle/Resources/config/routing.yml#L570)
- Add [`GET /content/types/{contentTypeId}/draft`](https://github.com/ibexa/rest/blob/v4.3.0/src/bundle/Resources/config/routing.yml#L603)
- Add [`GET /content/types/{contentTypeId}/draft/fieldDefinitions`](https://github.com/ibexa/rest/blob/v4.3.0/src/bundle/Resources/config/routing.yml#L611)
- Add [`POST /content/objects/{contentId}/hide`](https://github.com/ibexa/rest/blob/v4.3.0/src/bundle/Resources/config/routing.yml#L228)
- Add [`POST /content/objects/{contentId}/reveal`](https://github.com/ibexa/rest/blob/v4.3.0/src/bundle/Resources/config/routing.yml#L236)
- Add [`GET /product/catalog/attribute_types`](https://github.com/ibexa/product-catalog/blob/v4.3.0/src/bundle/Resources/config/routing_rest.yaml#L130)
- Add [`GET /product/catalog/attribute_types/{identifier}`](https://github.com/ibexa/product-catalog/blob/v4.3.0/src/bundle/Resources/config/routing_rest.yaml#L137)
- Add [`GET /product/catalog/regions/{identifier}`](https://github.com/ibexa/product-catalog/blob/v4.3.0/src/bundle/Resources/config/routing_rest.yaml#L436)
[...]

PR #1871, opened 2023-01-27
IBX-4498: raml2html: Upgrade dependencies

adriendupuis:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | [IBX-4498](https://issues.ibexa.co/browse/IBX-4498)
| Versions      | N/A

Update to PHP 8.0+

I moved from slightly out-dated [adamwojs/raml-php-parser](https://github.com/adamwojs/raml-php-parser) (forked around [v4.6.1](https://github.com/raml-org/raml-php-parser/releases/tag/v4.6.1)) back to [raml-org/raml-php-parser](https://github.com/raml-org/raml-php-parser) for actual [v4.8.1](https://github.com/raml-org/raml-php-parser/releases/tag/v4.8.1) and next versions. It should be easier to keep up-to-date using `composer update` than synchronising a repository. I still use modifications from Ibexa members thanks to [cweagans/composer-patches](https://github.com/cweagans/composer-patches). If one day, those patches are rejected, we'll improvise.

This needed some fixes in the RAML files started a bit with #1642 and mainly ended in #2042.

How-to update: default PHP must 8.2, 8.1 or 8.0 to reinstall with the following commands:
```bash
php -v; #To check actual default PHP version
cd tools/raml2html;
rm -rf var vendor composer.lock;
composer install;
cd -;
```

#### Checklist

- [ ] Text renders correctly

### DIFF
--- a/tools/raml2html/README.md
+++ b/tools/raml2html/README.md
@@ -14,5 +14,5 @@ To generate static HTML from RAML definitions, use the following code:
 
 
 ```sh
-php tools/raml2html/raml2html.php build --non-standard-http-methods=COPY,MOVE,PUBLISH,SWAP -t default -o docs/api/rest_api_reference/output/ docs/api/rest_api_reference/input/ez.raml
+php tools/raml2html/raml2html.php build --non-standard-http-methods=COPY,MOVE,PUBLISH,SWAP -t default -o docs/api/rest_api/rest_api_reference/ docs/api/rest_api/rest_api_reference/input/ez.raml
 ```


tools/raml2html/patches/php_8.2.patch
```diff
@@ -0,0 +1,111 @@
...
+         return isset($this->collection[$this->position]);
+     }
+--- /src/Body.php	2023-01-27 15:49:58
++++ /src/Body.php	2023-01-27 15:51:03
+@@ -228,7 +228,7 @@
+      */
+     public function getExample()
+     {
+-        return $this->examples[0];
++        return $this->examples[0] ?? '';
```
adriendupuis, line 97: `PHP Notice:  Trying to access array offset on value of type null in tools/raml2html/vendor/raml-org/raml-php-parser/src/Body.php on line 231` appears since the usage of PHP 7.4 instead of PHP 7.2

So, the patch file name is not really accurate.

tools/raml2html/composer.json
```diff
@@ -1,37 +1,42 @@
...
             "email": "adam@wojs.pl"
-        }
-    ],
-    "repositories": [
+        },
         {
-            "type": "vcs",
-            "url": "https://github.com/adamwojs/raml-php-parser"
+            "name": "Adrien Dupuis",
+            "email": "adrien.dupuis@ibexa.co"
```
adriendupuis, line 12: TODO: Add other contributors like @Steveb-p

tools/raml2html/.gitignore
```diff
@@ -1,4 +1,3 @@
 var
 vendor
-composer.lock
```
adriendupuis, line 3: Still ignored by [root's .gitignore](https://github.com/ezsystems/developer-documentation/blob/2278ce29f91bab2640a08a37cb15443b810d3875/.gitignore#L7) 😞 
I should also remove it as a [best practice](https://getcomposer.org/doc/01-basic-usage.md#commit-your-composer-lock-file-to-version-control) but this is not the goal of this PR to change the whole practice at root level.

tools/raml2html/composer.json
```diff
@@ -1,39 +1,49 @@
...
         }
     ],
     "require": {
-        "php": "^7.4",
-        "symfony/console": "^4.2",
-        "raml-org/raml-php-parser": "dev-master",
-        "twig/twig": "^2.0",
-        "symfony/filesystem": "^4.2",
-        "ramsey/uuid": "^3.9",
+        "php": "^8.1",
```
adriendupuis, line 16:
```suggestion
        "php": "^8.1",
        "ext-json": "*",
```

docs/api/rest_api/rest_api_reference/input/ez-types.raml
```diff
@@ -1071,6 +1081,7 @@ Message:
         values:
             description: 'Translation value objects. May not contain any numbers, which might result in requiring plural forms. Use Plural for that.'
             type: array
+            items: object
```
adriendupuis, line 1084: This has been added to avoid `PHP Deprecated:  mb_substr(): Passing null to parameter #1 ($string) of type string is deprecated in tools/raml2html/vendor/raml-org/raml-php-parser/src/Type.php on line 85`

While trying to better define what those item objects are, I discovered that the type `Message` is in fact never used. If `Message` was used, removing it would produce `An exception has been thrown during the rendering of a template ("No type found for name 'Message'").` which is not the case.

I wrote a script testing if types defined in ez-types.raml are used and it found several unused ones.

I think that this is not the PR to use or to remove unused types. I stay with the minimal fix.

tools/raml2html/patches/php_8.1.patch
adriendupuis, line 1: See https://github.com/raml-org/raml-php-parser/pull/180 for a proposal of this fix on vendor side.

docs/api/rest_api/rest_api_reference/input/ez-types.raml
adriendupuis: See #2042

tools/raml2html/composer.json
adriendupuis: Let's not expose @Steveb-p ;)

adamwojs (approved):

PR #1896, opened 2023-02-15
Stabilized render of REST HTML

Steveb-p:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | N/A
| Versions      | N/A
| Edition       | N/A

This PR ensures that HTML generated for REST API documentation remains the same across multiple re-renders.

What it basically does, it replaces the `uuid` function usage with `hash` function. Since the context provided to the hashing function does not change, it ensures that each and every time the same ID string is generated.

What we will need to see is if we ever encounter the same hash twice, as that would mean we have the same context in multiple locations. To prevent collision, all hashes are stored and rendering will fail if the same hash is generated twice.

#### Checklist

- [ ] Text renders correctly

### DIFF
--- a/tools/raml2html/composer.json
+++ b/tools/raml2html/composer.json
@@ -15,7 +15,7 @@
         }
     ],
     "require": {
-        "php": "^7.2",
+        "php": "^7.4",
         "symfony/console": "^4.2",
         "raml-org/raml-php-parser": "dev-master",
         "twig/twig": "^2.0",
--- a/tools/raml2html/src/Twig/Extension/RenderExtension.php
+++ b/tools/raml2html/src/Twig/Extension/RenderExtension.php
@@ -68,8 +68,7 @@ class RenderExtension extends AbstractExtension
 
         /** @var \Raml\Body $body */
         foreach ($responseBodies as $body) {
-
-            if (!empty($body->getExample())) {
+            if (!empty($body->getExamples()) && !empty($body->getExample())) {
                 $tabs[] = $this->getSchemaFormat($body->getMediaType());
             }
         }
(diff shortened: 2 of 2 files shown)


adriendupuis (approved):

tools/raml2html/src/Twig/Extension/HashExtension.php
```diff
@@ -0,0 +1,35 @@
...
+
+namespace EzSystems\Raml2Html\Twig\Extension;
+
+use RuntimeException;
+use Twig\Extension\AbstractExtension;
+use Twig\TwigFunction;
+
+class HashExtension extends AbstractExtension
+{
+    private $hashes = [];
```
alongosz, line 17: is it possible for us to force higher PHP version here and use its features, like strict type in this case? I'd go for the latest PHP 8.

adriendupuis: @alongosz I moved to PHP8.1+ in #1871. I can take care of this there.

Steveb-p: Technically possible, it's an internal tool.

Well, copied from an open source library afaik.

However, I'm not sure what versions of PHP our doc team have, and I'm not sure if we can just force PHP 8. PHP 7.4 would be more suitable?

alongosz (approved):

tools/raml2html/src/Twig/Extension/HashExtension.php
adriendupuis: I propose #1898 as quick fix for 7.4 as it doesn't work with PHP 7.2 anymore.

For PHP 8.1+, I'll make a survey.

DominikaK: We're all good with at least PHP 8.1

adriendupuis: Let's merge this w/ 7.4 and I'll continue my 8.1 PR afterward.

PR #1898, opened 2023-02-16
raml2html: Upgrade to PHP 7.4

adriendupuis:
Upgrade from PHP 7.2 to PHP 7.4

Fix `Body::getExample()` usage to avoid `PHP Notice:  Trying to access array offset on value of type null in tools/raml2html/vendor/raml-org/raml-php-parser/src/Body.php on line 231` appearing with PHP 7.4

PHP 7.4 is the minimum requirement for commonmark package:
```
  Problem 1
    - Root composer.json requires league/commonmark ^2.3 -> satisfiable by league/commonmark[2.3.0, ..., 2.3.9].
    - league/commonmark[2.3.0, ..., 2.3.9] require php ^7.4 || ^8.0 -> your php version (7.2.34) does not satisfy that requirement.
```

### DIFF
--- a/tools/raml2html/src/Twig/Extension/RenderExtension.php
+++ b/tools/raml2html/src/Twig/Extension/RenderExtension.php
@@ -68,8 +68,7 @@ class RenderExtension extends AbstractExtension
 
         /** @var \Raml\Body $body */
         foreach ($responseBodies as $body) {
-
-            if (!empty($body->getExamples())) {
+            if (!empty($body->getExamples()) && !empty($body->getExample())) {
                 $tabs[] = $this->getSchemaFormat($body->getMediaType());
             }
         }
@@ -85,7 +84,7 @@ class RenderExtension extends AbstractExtension
         foreach ($responseBodies as $body) {
             $schemaFormat = $this->getSchemaFormat($body->getMediaType());
 
-            if (!empty($body->getExamples())) {
+            if (!empty($body->getExamples()) && !empty($body->getExample())) {
                 $examples[$schemaFormat][] = $body->getExample();
             }
         }


tools/raml2html/composer.json
```diff
@@ -33,7 +34,19 @@
             "EzSystems\\Raml2Html\\": "src"
         }
     },
+    "extra": {
+        "patches": {
+            "raml-org/raml-php-parser": {
+                "Fix warnings and depreciations for PHP 7.4": "patches/php_7.4.patch"
```
alongosz, line 40: do you intend to keep this in the upstream or is this temporary change?

tools/raml2html/patches/php_7.4.patch
```diff
@@ -0,0 +1,11 @@
...
++++ /src/Body.php	2023-01-27 15:51:03
+@@ -228,7 +228,7 @@
+      */
+     public function getExample()
+     {
+-        return $this->examples[0];
++        return $this->examples[0] ?? '';
+     }
+
+     /**
```
alongosz, line 11: Looks like we have some trailing characters here and there's no EOL at the end of this file.

adriendupuis: EOL added in 9b726c0fdbb758949f5c2d8cee7f268f7e19bbdc

tools/raml2html/composer.json
adriendupuis: @alongosz  I'll make a PR on https://github.com/raml-org/raml-php-parser/blob/master/src/Body.php#L231 at some point but this is a "middle term" projection. So, this is "temporary" for a while.

alongosz: > @alongosz I'll make a PR on https://github.com/raml-org/raml-php-parser/blob/master/src/Body.php#L231 at some point but this is a "middle term" projection. So, this is "temporary" for a while.

TBH I'm not sure if we should keep this in the upstream at any point. Seems rather unstable and indicates that there are possibly more issues with PHP 7.4 support for that 3rd party.
POV ping @ezsystems/php-dev-team

adriendupuis: @alongosz So, I studied it a little bit deeper and it was fixable on our side: see b79a57bac62578762a0badc36825b960087e28f9
Sorry for the time lost.

alongosz (approved):

tools/raml2html/composer.json
alongosz: No worries, great find :muscle:

PR #1958, opened 2023-04-19
Changed PostgreSQL to 14

MagdalenaZuba:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | https://issues.ibexa.co/browse/IBX-2989
| Versions      | 4.4
| Edition       | all

In version 3.3 it will read 'PostgreSQL 10+ (PostgreSQL 10 has reached its End of Life. We highly recommend to use PostgreSQL 14)

#### Checklist

- [ ] Text renders correctly

### DIFF
--- a/docs/getting_started/requirements.md
+++ b/docs/getting_started/requirements.md
@@ -17,7 +17,7 @@ Recommendation for production setups is to use Varnish/Fastly, Redis/Memcached, 
 
     Review all the recommended versions carefully. If you see a "+" next to the product version, it means that we recommend this version or higher within the same major release. For example, "Nginx 1.18+" means any 1.x version higher or equal to 1.18, but not 2.x.
 
-    Using the latest listed version of each product or component is recommended. Always use a version that receives security updates, either by the vendor themselves or by a trusted third party for example, the distribution vendor.
+    Using the latest listed version of each product or component is recommended. Always use a version that receives security updates, either by the vendor themselves or by a trusted third party, such as the distribution vendor.
 
 ## Operating system
 


docs/getting_started/requirements.md
```diff
@@ -33,7 +33,7 @@ event MPM is recommended, if you need to use prefork you also need the `mod_php`
 
 - MariaDB 10.3+
 - MySQL 8.0
-- PostgreSQL 10+
+- PostgreSQL 14
```
alongosz, line 36: As mentioned internally, I think it would be safer for 3.3 to explicitly recommend it instead of requiring (unless this section is all about recommendation?). Do we intentionally skip `+` sign?

glye, line 36: I'm not too happy about this 😑 For one, we might want to wait for the resolution of https://github.com/ezsystems/ezpublish-kernel/pull/3152 even if it isn't the same thing.

So we say "we recommend this version or higher within the same major release". That's problematic here since postgres mainly uses major versions. If we truly do recommend 14 or higher then sure, fine. But do we? I thought we just recommend currently supported postgres versions, which would be `11 - 15`. I think I would prefer that, once it is QA-ed 😉

glye (changes requested):

docs/getting_started/requirements.md
```diff
@@ -13,7 +13,7 @@ The following server requirements cover both running the software on-premise and
 The minimal setup requires PHP, MySQL/MariaDB, Apache/Nginx, Node.js and `yarn`.
 Recommendation for production setups is to use Varnish/Fastly, Redis/Memcached, NFS/EFS/S3 and Solr/Elasticsearch in a [clustered setup](clustering.md).
 
-!!! note "Recommended versions"
+!!! caution "Recommended versions"
 
     Using the latest listed version of each product or component is always recommended. Review all the recommended versions carefully. If you see a "+" next to the product version, it means that we recommend this version or higher within the same major release. For example, "Nginx 1.18+" means any 1.x version higher or equal to 1.18, but not 2.x.
```
glye, line 18: Maybe we could add to this text something like: "Always use a version that receives security updates, either by the vendor themselves or by a trusted 3rd party like the distribution vendor."

docs/getting_started/requirements.md
micszo: I'm staring to think we should drop the "+" sign all together and replace it with lists explicitly stating (major) versions.

docs/getting_started/requirements.md
```diff
@@ -15,7 +15,9 @@ Recommendation for production setups is to use Varnish/Fastly, Redis/Memcached,
 
 !!! caution "Recommended versions"
 
-    Using the latest listed version of each product or component is always recommended. Review all the recommended versions carefully. If you see a "+" next to the product version, it means that we recommend this version or higher within the same major release. For example, "Nginx 1.18+" means any 1.x version higher or equal to 1.18, but not 2.x.
+    Review all the recommended versions carefully. If you see a "+" next to the product version, it means that we recommend this version or higher within the same major release. For example, "Nginx 1.18+" means any 1.x version higher or equal to 1.18, but not 2.x.
+
+    Using the latest listed version of each product or component is recommended. Always use a version that receives security updates, either by the vendor themselves or by a trusted third party for example, the distribution vendor.
```
glye, line 20: Good reshuffle of the text. 👍  There should be a comma after "party". Also I prefer "like" or "such as", to reduce the number of commas and because there aren't many kinds of trusted 3rd parties here.
```suggestion
    Using the latest listed version of each product or component is recommended. Always use a version that receives security updates, either by the vendor themselves or by a trusted third party, such as the distribution vendor.
```

docs/getting_started/requirements.md
glye: ⬆️ 🔥  You have a point.

glye (approved):
I'm OK with the current state if devs accept postgres 14 here. Though I rather like the idea of listing specific versions like @micszo said. And let's see how the CI PR concludes, please.

MagdalenaZuba:
We will take suggestion about listing all the versions into account, when we will redesign this page for LTS.

adamwojs (approved):

PR #1960, opened 2023-04-19
Fix order criteria and clauses reference

DominikaK:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | https://github.com/ezsystems/developer-documentation/pull/1949
| Versions      | master (4.5)
| Edition       | commerce

Follow-up to https://github.com/ezsystems/developer-documentation/pull/1949, fixing incorrect arguments

#### Checklist

- [x] Text renders correctly
- [x] PHP code samples have been fixed with PHP CS fixer

### DIFF
--- a/docs/search/criteria_reference/order_search_criteria.md
+++ b/docs/search/criteria_reference/order_search_criteria.md
@@ -7,9 +7,9 @@ edition: commerce
 
 Order Search Criteria are part of [Search Criteria](search_criteria_reference.md) and they are supported only by Order Search.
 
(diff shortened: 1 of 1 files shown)


dabrt (approved):

dabrt:
In lines similar to this one:
```
(optional) `sortDirection` - `Created` constant, either `Created::SORT_ASC` or `Created::SORT_DESC`
```
Shouldn't "Created" be shown as FCQN, too?

DominikaK:
I think the short form here is not an issue and makes the text clearer, as long as we have a full example. As long as we keep it consistent with the other sections (payments etc.)

PR #1962, opened 2023-04-20
custom.js: Fix page TOC/hash bug

adriendupuis:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | N/A
| Versions      | All
| Edition       | All

Fix page TOC bug presumably due to MkDocs Material `navigation.tracking` feature (maybe in conflict with our custom scripts):

When on top of a freshly opened page like `/a/page/`,
- a click on right menu item _Section 2_ set the URL to `/a/page/#section-2` and the page scroll to it,
- but the URL hash is then changed to correspond to the section just above like `/a/page/#section-1-2` (the page do not scroll to it).

This is probably what happened to @glye in https://issues.ibexa.co/browse/IBX-5361#description-val where "[password rules page](https://doc.ibexa.co/en/latest/users/user_management/#revoking-passwords)" do not link to the _Password rules_ level-2 section but the one just above, the level-3 section _Reworking passwords_. At least, I experimenting this on Chrome 112.0.5615.49:
1. I go to https://doc.ibexa.co/en/latest/users/user_management/
1. I click on _Password rules_
1. The browser URL becomes https://doc.ibexa.co/en/latest/users/user_management/#password-rules, and the page scroll to this section (the two actions seem simultaneous with naked eyes)
1. After less than a second, the browser URL becomes https://doc.ibexa.co/en/latest/users/user_management/#revoking-passwords

The fix set the URL hash back to the clicked item half a second after the click.
The time between the click and the wrong hash injection is a quarter of second on my machine.
I hope half a second will cover enough cases without bad side effects of re-injecting a hash which was right half a second ago but might be already outdated because of reader's fast navigation and scrolling.

I didn't find how `navigation.tracking` is implemented, less how I could fix it or override it more properly.

I didn't find on our side a script tempering with scroll animation or a style causing sections to overlay eachothers.

#### Checklist

- [ ] Text renders correctly

### DIFF
--- a/docs/js/custom.js
+++ b/docs/js/custom.js
@@ -141,4 +141,11 @@ $(document).ready(function() {
         $('.md-sidebar--primary .md-sidebar__scrollwrap')[0].scrollTop =
         $('.md-sidebar--primary .md-nav__item--active:not(.md-nav__item--nested)')[0].offsetTop - 33;
     }
+
+    // Fix page TOC/hash bug
+    $('.md-sidebar.md-sidebar--secondary nav a').click(function(event) {
+        window.setTimeout(function() {
+            document.location.hash = event.target.hash;
+        }, 500);
+    })
 });


DominikaK:
This had been an issue well before we enabled navigation tracking, but still: it works, and that was an very irritating one.

DominikaK (approved):

glye:
Nice catch!

PR #1967, opened 2023-04-24
IBX-5361: Use NotCompromisedPassword constraint

DominikaK:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | https://issues.ibexa.co/browse/IBX-5361
| Versions      | master (4.5)
| Edition       | all

#### Checklist

- [x] Text renders correctly

### DIFF
--- a/docs/users/user_management.md
+++ b/docs/users/user_management.md
@@ -111,6 +111,7 @@ To do this, in the **User account (ezuser)** Field definition, select "Password 
 ![Protection against using breached passwords](password_breached.png)
 
 This rule checks the password against known password dumps by using the https://haveibeenpwned.com/ API.
+It does not check existing passwords, so it will not block login for anyone. It applies only to new passwords when users change them.
 
 !!! note
 


docs/users/user_management.md
```diff
@@ -103,6 +103,21 @@ A password that has been used before can be used again.
...
 This rule is valid by default when password expiration is set.
 
+### Breached passwords
+
+You can set a rule that prevents using passwords which have been exposed in a public breach.
+To do this, in the **User account (ezuser)** Field definition, select "Password must not be contained in a public breach".
+
+![Protection against using breached passwords](password_breached.png)
+
+This rule checks the password against known password dumps by using the https://haveibeenpwned.com/ API.
```
glye, line 113:
```suggestion
This rule checks the password against known password dumps by using the https://haveibeenpwned.com/ API. It does not check passwords that are in use, so it will not block login for anyone. It applies to new passwords when users are changing their passwords. 
```

I realised it might be good to mention this point.

glye (approved):
Just one addition, move or reword it as you find best.

Steveb-p (approved):

docs/users/user_management.md
glye: @Steveb-p We could check existing passwords during login, but I don't want to do that.

(I was considering to add a "check my password"-button on the login form and/or in the user settings, but concluded that's an unnecessary complication. When enabled it will apply to all new users registering, and if site owners really want to they can enable it and then expire all passwords. That will ensure all passwords are un-breached, and all users are pissed off 😁 )

DominikaK, line 113:
```suggestion
This rule checks the password against known password dumps by using the https://haveibeenpwned.com/ API.
It does not check existing passwords, so it will not block login for anyone. It applies only to new passwords when users change them.
```

glye: 👍

MagdalenaZuba (approved):

PR #2011, opened 2023-06-12
IBX-5805: Vale.sh PoC

ibexa-yuna:
PoC PR to allow Vale.sh to post result

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -41,7 +41,7 @@ jobs:
                 run: |
                     mkdocs build --strict
 
-    vale:
+    vale-check:
         runs-on: ubuntu-latest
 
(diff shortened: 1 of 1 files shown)


DominikaK:
@ibexa-yuna, if I understand correctly, `filter_mode: diff_context` means the action will only check lines modified in a given PR, correct?

ibexa-yuna:
> @ibexa-yuna, if I understand correctly, `filter_mode: diff_context` means the action will only check lines modified in a given PR, correct?

As per https://github.com/reviewdog/reviewdog#filter-mode
Essentially, yes.

DominikaK (approved):

adamwojs (approved):

PR #2031, opened 2023-06-23
Fixes REST API examples

DominikaK:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | n/a
| Versions      | v4+
| Edition       | n/a

Fixes two invalid REST API endpoint examples.

#### Checklist

- [x] Text renders correctly


docs/api/rest_api/rest_api_reference/input/examples/user/groups/path/users/POST/UserCreate.xml.example
```diff
@@ -1,20 +1,32 @@
...
+				<fieldValue>
+					<value key="hasStoredLogin">true</value>
+					<value key="contentId">14</value>
+					<value key="login">johnsmith</value>
+					<value key="email">example@example.com</value>
+					<value key="enabled">true</value>
+					<value key="maxLogin">10</value>
+				</fieldValue>
+			</field>
+	</fields>
```
alongosz, line 31: Formatting doesn't look good here. I think you have tabs instead of spaces. It should be always spaces and AFAICS we have here convention of 2 spaces indents.

docs/api/rest_api/rest_api_reference/input/examples/user/roles/id/policies/POST/PolicyCreate.xml.example
```diff
@@ -5,12 +5,12 @@
     <limitations>
         <limitation identifier="Class">
             <values>
-                <ref href="/api/ibexa/v2/content/types/13"/>
+                <ref media-type="application/vnd.ibexa.api.ref+xml" href="2"/>
```
alongosz, line 8: Oh my... This is definitely not correct, but partially on the implementation level, not the doc.

Question: do you need this `media-type`? `application/vnd.ibexa.api.ref+xml` is not a recognized type anywhere. It rather should be a type describing what `href` is, however href is not correct either. Can we for now skip media-type?

As for `href` something clearly didn't come together during the implementation. What you had before was what it should be, what you have now is what actually works... Reported as [IBX-6028](https://issues.ibexa.co/browse/IBX-6028).

alongosz (changes requested):

docs/api/rest_api/rest_api_reference/input/examples/user/groups/path/users/POST/UserCreate.xml.example
```diff
@@ -1,20 +1,32 @@
...
+        <fieldValue>
+          <value key="hasStoredLogin">true</value>
+          <value key="contentId">14</value>
+          <value key="login">johnsmith</value>
+          <value key="email">example@example.com</value>
+          <value key="enabled">true</value>
+          <value key="maxLogin">10</value>
+        </fieldValue>
+      </field>
+	</fields>
```
alongosz, line 31: misaligned closing tag, it should be on the same level like `<fields>` in the line 8.

alongosz (approved):

MagdalenaZuba (approved):

PR #2032, opened 2023-06-23
ENG-69: Revamp documentation design: homepage

GrabowskiM:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | https://issues.ibexa.co/browse/ENG-69

I added scss compiler and recreated front-page using scss for better readability and coding using nesting, variables and mixins.

I put all of this in single front-page.scss file, as currently script to parse scss directory is dumb and transforms all files on save instead of only front-page.scss, but I'm not sure if in this case it's worth writing specific rules (especially if in future there can be other files like custom.scss etc), but I'm open on criticism and suggestions here. :)

Selection_696

Selection_697


lucasOsti (approved):

adamwojs:
@GrabowskiM Could you please attach screenshot to PR description ? :wink:

Gengar-i (approved):

dew326 (approved):

docs/index.md
```diff
@@ -1,115 +1,217 @@
...
+                    <summary>
+                        <h2>Notable changes in v4.5</h2>
+                        <div class="accordion__toggler">
+                            <svg><use xlink:href="images/icons.svg#toggler" /></svg>
+                        </div>
+                    </summary>
+                    <div class="row">
+                        <div class="col-12 col-lg-6 col-fhd-3">
+                            <ul>
+                                <li><a href="api/php_api/php_api/">PHP API</a></li>
```
DominikaK, line 99: The links have switched places: These should go under "Most popular pages", and links from most popular here.

dabrt:
While looking at the preview I noticed that the “The latest release" rectangle is narrower than the tiles above and below. Is this how it is supposed to be?
obraz

GrabowskiM:
> While looking at the preview I noticed that the “The latest release" rectangle is narrower than the tiles above and below. Is this how it is supposed to be? obraz

Hi, could you recheck? There should be image on right side of this block, perhaps there was some problem with this svg on mac, I changed it to png, it should work now.

dabrt:
> > While looking at the preview I noticed that the “The latest release" rectangle is narrower than the tiles above and below. Is this how it is supposed to be? obraz
> 
> Hi, could you recheck? There should be image on right side of this block, perhaps there was some problem with this svg on mac, I changed it to png, it should work now.

It looks perfect now, with the image displaying where it should.
obraz

PR #2078, opened 2023-07-24
Default regions value

julitafalcondusza:
Default regions value

| Question      | Answer
| ------------- | ---
| JIRA Ticket   | N/A
| Versions      | master
| Edition       | Commerce

- Sentence about `default` value for regions added
- Correction: Commerce -> Product Catalog

#### Checklist

- [x] Text renders correctly


micszo (approved):
Thanks!

adamwojs (approved):

docs/pim/enable_purchasing_products.md
```diff
@@ -45,6 +45,19 @@ ibexa:
...
+
+```yaml
+ibexa:
+    system:
+        storefront_group:
+            product_catalog:
+                currencies:
+                    - EUR
+                regions:
+                    - default
```
adriendupuis, line 58: Maybe in the VAT rates example below, we should add the `default` region to avoid `An exception has been thrown during the rendering of a template ("Could not find 'Ibexa\Contracts\ProductCatalog\Values\RegionInterface' with identifier 'default'").`

I can't make suggestion outside PR scope so I put it here:

``` yaml
ibexa:
    repositories:
        default:
            product_catalog:
                engine: default
                regions:
                    default: ~
                    germany:
                        vat_categories:
                            standard: 19
                            reduced: 7
                            none: ~
                    uk:                                               
                        vat_categories:                            
                            standard: 20                            
                            reduced: 5                            
                            none: ~
                    poland:                                               
                        vat_categories:                            
                            standard: 23                            
                            reduced: 8                            
                            none: ~
```

adriendupuis: @micszo @adamwojs A reaction to this idea? I came to this because I crashed my own front (4.6.0-beta1) by copy-pasting the examples directly in config/packages/ibexa.yaml

micszo: @adriendupuis these are great ides because then user doesn't have to take any action "ootb". Both pieces of configuration can coexist.
One thing that comes to mind, we need to consider how this could impact other default configurations, namely Nightly setup.

micszo: @adriendupuis to double-check, do you have in mind adding such an idea to the doc? or to the default config in a clean installation?

adriendupuis: @micszo My goal is only to have the doc example working when its three code samples are used. The default config coming with the clean install is out of PR scope and probably should stay unchanged.

adriendupuis, line 58: After discussion, the example should be coherent with each others, and `ibexa.system.<siteaccess_or_siteaccess_group>.product_catalog.regions` must use the region defined in `ibexa.repositories`. So, this example should, from my POV, use the regions from the previous and incoming examples. It is also the occasion to show that those regions can be split over several siteaccess and siteaccess groups.
I propose the following example:

```suggestion
ibexa:
    system:
        storefront_group:
            product_catalog:
                currencies:
                    - EUR
                    - PLN
                regions:
                    - germany
                    - poland 
        another_storefront_group:
            product_catalog:
                currencies:
                    - GBP
                regions:
                    - uk
```
With this example, the 3rd example won't crash the storefront.
:warning: Important counter part: This second example can't be used without the third one or it crash the storefront.
There should be an instruction that link this 2nd example with the 3rd.

docs/pim/enable_purchasing_products.md
```diff
@@ -45,6 +45,19 @@ ibexa:
...
+```
+
```
adriendupuis, line 60:
```suggestion

This example uses the currencies and regions set in the [VAT example below](#vat-rates).

```

ViniTou (approved):

MagdalenaZuba (approved):

adriendupuis (approved):

PR #2115, opened 2023-08-24
Fix CP configuration

dabrt:
| Question      | Answer |
| ------------- | ---
| JIRA Ticket   | none |
| Versions      | 4.5 |
| Edition       |  Commerce |

Add a mention that one must have all potential currencies configured before CP users can review their orders.

#### Checklist

- [x] Text renders correctly

### DIFF
--- a/docs/customer_management/cp_configuration.md
+++ b/docs/customer_management/cp_configuration.md
@@ -92,6 +92,6 @@ ibexa:
 
 # Order management
 
-Reviewing pending and past orders in Customer Portal requires that you configure all currencies that any of the customers may use under the `ibexa.system.default.product_catalog.currencies` key. 
+Reviewing pending and past orders in Customer Portal requires that you configure all currencies that any of the customers may use under the `ibexa.system.<siteaccess_name>.product_catalog.currencies` key. 
 The first currency from the list is then used for filtering the orders list and calculating the **Average order** and **Total amount** values.
 For more information, see [Enable purchasing products](https://doc.ibexa.co/en/latest/pim/enable_purchasing_products/).


docs/customer_management/cp_configuration.md
```diff
@@ -89,3 +89,8 @@ ibexa:
                         match:
                             Identifier\ContentType: confirmation_page
 ```
+
+# Order management
+
+Reviewing pending and past orders in Customer Portal requires that you configure all currencies that any of the customers may use under the `ibexa.system.default.product_catalog.currencies` key.
```
konradoboza, line 95: Feel free to rephrase it if needed, we need to emphasize that first configured currency will be the one that the list is showing:
```suggestion
Reviewing pending and past orders in Customer Portal requires that you configure all currencies that any of the customers may use under the `ibexa.system.default.product_catalog.currencies` key. The first currency from the list will be taken for filtering orders list as well as for calculating "Average order" and "Total amount".
```

konradoboza (approved):

docs/customer_management/cp_configuration.md
```diff
@@ -89,3 +89,9 @@ ibexa:
...
```
adamwojs, line 95: Why currencies list needs to be defined in `default` scope ?

konradoboza: Good catch, it actually doesn't. We can change it to sth like:
`ibexa.system.<your_siteaccess>.product_catalog.currencies`.

julitafalcondusza (approved):

adamwojs (approved):

docs/customer_management/cp_configuration.md
```diff
@@ -89,3 +89,9 @@ ibexa:
                         match:
                             Identifier\ContentType: confirmation_page
 ```
+
+# Order management
+
+Reviewing pending and past orders in Customer Portal requires that you configure all currencies that any of the customers may use under the `ibexa.system.<siteaccess_name>.product_catalog.currencies` key. 
+The first currency from the list is then used for filtering the orders list and calculating the **Average order** and **Total amount** values.
+For more information, see [Enable purchasing products](https://doc.ibexa.co/en/latest/pim/enable_purchasing_products/).
```
adriendupuis, line 97: It shouldn't have been passed as an external link.
`[Enable purchasing products](enable_purchasing_products.md)` would have been better.

PR #2211, opened 2023-11-06
Fix event source class for calendar customizing

ghost:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | 
| Versions      | 
| Edition       | 

#### Checklist

- [ ] Text renders correctly

### DIFF
--- a/code_samples/back_office/calendar/config/custom_services.yaml
+++ b/code_samples/back_office/calendar/config/custom_services.yaml
@@ -10,6 +10,7 @@ services:
             $eventType: '@App\Calendar\Holidays\EventType'
 
     App\Calendar\Holidays\EventSource:
+        class: Ibexa\Calendar\EventSource\InMemoryEventSource
         factory: [ '@App\Calendar\Holidays\EventSourceFactory', 'createEventSource' ]
         tags:
             - { name: ibexa.calendar.event.source }


adamwojs (approved):

code_samples/back_office/calendar/config/custom_services.yaml
```diff
@@ -10,7 +10,7 @@ services:
             $eventType: '@App\Calendar\Holidays\EventType'
 
     App\Calendar\Holidays\EventSource:
-        class: Ibexa\Tests\Calendar\Calendar\EventSource
-        factory: [ '@App\Calendar\Holidays\EventSourceFactory','createEventSource' ]
+        class: Ibexa\Calendar\EventSource\InMemoryEventSource
```
adriendupuis, line 13: Hi,

Thank you for highlighting the use of a Tests class here.

We do not need to link this service class to another class than itself.

```suggestion
```

adriendupuis (changes requested):
Hi @ms2ad,

Thank you for your contribution.

~~In fact, the right class is `App\Calendar\Holidays\EventSource` itself. A `class` property is not necessary as the actual class name as service ID is self-sufficient.~~
Edit: You're right about the real class being returned by the factory. The fake class name as service ID is a bit lazt but OK.

code_samples/back_office/calendar/config/custom_services.yaml
ms2ad: Well, I admit I was a bit lazy when creating this PR and not tackling the "other issue". `App\Calendar\Holidays\EventSource` is no class. In this case, it is just a service name looking like a class. If you look in `App\Calendar\Holidays\EventSourceFactory::createEventSource()`, then an `Ibexa\Calendar\EventSource\InMemoryEventSource` object is returned and not a `App\Calendar\Holidays\EventSource` object.

May-be the service should be renamed to a dot-separated string.

adriendupuis: @ms2ad  No need to name the service as it won't be used elsewhere.
If you look at `php bin/console debug:container --tag=ibexa.calendar.event.source` none is named 😉

adriendupuis (approved):

code_samples/back_office/calendar/config/custom_services.yaml
ms2ad: @adriendupuis While I cannot run the command right now, I checked the code and every service like `Ibexa\Scheduler\Calendar\FutureHideEventSource` is an acutal class in contrast to `App\Calendar\Holidays\EventSource` which is no class (see above).

adriendupuis: But you're right about this not being a true class.
And https://symfony.com/doc/5.4/service_container/factories.html seems to prefer to name the service with output of the factory…
OK. Let's precise the real output class. The fake class as a service name is not problematic. But we are the lazy ones here, not you.

adriendupuis: That looks OK to me with the "Class name" column indicating the real class and the "Service ID" column having the same naming pattern.

```
------------------------------------------------------- ------------------------------------------------------- 
  Service ID                                              Class name                                             
 ------------------------------------------------------- ------------------------------------------------------- 
  App\Calendar\Holidays\EventSource                       Ibexa\Calendar\EventSource\InMemoryEventSource         
  Ibexa\Scheduler\Calendar\FuturePublicationEventSource   Ibexa\Scheduler\Calendar\FuturePublicationEventSource  
  Ibexa\Scheduler\Calendar\FutureHideEventSource          Ibexa\Scheduler\Calendar\FutureHideEventSource         
  Ibexa\FieldTypePage\Calendar\BlockRevealEventSource     Ibexa\FieldTypePage\Calendar\BlockRevealEventSource    
  Ibexa\FieldTypePage\Calendar\BlockHideEventSource       Ibexa\FieldTypePage\Calendar\BlockHideEventSource      
 ------------------------------------------------------- -------------------------------------------------------
 ```

PR #2212, opened 2023-11-06
Add taxonomy name to loadEntryByIdentifier() examples

ghost:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | 
| Versions      | 
| Edition       | 

#### Checklist

- [ ] Text renders correctly

### DIFF
--- a/code_samples/api/public_php_api/src/Command/TaxonomyCommand.php
+++ b/code_samples/api/public_php_api/src/Command/TaxonomyCommand.php
@@ -41,7 +41,7 @@ class TaxonomyCommand extends Command
 
         $allEntries = $this->taxonomyService->loadAllEntries('tags', 30, 0);
 
-        $entry = $this->taxonomyService->loadEntryByIdentifier('desks');
+        $entry = $this->taxonomyService->loadEntryByIdentifier('desks', 'tags');
 
         $output->writeln($entry->name . ' with parent ' . $entry->parent->name);
 
@@ -52,12 +52,12 @@ class TaxonomyCommand extends Command
             $output->writeln($child->name);
         }
 
-        $entryToMove = $this->taxonomyService->loadEntryByIdentifier('standing_desks');
-        $newParent = $this->taxonomyService->loadEntryByIdentifier('desks');
+        $entryToMove = $this->taxonomyService->loadEntryByIdentifier('standing_desks', 'tags');
+        $newParent = $this->taxonomyService->loadEntryByIdentifier('desks', 'tags');
 
         $this->taxonomyService->moveEntry($entryToMove, $newParent);
 
-        $sibling = $this->taxonomyService->loadEntryByIdentifier('school_desks');
+        $sibling = $this->taxonomyService->loadEntryByIdentifier('school_desks', 'tags');
         $this->taxonomyService->moveEntryRelativeToSibling($entryToMove, $sibling, TaxonomyServiceInterface::MOVE_POSITION_PREV);
 
         return self::SUCCESS;
--- a/docs/content_management/taxonomy/taxonomy_api.md
+++ b/docs/content_management/taxonomy/taxonomy_api.md
@@ -9,18 +9,20 @@ To manage taxonomies, use `Ibexa\Contracts\Taxonomy\Service\TaxonomyServiceInter
 ## Getting taxonomy entries
 
 To get a single taxonomy entry, you can use `TaxonomyServiceInterface::loadEntryById()`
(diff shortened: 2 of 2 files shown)


code_samples/api/public_php_api/src/Command/TaxonomyCommand.php
```diff
@@ -41,7 +41,7 @@ public function execute(InputInterface $input, OutputInterface $output)
 
         $allEntries = $this->taxonomyService->loadAllEntries('tags', 30, 0);
 
-        $entry = $this->taxonomyService->loadEntryByIdentifier('desks');
+        $entry = $this->taxonomyService->loadEntryByIdentifier('desks', 'tags');
```
adriendupuis, line 44: Hi,

This is not mandatory. If not given, [`TaxonomyService::loadEntryByIdentifier`](https://github.com/ibexa/taxonomy/blob/main/src/lib/Service/TaxonomyService.php#L77) asks [`TaxonomyConfiguration::getDefaultTaxonomyName`](https://github.com/ibexa/taxonomy/blob/main/src/lib/Service/TaxonomyConfiguration.php#L86-L100) which returns [`'tags'`](https://github.com/ibexa/taxonomy/blob/main/src/lib/Service/TaxonomyConfiguration.php#L26). I confirmed that on a fresh install of a 4.5.3. If you have a bug making `'tags'` not being this default, can you, please, try to explain a bit more why you had to add this?

docs/content_management/taxonomy/taxonomy_api.md
```diff
@@ -9,18 +9,20 @@ To manage taxonomies, use `Ibexa\Contracts\Taxonomy\Service\TaxonomyServiceInter
...
 
 To get a single taxonomy entry, you can use `TaxonomyServiceInterface::loadEntryById()`
-and provide it with the numerical entry ID, or pass entry identifier and use `TaxonomyServiceInterface::loadEntryByIdentifier()`:
+and provide it with the numerical entry ID, or pass the entry identifier and taxonomy name to `TaxonomyServiceInterface::loadEntryByIdentifier()`:
 
 ``` php
 [[= include_file('code_samples/api/public_php_api/src/Command/TaxonomyCommand.php', 43, 46) =]]
 ```
 
+If no taxonomy name is given, it defaults to `Ibexa\Taxonomy\Service\TaxonomyConfiguration::getDefaultTaxonomyName()`.
```
adriendupuis, line 18: You're right on this point. e should mention that a taxonomy name can be passed if the currently used taxonomy is not `'tags'`.

code_samples/api/public_php_api/src/Command/TaxonomyCommand.php
adriendupuis: For the example, we won't pass the default value of the default installation. But, as seen in your taxonomy_api.md change proposal, we'll mention that a second value can be passed when using another taxonomy.

adriendupuis (commented):
Counter-proposal in #2213

code_samples/api/public_php_api/src/Command/TaxonomyCommand.php
ms2ad: I am guessing here a bit, but I checked and in 4.0.8 that there, the parameter was not present and I have to admit that I only recently learned that identifiers were not globally unique (anymore) but only unique per taxonomy. Without fully setting up an older Ibexa DXP installation, I am sure that when testing taxonomy the first time last year, that was the case so that previously, there was no need for this parameter. I assume to keep backward compability, the parameter value was added as an optional value and "tags" was used as the best possible fallback parameter. But from a conceptual point, in my opinion, it does not really make sense to not expect the taxonomy name if the identifier is not globally unique. Then, a pair of taxonomy entry identifier and taxonomy name must always be given.

In multi-taxonomy setups, this code example of not specifying the taxonomy name might lead to sub-optimal code if in different taxonomies, the same identifier might be used.

I just think that the documentation code should contain as little obstacles as possible for code that will work in as many cases as possible.

It appears that only in `vendor/ibexa/installer/upgrade/db/mysql/ibexa-4.4.latest-to-4.5.0.sql` (i.e. in the update from 4.4 to 4.5) was the constaint for global uniqueness of the taxonomy entry identifier column was removed. May-be this topic of what combination of taxonomy entry data (just identifier or identifier+taxonomy name) was unique in which version is something that is missing in the documentation and should be added if it is really missing in the documentation/release notes and not just something I missed?

adriendupuis:
Covered in #2213 and #2215

Thanks a lot for your contribution @ms2ad!

PR #2237, opened 2023-12-05
PHP 8.0 is EOL now

glye:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | N/A
| Versions      | All
| Edition       | All

PHP 8.0 has reach end of life, so we should update the info in Requirements.
(The last change in the diff must be an end-of-line change or something like that, caused by the github in-place editing interface.)

Aside: Can we really say "you should use PHP 8.1" for v2.5, where PHP 8 isn't supported at all? Should we rather say "you should upgrade to Ibexa DXP v3.3 or v4.5"?

#### Checklist

- [ ] Text renders correctly
- [x] ~Code samples are working~
- [x] ~PHP code samples have been fixed with PHP CS fixer~

### DIFF
--- a/docs/getting_started/requirements.md
+++ b/docs/getting_started/requirements.md
@@ -125,10 +125,10 @@ Recommendation for production setups is to use Varnish/Fastly, Redis/Memcached, 
 
 === "eZ Platform v2.5"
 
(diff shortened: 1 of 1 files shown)


MagdalenaZuba (approved):

glye:
Added "upgrade to v3.3 / v4.x" for v2.5 users.

PR #2273, opened 2024-01-23
locationId parameter is missing in call to BlockController::renderAct…

vidarl:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | 
| Versions      | 3.3, 4.5
| Edition       | Content/Headless, Experience, Commerce

The template is missing `locationId` parameter which is needed. See https://github.com/ezsystems/ezplatform-page-fieldtype/blob/2.3/src/bundle/Resources/views/default_layout.html.twig#L9

Doc pages:
https://doc.ibexa.co/en/3.3/guide/content_rendering/render_content/render_page/#configure-layout
https://doc.ibexa.co/en/4.5/templating/render_content/render_page/#configure-layout

#### Checklist

- [ ] Text renders correctly

### DIFF
--- a/code_samples/front/render_page/templates/themes/my_theme/layouts/sidebar.html.twig
+++ b/code_samples/front/render_page/templates/themes/my_theme/layouts/sidebar.html.twig
@@ -4,6 +4,7 @@
             {% for block in zones[0].blocks %}
                 <div class="landing-page__block block_{{ block.type }}" data-ez-block-id="{{ block.id }}">
                     {{ render_esi(controller('EzSystems\\EzPlatformPageFieldTypeBundle\\Controller\\BlockController::renderAction', {
+                        'locationId': locationId,
                         'contentId': contentInfo.id,
                         'blockId': block.id,
                         'versionNo': versionInfo.versionNo,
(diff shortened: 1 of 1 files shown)


adamwojs:
@vidarl Could you please rebase your PR ?

vidarl:
@adamwojs : Rebased, and CI issues are fixed again

adamwojs (approved):

adriendupuis:
The controller itself doesn't seem to make location mandatory https://github.com/ezsystems/ezplatform-page-fieldtype/blob/2.3/src/bundle/Controller/BlockController.php#L72

@vidarl Did you encounter an error when using it without this `locationId` parameter?

dabrt:
Please see update_db_to_2.5.md file, which has "Page migration example" section, if a code sample should be changed there as well.

vidarl:
> The controller itself doesn't seem to make location mandatory https://github.com/ezsystems/ezplatform-page-fieldtype/blob/2.3/src/bundle/Controller/BlockController.php#L72
> 
> @vidarl Did you encounter an error when using it without this `locationId` parameter?

The argument is nullable, but has no default value. But Symfony seems indeed able to inject `null` value in this case ( I was not aware that Symfony would honestly) . A client claimed that translating landing page doesn't work without that argument, but `$location` is never used inside the controller so I doubt that is the case. I'll double check with the client and come back to you. Likely the PR can be closed

vidarl:
Client now reports back that he is not able to reproduce the issue. I'll close this PR, seems like it is not needed

PR #2309, opened 2024-02-20
Add instructions for making block attribute required with custom validator

ghost:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | 
| Versions      | master, 4.6, 4.5
| Edition       | 

#### Checklist

- [x] Text renders correctly

### DIFF
--- a/docs/content_management/pages/page_block_validators.md
+++ b/docs/content_management/pages/page_block_validators.md
@@ -71,7 +71,7 @@ Finally, add the validator to one of your block attributes, for example:
 
 By default, only `not_blank` and `not_blank_richtext` validators mark a block attribute as required.
 
-If you create a custom validator `custom_not_blank` with attribute-specific logic, you can extend the `AttributeType` class with a Symfony form type extension to ensure that the attribute is also considered required:
+If you create a custom validator `custom_not_blank` with attribute-specific logic, you can extend the `AttributeType` class with a Symfony form type extension to make sure that the attribute is also considered required:
 
 ``` php hl_lines="16"
 [[= include_file('code_samples/page/custom_block_validator/src/Form/Extension/AttributeTypeExtension.php') =]]


Steveb-p (approved):
Using form extension to work with custom options is the way to go :+1:

Note that this applies to Attribute Definitions, which as I understand it is slightly different from Attributes. Attribute Definition is a whole group which represents an Attribute (`value`), including it's `name` and `id`.

See `Ibexa\PageBuilder\Form\Type\Attribute\AttributeType::buildForm()`, which in turn creates specific Attribute value types (using a mapper).

To make this easier on developers (we should gauge the need for it), we could expose a configuration option to add more constraints that should be treated as marking the form field as required :thinking:

ms2ad:
@Steveb-p I see what you mean that whether using the whole `$builder` object might be the right case because all fields are children of it, including the name and id field, not just the value.

> To make this easier on developers (we should gauge the need for it), we could expose a configuration option to add more constraints that should be treated as marking the form field as required 🤔

Your current implementation might also benefit from it because while having two specific validators hard-coded in `AttributeType::buildForm()` works, but is not the most clean code imaginable :)

mnocon (approved):
Thank you for the contribution! 🙇

dabrt (approved):

docs/content_management/pages/page_block_validators.md
```diff
@@ -66,3 +66,12 @@ Finally, add the validator to one of your block attributes, for example:
 ``` yaml hl_lines="16-18"
 [[= include_file('code_samples/page/custom_block_validator/config/packages/page_blocks.yaml', 0, 1) =]][[= include_file('code_samples/page/custom_block_validator/config/packages/page_blocks.yaml', 3, 20) =]]
 ```
+
+### Custom required validator
+
+By default, only `not_blank` and `not_blank_richtext` cause a block attribute to be marked as required.
+If you create a custom validator `custom_not_blank` with attribute-specific logic, you can add extend the `AttributeType` class with a Symfony form type extension to ensure that the attribute is also considered required:
+
+``` php
```
adriendupuis, line 75: I would highlight `$builder->setRequired(true);`:
```suggestion
``` php hl_lines="16"
```

docs/content_management/pages/page_block_validators.md
```diff
@@ -66,3 +66,12 @@ Finally, add the validator to one of your block attributes, for example:
 ``` yaml hl_lines="16-18"
 [[= include_file('code_samples/page/custom_block_validator/config/packages/page_blocks.yaml', 0, 1) =]][[= include_file('code_samples/page/custom_block_validator/config/packages/page_blocks.yaml', 3, 20) =]]
 ```
+
+### Custom required validator
+
+By default, only `not_blank` and `not_blank_richtext` cause a block attribute to be marked as required.
```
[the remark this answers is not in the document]
adriendupuis: Nice to have a list of validators automatically triggering the "required" mark. It justifies a new section by itself. I would also add a new paragraph separator:

```suggestion
By default, only `not_blank` and `not_blank_richtext` validators mark a block attribute as required.

```

adriendupuis (changes requested):
~~See review comments~~ ✅ applied

adriendupuis (approved):

docs/content_management/pages/page_block_validators.md
```diff
@@ -66,3 +66,13 @@ Finally, add the validator to one of your block attributes, for example:
 ``` yaml hl_lines="16-18"
 [[= include_file('code_samples/page/custom_block_validator/config/packages/page_blocks.yaml', 0, 1) =]][[= include_file('code_samples/page/custom_block_validator/config/packages/page_blocks.yaml', 3, 20) =]]
 ```
+
+### Custom required validator
+
+By default, only `not_blank` and `not_blank_richtext` validators mark a block attribute as required.
+
+If you create a custom validator `custom_not_blank` with attribute-specific logic, you can extend the `AttributeType` class with a Symfony form type extension to ensure that the attribute is also considered required:
```
adriendupuis, line 74: https://learn.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/e/ensure-insure-assure
```suggestion
If you create a custom validator `custom_not_blank` with attribute-specific logic, you can extend the `AttributeType` class with a Symfony form type extension to make sure that the attribute is also considered required:
```

adriendupuis:
@ms2ad Thank you!

PR #2324, opened 2024-03-05
IBX-7873: Add links to interactive demos to dev guide

dabrt:
| Question      | Answer |
| ------------- | --- |
| JIRA Ticket   | IBX-7873 |
| Versions      | 4.6 and up |
| Edition       | all |

Add links to interactive demos to dev guide

#### Checklist

- [x] Text renders correctly


mkdocs.yml
```diff
@@ -331,6 +331,7 @@ nav:
         - Add remote PIM support: pim/add_remote_pim_support.md
     - Commerce:
         - Commerce: commerce/commerce.md
+        - Interactive demo test: commerce/interactive_demo.md
```
mnocon, line 334: Embedding the full demo doesn't work for me, it covers navigation and breaks the page. I suppose we should remove this entry and the test file?

mnocon (commented):
Thanks, I like the result!

mkdocs.yml
mnocon: Yep, I see that now, I didn't know it when writing this comment 😄

juskora:
General comment, I think we can discuss about a separate space for all those videos 'cause in this form it looks little bit messy as these video are placed in different places, hard to find at first glance and lacks consistency.

julitafalcondusza:
I agree with @juskora - it lacks consistency. Each page should have same placement of the video. 
In "Content management product guide" you can easily miss interactive demo - it is placed in the middle of the content, between many other images, so you just keep scrolling. 
I like the idea of "Click to see an interactive demo" frame. I think that you could use it in all the pages.
So: same placement, same style.

PR #2407, opened 2024-06-18
Marked DB update to 4.5.3 as Experience and Commerce only

mnocon:
Applying internal feedback:
```
hello, can we make some annotation here to let people know that this migration should be run on editions higher or equal to Experience? https://doc.ibexa.co/en/latest/update_and_migration/from_4.4/update_from_4.4/#v453
```

The file contains only changes for Form Builder, which is available in Experience and Commerce:
https://github.com/ibexa/installer/blob/main/upgrade/db/mysql/ibexa-4.5.2-to-4.5.3.sql

I'm mentioning Content because in that version it's still called by that name.

I realize the sentence is a bit clunky - suggestions are welome 😄

### DIFF
--- a/docs/update_and_migration/from_4.4/update_from_4.4.md
+++ b/docs/update_and_migration/from_4.4/update_from_4.4.md
@@ -250,7 +250,6 @@ Run the following scripts:
 
 #### Database update [[% include 'snippets/experience_badge.md' %]] [[% include 'snippets/commerce_badge.md' %]]
 
-Update the database if you are using [[= product_name_exp =]] or [[= product_name_com =]]. [[= product_name_content =]] does not require the database update.
 Run the following scripts:
 
 === "MySQL"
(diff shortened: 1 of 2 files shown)


konradoboza (approved):

barw4 (approved):

docs/update_and_migration/from_4.4/update_from_4.4.md
```diff
@@ -236,31 +236,32 @@ Run the following scripts:
...
     psql <database_name> < vendor/ibexa/installer/upgrade/db/postgresql/ibexa-4.5.1-to-4.5.2.sql
     ```
 
 ### v4.5.3
 
-#### Database update
+#### Database update [[% include 'snippets/experience_badge.md' %]] [[% include 'snippets/commerce_badge.md' %]]
 
+Update the database if you are using [[= product_name_exp =]] or [[= product_name_com =]]. [[= product_name_content =]] does not require the database update.
 Run the following scripts:
```
dabrt, line 254:
```suggestion
If you are using [[= product_name_exp =]] or [[= product_name_com =]], run the following scripts:
```
or actually

```suggestion
Run the following scripts:
```
I don't think you need the sentence at all. We've been using badges above and I think their meaning is clear - whatever is marked with one applies to the indicated edition.

Thus my suggestion, which applies to the other occurrence as well.

PR #2422, opened 2024-07-02
Improvements to the Install DXP page

mnocon:
Target: master, 4.6

Things done:
1) Fixed missing space and wrong usage of `eventually`. It's a trap that I fall into quite easily, so I started noticing it :D See https://quicksilvertranslate.com/243/false-friends-eventually/
2) Removed things that are no longer relevant:
- search in the frontend is no longer needed, since we removed the Old Commerce solution (but it's still strongly recommended)
- JMS secret is no longer required - it's gone with the Old Commerce
- the `.gitignore` file is already present in recent releases: https://github.com/ibexa/experience-skeleton/blob/v4.6.7/.gitignore
3) We need to send people to Composer doc to install it, because checking the checksum is a security recommendation and we should not advise against it
4) I want to change the header name for setting up the HTTP server - right now it mentions `production`, but the very first instruction is for a development setup.
5) Fixed list formatting in the Elasticsearch article

Previews:
Old: https://doc.ibexa.co/en/latest/getting_started/install_ibexa_dxp/
New: https://ez-systems-developer-documentation--2422.com.readthedocs.build/en/2422/getting_started/install_ibexa_dxp/

Old: https://doc.ibexa.co/en/master/search/search_engines/elastic_search/configure_elastic_search/
New:https://ez-systems-developer-documentation--2422.com.readthedocs.build/en/2422/search/search_engines/elastic_search/configure_elastic_search/

### DIFF
--- a/docs/getting_started/install_ibexa_dxp.md
+++ b/docs/getting_started/install_ibexa_dxp.md
@@ -13,7 +13,7 @@ description: Install Ibexa DXP on a Linux system and prepare your installation f
 
 !!! note "Installing [[= product_name_oss =]]"
 
-    This installation guide details the steps to install [[= product_name =]] for users who have a subscription agreement with [[= product_name_base =]].
+    This installation guide shows in details how to install [[= product_name =]] for users who have a subscription agreement with [[= product_name_base =]].
     If you want to install [[= product_name_oss =]], you do not need authentication tokens or an account on updates.ibexa.co,
     but must adapt the steps shown here to the product edition and the `ibexa/oss-skeleton` repository.
 
@@ -50,7 +50,7 @@ composer -V
 
 !!! tip "Install Composer locally"
 
-    If you want to install Composer inside your project root directory only, follow the instructions on [installing Composer in the current directory](https://getcomposer.org/download/).
+    If you want to install Composer inside your project root directory only, follow the instructions for [installing Composer in the current directory](https://getcomposer.org/download/).
 
     If you do so, you must replace `composer` with `php -d memory_limit=-1 composer.phar` in all commands below.
 


docs/getting_started/install_ibexa_dxp.md
```diff
@@ -13,7 +13,7 @@ description: Install Ibexa DXP on a Linux system and prepare your installation f
 
 !!! note "Installing [[= product_name_oss =]]"
 
-    This installation guide details the steps to install[[= product_name =]] for users who have a subscription agreement with [[= product_name_base =]].
+    This installation guide details the steps to install [[= product_name =]] for users who have a subscription agreement with [[= product_name_base =]].
```
juskora, line 16:
```suggestion
    This installation guide shows in details how to install [[= product_name =]] for users who have a subscription agreement with [[= product_name_base =]].
```

docs/getting_started/install_ibexa_dxp.md
```diff
@@ -51,12 +50,7 @@ composer -V
 
 !!! tip "Install Composer locally"
 
-    If you want to install Composer inside your project root directory only,
-    run the following command in the terminal:
-
-    ``` bash
-    php -r "readfile('https://getcomposer.org/installer');" | php
-    ```
+    If you want to install Composer inside your project root directory only, follow the instructions on [installing Composer in the current directory](https://getcomposer.org/download/).
```
juskora, line 53:
```suggestion
    If you want to install Composer inside your project root directory only, follow the instructions for [installing Composer in the current directory](https://getcomposer.org/download/).
```

juskora (approved):
@mnocon I know it's not the scope but maybe with this PR we could also get rid of collapsible section in line 134 as it brought some problems recently.

mnocon:
> @mnocon I know it's not the scope but maybe with this PR we could also get rid of collapsible section in line 134 as it brought some problems recently.

We are switching the default PHP version to 8.3 in the upcoming release (next week) and we will have to adjust this section anyway - that's why I didn't touch it in this PR (I agree that we should improve this)

julitafalcondusza (approved):

PR #2394, opened 2024-06-04
"new" pill in the ToC

adriendupuis:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | 
| Versions      | 
| Edition       | 

Each month, along side the "New in documentation" page, pages with changes can be highlighted in the table of content / navigation.

"new" doesn't necessarily mean that the page is new but that something is new in it. "Update" is too long for the little place there is in the ToC. To distinguish "new" and "updated" isn't relevant enough. Let's have only "new".

By having `month_change: true` in the page metadata (at the top of its .md file), a "New" pill is displayed at the right side of its ToC entry.

Because of this mechanism, it can't be added to the PHP API Reference which is directly in HTML and can't be converted to a Markdown trick.

Previews:

Look in ToC for May changes:
- API > PHP API > [PHP API](https://ez-systems-developer-documentation--2394.com.readthedocs.build/en/2394/api/php_api/php_api/)
- Administration > Back Office > [Recent activity](https://ez-systems-developer-documentation--2394.com.readthedocs.build/en/2394/administration/recent_activity/recent_activity/)

It works when the item is the current page, and when it isn't.

"new" on current
"new" elsewhere

#### Checklist

- [ ] Text renders correctly
- [ ] Text has been checked with vale
- [ ] Description metadata is up to date
- [ ] Redirects cover removed/moved pages
- [ ] Code samples are working
- [ ] PHP code samples have been fixed with PHP CS fixer
- [ ] Added link to this PR in relevant JIRA ticket or code PR

PR #2429, opened 2024-07-09
Mark higher-level nodes with "New" pill as well, not only the actual item

mnocon:
This PR builds upon https://github.com/ibexa/documentation-developer/pull/2394

Target: master, 4.6

When compiling the changelog I've noticed that while the New pill is great it's not so easy to notice when it's hidden deep in the tree. If you go to https://doc.ibexa.co/en/latest/ it's hard to find that we've added a page for the ContentName criterion. You also don't learn that Ibexa Engage is new unless you decide to expand this section.

This PR propagates the "New" pill to higher nodes - if there is a child node with a "New" label the parent node will display "New" as well.

Preview: https://ez-systems-developer-documentation--2429.com.readthedocs.build/en/2429/

I'm finding the visible pills -> then finding their parents in the DOM -> then finding the closest pill to each parent -> making them visible

My JS skills are very basic, maybe there's a better way to do this - all suggestions are welcome.

### DIFF
--- a/docs/js/custom.js
+++ b/docs/js/custom.js
@@ -204,5 +204,5 @@ $(document).ready(function() {
     });
 
     // Mark higher-level nodes with "New" pill, not only the actual item
-    $(".pill.new:not([hidden])").parents(".md-nav__item").uniqueSort().children('label').children(".pill.new[hidden]").removeAttr('hidden');
+    $(".pill.new:not([hidden])").parents(".md-nav__item").children('label').children(".pill.new[hidden]").removeAttr('hidden');
 });


docs/js/custom.js
```diff
@@ -202,4 +202,7 @@ $(document).ready(function() {
             closeBtn.closest('.notification').setAttribute('hidden', 'hidden');
         });
     });
+
+    // Mark higher-level nodes with "New" pill, not only the actual item
+    $(".pill.new:not([hidden])").parents(".md-nav__item").uniqueSort().children('label').children(".pill.new[hidden]").removeAttr('hidden')
```
dew326, line 207:
```suggestion
    $(".pill.new:not([hidden])").parents(".md-nav__item").uniqueSort().children('label').children(".pill.new[hidden]").removeAttr('hidden');
```

dew326 (approved):

julitafalcondusza (approved):

juskora (approved):

docs/js/custom.js
```diff
@@ -202,4 +202,7 @@ $(document).ready(function() {
             closeBtn.closest('.notification').setAttribute('hidden', 'hidden');
         });
     });
+
+    // Mark higher-level nodes with "New" pill, not only the actual item
+    $(".pill.new:not([hidden])").parents(".md-nav__item").uniqueSort().children('label').children(".pill.new[hidden]").removeAttr('hidden');
```
adriendupuis, line 207: I run the test, two new-pilled siblings doesn't bring twice their common parent, `uniqueSort` is useless.
```suggestion
    $(".pill.new:not([hidden])").parents(".md-nav__item").children('label').children(".pill.new[hidden]").removeAttr('hidden');
```

adriendupuis (approved):
A protection against duplicate seems useless but no performance issue if kept.
Nice trick to have the pill hidden on every node and reveal the one from new leafs' ancestors.

docs/js/custom.js
mnocon: You're right, this makes it a bit simpler to understand, thanks!

My initial solution was inserting new elements directly into the DOM, only in the second iteration I've changed to removing the attribute - and indeed it's not needed here

PR #2440, opened 2024-07-17
Added documentation for Location swap migration

mnocon:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | https://issues.ibexa.co/browse/IBX-8398
| Versions      | master, 4.6, 3.3 ‼️ 
| Edition       | all

Added documentation for https://github.com/ibexa/migrations/pull/383

Preview: https://ez-systems-developer-documentation--2440.com.readthedocs.build/en/2440/content_management/data_migration/importing_data/#available-migrations

I'm adding the new mode to the table. I've also noticed that some migration types are missing from it (`sql:execute`, `reference:load`, `reference:save`) - I will create a follow-up for them so that the table is exhaustive and mentions all the migrations we offer.

#### Checklist

- [x] Text renders correctly

### DIFF
--- a/code_samples/data_migration/examples/swap_location.yaml
+++ b/code_samples/data_migration/examples/swap_location.yaml
@@ -1,0 +1,9 @@
+-
+    type: location
+    mode: swap
+    match1:
+        field: location_remote_id
+        value: f3e90596361e31d496d4026eb624c983
+    match2:
+        field: location_id
+        value: 5
--- a/docs/content_management/data_migration/importing_data.md
+++ b/docs/content_management/data_migration/importing_data.md
@@ -49,31 +49,31 @@ Then, the step is described by additional properties depending on its type and m
 
 The following data migration step modes are available:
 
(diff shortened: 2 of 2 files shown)


Steveb-p (approved):
> I'm adding the new mode to the table. I've also noticed that some migration types are missing from it (sql:execute, reference:load, reference:save) - I will create a follow-up for them so that the table is exhaustive and mentions all the migrations we offer.

Issue with those is they are highly custom, just like `swap` is. It's something that doesn't make sense for a lot of other migration operations.

Ideally they'd be kept separate from this table, simply because we cannot create a new column for each and every custom type. I already have plans to introduce migrations steps that "allow some migration steps to fail" (basically a try...catch on execution).
And `repeatable` is also an example of a custom type.

adriendupuis (approved):
I like the idea of illustrating two different matches.

julitafalcondusza (approved):

PR #2450, opened 2024-07-31
IBX-8478: Repository-aware migration directory

dabrt:
| Question      | Answer |
| ------------- | --- 
| JIRA Ticket   | IBX-8478 |
| Versions      | master |
| Edition       | all |

Describe a new feature in the migrations realm.

#### Checklist

- [x] Text renders correctly
- [x] Text has been checked with vale
- [x] Added link to this PR in relevant JIRA ticket or code PR

PR #2441, opened 2024-07-17
Added documentation for env function in migrations

mnocon:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   |  https://issues.ibexa.co/browse/IBX-8483
| Versions      | master, 4.6
| Edition       | all

Documentation for https://github.com/ibexa/migrations/pull/395

Preview: https://ez-systems-developer-documentation--2441.com.readthedocs.build/en/2441/content_management/data_migration/importing_data/#built-in-functions

1) Added the new `env` function with an example
2) Changed from the Symfony container service name (which I don't think anybody cares about) to the function name for all the functions (reference, project_dir, env)

#### Checklist

- [x] Text renders correctly
- [x] Text has been checked with vale
- [x] Code samples are working
- [x] Added link to this PR in relevant JIRA ticket or code PR

PR #2453, opened 2024-08-01
Update changelog for July 2024

dabrt:
| Question      | Answer |
| ------------- | --- |
| JIRA Ticket   | N/A |
| Versions      | 4.6/master |
| Edition       | all |

#### Checklist

- [x] Text renders correctly

### DIFF
--- a/docs/content_management/data_migration/importing_data.md
+++ b/docs/content_management/data_migration/importing_data.md
@@ -1,6 +1,7 @@
 ---
 description: Import data into your Repository from prepared YAML files.
 page_type: reference
+month_change: true
 ---
 
 # Importing data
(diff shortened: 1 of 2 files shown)


mnocon (commented):
Thanks! I like the entry, I think it covers everything we did.

One thing: I was thinking about using the `New` pill to indicate the new content in the Migrations as well:
https://github.com/ibexa/documentation-developer/pull/2440
https://github.com/ibexa/documentation-developer/pull/2441
https://github.com/ibexa/documentation-developer/pull/2450

These are small changes, but basically all of them are new features that might be interesting to the readers (and we can use the pill to highlight updates as well)

mnocon (approved):
Thanks!

PR #2475, opened 2024-08-21
Added paragraph about siteaccess scopes are merged

vidarl:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | N/A
| Versions      | 3.3, 4.6
| Edition       | Content/Headless,

Added paragraph about settings for the different siteaccess scopes are merged

#### Checklist

- [ ] Text renders correctly

### DIFF
--- a/docs/multisite/multisite_configuration.md
+++ b/docs/multisite/multisite_configuration.md
@@ -147,7 +147,7 @@ ibexa:
           languages: ['ger-DE']
 ```
 
-The language setting for the `de` siteaccess will in this case be `['ger-DE','eng-GB']`, not `['de-de']`.
+The language setting for the `de` siteaccess will in this case be `['ger-DE','eng-GB']`, not `['ger-DE']`.
 
 ### SiteAccesses and Page Builder [[% include 'snippets/experience_badge.md' %]] [[% include 'snippets/commerce_badge.md' %]]
 


mnocon:
Thank you Vidar! So for scalar values (string, int) the "higher level" scope simply overwrites the lower levels - but in terms of arrays they get merged together?

alongosz (approved):

konradoboza (approved):

alongosz:
> > Thank you Vidar! So for scalar values (string, int) the "higher level" scope simply overwrites the lower levels - but in terms of arrays they get merged together?
> 
> That is correct

This is more complicated than that. It depends on specific set of settings and their nesting level. In case of languages this is true, but for each case I'd check the container. We had and probably still have some config merging issues in some places.

vidarl:
@alongosz already approved this one, so I assume it can be merged?

docs/multisite/multisite_configuration.md
```diff
@@ -135,6 +135,20 @@ ibexa:
...
+``` yaml
+ibexa:
+    system:
+        default:
+          languages: ['eng-GB']
+        de:
+          languages: ['ger-DE']
+```
+
+The language setting for the `de` siteaccess will in this case be `['ger-DE','eng-GB']`, not `['ger-DE']`.
```
mnocon, line 150:
````suggestion
The final value of the `languages` setting for the `de` SiteAccess is `['ger-DE','eng-GB']`, not `['ger-DE']`.
You can always check the final values by running the `ibexa:debug:config` console command. 
For this example, the command looks like this:
``` bash
php bin/console ibexa:debug:config --siteaccess=de languages
```
````

File: docs/multisite/multisite_configuration.md [hunk 2]
```diff
@@ -135,6 +135,20 @@ ibexa:
                             Identifier\ContentType: [article]
 ```
 
+Be aware of the fact that configurations in the different scopes are merged, not overwritten. This is important to have
+in mind when dealing with arrays:
```
mnocon, line 139:
```suggestion
When working with arrays, configurations from different scopes are merged together. See the following example:
```

mnocon (commented):
Thank you Vidar!

A couple suggestions from me to make it simpler - I'd also mention the `ibexa:debug:config` command to make sure people know how to check this themselves

docs/multisite/multisite_configuration.md
julitafalcondusza: Just a small remark: "See the following (...)" - new line.

PR #2541, opened 2024-11-13
main.py cards: get version from abs URL

adriendupuis:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | 
| Versions      | 
| Edition       | 

When generating card link, get the current version from the absolute URL instead of the canonical URL.

The canonical URL seems to always be the "latest" version. So, previously, version was always "latest" even when on another version (like "master" for example), potentially generating a broken link.

For example, 

https://doc.ibexa.co/en/master/users/users/ “Update basic user data from CLI” card is a broken link because the target page doesn't exists on latest, only on master.

#### Checklist

- [ ] Text renders correctly

### DIFF
--- a/main.py
+++ b/main.py
@@ -50,6 +50,9 @@ def define_env(env):
         canonical = current_page.canonical_url
         url_parts = re.search("//([^/]+)/([^/]+)/([^/]+)/", canonical)
         (site, language, version) = url_parts.groups()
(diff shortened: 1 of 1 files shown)


adriendupuis:
Wrong track

In #271, I already noticed that current version can't be obtained this way. See (#288) for other tries to fix this.

PR #2562, opened 2024-11-30
[CI] Added code samples testing with Deptrac

mnocon:
Target: 5.0 only, for new code only.

[Deptrac](https://github.com/qossmic/deptrac) is a code testing tool for architecture.

I've added a very simple ruleset - code samples should not use classes outside of the `Ibexa\Contracts` namespace.

Together with @konradoboza we've also added a few additional exceptions - like SiteAccess configuration classes etc.

This tool should act as a:
1) sanity check when adding code samples (as it's easy to miss the non-Contracts usage that can be replaced with their Contracts counterpart)
2) point for discussion with developers (is there a `Contracts-compliant` way of writing the code sample)

But we will never get rid of all the errors, as they would require architecture changes on the DXP side - so we will have to keep expanding the baseline with new code samples.

Generated baseline:
https://github.com/ibexa/documentation-developer/blob/add-deptrac/deptrac.baseline.yaml


code_samples/back_office/limitation/src/Controller/CustomController.php
```diff
@@ -0,0 +1,57 @@
...
+
+    private function getCustomLimitationValue(): bool
+    {
+        $hasAccess = $this->permissionResolver->hasAccess('custom_module', 'custom_function_2');
+
+        if (is_bool($hasAccess)) {
+            return $hasAccess;
+        }
+
+        $customLimitationValues = $this->permissionChecker->getRestrictions($hasAccess, CustomLimitationValue::class);
```
konradoboza: Nitpick:
```suggestion
        $customLimitationValues = $this->permissionChecker->getRestrictions(
            $hasAccess,
            CustomLimitationValue::class
        );
```

code_samples/back_office/limitation/src/Security/Limitation/Mapper/CustomLimitationValueMapper.php
```diff
@@ -9,8 +9,9 @@
 
 class CustomLimitationValueMapper implements LimitationValueMapperInterface
 {
-    public function mapLimitationValue(Limitation $limitation): bool
+    /** return array<bool> */
```
konradoboza:
```suggestion
    /** 
     * @return array<bool>
     */
```

code_samples/back_office/limitation/templates/themes/standard/limitation/custom_limitation_value.html.twig
```diff
@@ -1,4 +1,5 @@
 {# templates/themes/standard/limitation/custom_limitation_value.html.twig #}
-{% block ez_limitation_customlimitation_value %}
-    <span style="color: {{ values ? 'green' : 'red' }};">{{ values ? 'Yes' : 'No' }}</span>
+{% block ibexa_limitation_customlimitation_value %}
+    {% set isSet = values | first %}
```
konradoboza: We tend to use snake case for Twig variables:
```suggestion
    {% set is_set = values | first %}
```

konradoboza (approved):
Great work! 💪

Just a few nitpicks:

code_samples/back_office/limitation/src/Controller/CustomController.php
mnocon: All suggestions applied in https://github.com/ibexa/documentation-developer/commit/82e6849d0fc74e6fda1e120d4ac90bfefce1c634, thank you!

adriendupuis (approved):

PR #2566, opened 2024-12-02
IBX-4064: Search adjustements

adriendupuis:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | [IBX-4064](https://issues.ibexa.co/browse/IBX-4064)
| Versions      | All
| Edition       | All

- "See all results" → "See more results" Algolia is limited to 1000 results. As I didn't find a way to increase it, the need was discarded as nobody goes to the 1000th result. "See more results" is more adapted to this situation of not showing absolutely all results.
- Disable keyboard selection, no more result highlight moving on keyboard navigation with arrows. To navigate with keyboard doesn't work as "Enter" key is a shortcut to open the search page instead of the item selected with the keyboard, so, highlight was useless and misleading.

#### Checklist

- [ ] Text renders correctly

### DIFF
--- a/docs/js/custom.js
+++ b/docs/js/custom.js
@@ -144,6 +144,7 @@ $(document).ready(function() {
         debug: false,
     });
     search.autocomplete.on('autocomplete:updated', event => {
+        $('.ds-cursor').removeClass('ds-cursor');
         const searchedText = $('#search_input')[0].value.trim();
         const separatorText = '›';
         const separatorClass = 'aa-suggestion-title-separator';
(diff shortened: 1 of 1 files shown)


dabrt:
I seem to have an issue with "Show more results" link. Contacted Adrien directly

dabrt (approved):

mnocon (approved):

adriendupuis:
> I seem to have an issue with "Show more results" link. Contacted Adrien directly with a recording

The "See more results" link appears only when there is more than 10 results. If all the results are in the drop-down list, no need to link to a search result screen which pagination will only present one page.

But, it's true that seeing the link blinking while typing could be disturbing. It's worth thinking more about this (a placeholder, a delay, …)

PR #2567, opened 2024-12-03
REST: Unescape slashes in JSON examples

adriendupuis:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | 
| Versions      | 
| Edition       | 

JSON examples were mixing escaped and not escaped slashes. This PR unify them as not escaped.

[ECMA-404 "The JSON data interchange syntax" 2nd edition, December 2017](https://ecma-international.org/wp-content/uploads/ECMA-404_2nd_edition_december_2017.pdf) states:

> All code points may be placed within the quotation marks except for the code points that must be escaped: quotation mark (U+0022), reverse solidus (U+005C), and the control characters U+0000 to U+001F.
> […]
> `\/` represents the solidus character (U+002F).

So escaping the [slash (a.k.a solidus)](https://en.wikipedia.org/wiki/Slash_(punctuation)) is optional; it doesn't matter.

```
% echo '{ "slash": "\/" }' | jq                                                                                     
{
  "slash": "/"
}
```
`jq` understands it but considers it's not worth keeping it.

#### Checklist

- [ ] Text renders correctly


adriendupuis:
Too much outdated and something went wrong in rebasing

PR #2582, opened 2024-12-16
Cast contentId and locationId command arguments to integer

adriendupuis:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | N/A
| Versions      | 4.6, master
| Edition       | All

Fix `PHP Fatal error:  Uncaught TypeError: Ibexa\Core\Repository\SiteAccessAware\LocationService::loadLocation(): Argument #1 ($locationId) must be of type int, string given, called in /var/www/html/src/Command/BrowseLocationsCommand.php on line 45 and defined in /var/www/html/vendor/ibexa/core/src/lib/Repository/SiteAccessAware/LocationService.php:53`

I didn't test the others but casting every 'contentId' or 'locationId' to `int` is probably a good habit.

#### Checklist

- [ ] Text renders correctly

### DIFF
--- a/code_samples/api/public_php_api/src/Command/AddLocationToContentCommand.php
+++ b/code_samples/api/public_php_api/src/Command/AddLocationToContentCommand.php
@@ -45,8 +45,8 @@ class AddLocationToContentCommand extends Command
         $user = $this->userService->loadUserByLogin('admin');
         $this->permissionResolver->setCurrentUserReference($user);
 
-        $parentLocationId = $input->getArgument('parentLocationId');
-        $contentId = $input->getArgument('contentId');
+        $parentLocationId = (int) $input->getArgument('parentLocationId');
+        $contentId = (int) $input->getArgument('contentId');
 
         $locationCreateStruct = $this->locationService->newLocationCreateStruct($parentLocationId);
 
--- a/code_samples/api/public_php_api/src/Command/BookmarkCommand.php
+++ b/code_samples/api/public_php_api/src/Command/BookmarkCommand.php
@@ -34,7 +34,7 @@ class BookmarkCommand extends Command
 
     protected function execute(InputInterface $input, OutputInterface $output): int
     {
-        $locationId = $input->getArgument('locationId');
+        $locationId = (int) $input->getArgument('locationId');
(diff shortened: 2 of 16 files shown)


code_samples/api/public_php_api/src/Command/BrowseLocationsCommand.php
```diff
@@ -40,7 +40,7 @@ private function browseLocation(Location $location, OutputInterface $output, int
 
     protected function execute(InputInterface $input, OutputInterface $output): int
     {
-        $locationId = $input->getArgument('locationId');
+        $locationId = (int) $input->getArgument('locationId');
```
adriendupuis, line 43: @mnocon Why PHPStan didn't see an issue here?

https://github.com/symfony/symfony/blob/5.4/src/Symfony/Component/Console/Input/InputInterface.php#L83-L90 declares returning `mixed` but in fact it's probably only `int|array`.

mnocon: This one would be detected using level 9:
https://phpstan.org/user-guide/rule-levels

Bumping it results in:
```
  Line   api/public_php_api/src/Command/BrowseLocationsCommand.php                                                                     
 ------ ------------------------------------------------------------------------------------------------------------------------------ 
  :45    Parameter #1 $locationId of method Ibexa\Contracts\Core\Repository\LocationService::loadLocation() expects int, mixed given.  
 ------ ------------------------------------------------------------------------------------------------------------------------------ 
```

(and another 147 errors)

mnocon: We can think if we want to introduce this (level 10 is the highest right now), in product code we use level 8

mnocon (approved):

code_samples/api/public_php_api/src/Command/BrowseLocationsCommand.php
adriendupuis: @mnocon I don't fix 147 errors here 😛. I would say that we don't need it in CI. I'm curious about those level 9 new errors. But there is already a hundred to check in the actual level 8 phpstan-baseline.neon 😅

PR #2621, opened 2025-02-19
Fix segment groups accept headers

dabrt:
| Question      | Answer |
| ------------- | --- |
| JIRA Ticket   | none |
| Versions      | 4.6 and up |
| Edition       | all |

Fix segment groups accept headers

#### Checklist

- [x] Text renders correctly


mnocon (approved):
Thanks!

We can test tomorrow if the `Load segment group` and `List segment group's segments` deserve the same treatment (I suspect they do)

PR #2627, opened 2025-02-20
IBX-9609: Upgrade mkdocs

adriendupuis:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | [IBX-9609](https://issues.ibexa.co/browse/IBX-9609)
| Versions      | All
| Edition       | All

Upgrade `mkdocs`, python&pip, and all requirements.txt entries to their latest version.

- [MkDocs 1.5.1 → 1.6.1](https://www.mkdocs.org/about/release-notes/)
    - Test internal anchor link, report them as `INFO` during build or at server start up.
        - `INFO    -  Doc file 'templating/layout/add_navigation_menu.md' contains a link '../twig_function_reference/url_twig_functions.md#ibexa.url.alias', but the doc
           'templating/twig_function_reference/url_twig_functions.md' does not contain an anchor '#ibexa.url.alias'.` fixed in #2625
    - Remove NBSP when building the ToC
    - Prevent #2637 to happen
    - Different behavior with navigation duplicates (#2632)
    - Allow absolute links relative to `docs/` dir
- Material for MkDocs 9.1.1 → 9.6.5
    - Update CSS to compensate theme changes
- Pygments 2.15.0 → 2.19.1
    - Add JSX colorization ([2.17.0](https://pygments.org/docs/changelog/#version-2-17-0))
        - [Before](https://doc.ibexa.co/en/latest/administration/back_office/back_office_elements/custom_icons/#access-icons-in-javascript)
        - [After](https://ez-systems-developer-documentation--2630.com.readthedocs.build/en/2630/administration/back_office/back_office_elements/custom_icons/#access-icons-in-javascript)

Also applied in ibexa/documentation-user#334

## MkDocs Validation Feature

We previously used the default validation settings and I suggest we continue with defaults.

The strict mode is used when building the online doc which means that `warn` and `error` levels are blocking the deployment.

````yaml
validation:
  nav:
    omitted_files: info # The whole `snippets/` directory and `search_results.md` are genuinely absent from navigation.
    not_found: warn # If a file from the ToC is not found, deployment is stopped.
    absolute_links: info # In the ToC, a link starting with a slash is seen as external. `INFO    -  An absolute path to '/…' is included in the 'nav' configuration, which presumably points to an external resource.` I didn't manage the exact behavior as python crashes afterward. It could be increased to `error`
  links:
    not_found: warn # A broken link was already stopping deployment. It's important and can stay this way. `WARNING -  AutoLinksPlugin unable to find … in directory …` `WARNING -  Doc file '…' contains a link '…', but the target '…' is not found among documentation files.` 
    anchors: info # If an internal link target an unknown anchor, it will be only reported in build log. No need to stop deployment, the page is there. But we have to discipline ourself and look at the log as it can be confusing to land on a tall page and to have to find out why here and what should be read.
    absolute_links:  info # By default, a link starting with a slash is seen as external. It can be changed to the special value `relative_to_docs` so it would be recognized as below the `docs/` dir. As `mkdocs-autolinks-plugin` is used, it useless for Markdow files, but this could be very useful for PHP API reference and its HTML files!
    unrecognized_links: info # TODO
````
#### Checklist

- [ ] Text renders correctly

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -4,6 +4,9 @@ site_name: Developer Documentation
 repo_url: https://github.com/ibexa/documentation-developer
 site_url: https://doc.ibexa.co/en/latest/
 copyright: "Copyright 1999-2024 Ibexa AS and others"
+validation:
+    links:
+        absolute_links: relative_to_docs
 nav:
     - Ibexa Developer Documentation: index.md
     - Ibexa DXP editions:

### CODE REVIEW

File: requirements.txt [hunk 1]
```diff
@@ -1,9 +1,9 @@
...
-mkdocs-autolinks-plugin==0.7.0
+mkdocs==1.6.1
+Pygments==2.19.1
+pymdown-extensions==10.14.3
+Markdown==3.7
+mkdocs-material==9.6.5
+mkdocs-macros-plugin==1.3.7
+mkdocs-redirects==1.2.2
+mkdocs-autolinks-plugin==0.7.1
 Jinja2==3.1.5
```
adriendupuis, line 9: Jinja2 version is maintained by dependabot like in #2657

PR #2645, opened 2025-03-03
Replace ddev get with ddev add-on get

ghost:
ddev get triggers deprecation warning: Command "get" is deprecated, use 'ddev add-on get' instead

See https://github.com/ddev/ddev/pull/6482


mnocon (approved):
Thank you!

PR #2663, opened 2025-03-07
Config activity log as a post-installation step

adriendupuis:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | N/A
| Versions      | 4.6, master
| Edition       | All

At the end of installation process, suggest to configure activity log to limit its length.
I add "Configure recent activity log" section before the two other cron related sections because it can be critical to fill up DB with activity log while the other two are just feature enablements.

[Preview](https://ez-systems-developer-documentation--2663.com.readthedocs.build/en/2663/getting_started/install_ibexa_dxp/#post-installation-steps)

#### Checklist

- [ ] Text renders correctly

### DIFF
--- a/docs/getting_started/install_ibexa_dxp.md
+++ b/docs/getting_started/install_ibexa_dxp.md
@@ -417,6 +417,11 @@ Finally, remove the temporary file:
 
 `rm ezp_cron.txt`
 
+### Configure recent activity log
+
+Actions in a repository are logged in its database,
+you may want to [limit this log length by setting its automatic truncation](recent_activity.md#configuration-and-cronjob).
+
 ### Enable the Link manager
 
 To make use of the [Link Manager](url_management.md#enabling-automatic-url-validation).
--- a/docs/getting_started/install_ibexa_dxp.md
+++ b/docs/getting_started/install_ibexa_dxp.md
@@ -396,6 +396,11 @@ You should see the welcome page.
 
     See the [Security checklist](security_checklist.md) for a list of security-related issues you should take care of before going live with a project.
 
+### Configure recent activity log
+
+Actions in a repository are logged in its database,
+you may want to [limit this log length by enabling its automatic truncation](recent_activity.md#configuration-and-cronjob).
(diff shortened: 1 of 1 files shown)


Sztig (approved):

docs/getting_started/install_ibexa_dxp.md
```diff
@@ -396,6 +396,11 @@ You should see the welcome page.
 
     See the [Security checklist](security_checklist.md) for a list of security-related issues you should take care of before going live with a project.
 
+### Configure recent activity log
+
+Actions in a repository are logged in its database,
+you may want to [limit this log length by enabling its automatic truncation](recent_activity.md#configuration-and-cronjob).
```
dabrt, line 402: "you may want" gives readers an option, but you do not explain they might want to limit the length of the log. Is there  defult value that is too big? Does it cause memory issues?

Sztig: Maybe we need to change the wording a bit. By default we are not truncating the log at all. The biggest problem is when the log is long overdue (milliions of records set to be deleted) on truncate the command will eat up resources rendering the site unavailable  for a while. Imo if the command is implemented as it is right now, we should strongly recommend to have this cronjob set up by default.

dabrt: Since the advice is... performance related, I would move it to the following article: [Performance](https://doc.ibexa.co/en/latest/infrastructure_and_maintenance/performance/)
where we have more such pieces of advice.

And there, I would provide a longer explanation:
"Actions in a repository are logged in its database.
When there's a large number of records and they need to be processed, it may impact the productivity of your application.
To avoid this, you may want to limit the log's length by [enabling its automatic truncation](recent_activity.md#configuration-and-cronjob) after a total number of entries reaches a certain threshold."

I would also include a direct link to the "Performance" article in the installation procedure, I feel like it is missing there.

adriendupuis:
Done in #2664

PR #2693, opened 2025-04-04
Fixed link to SiteAccess-aware settings

mnocon:
- Ticket: [IBX-9578](https://issues.ibexa.co/browse/IBX-9578)
- Target: master, 4.6

The current link was not a valid target


dabrt (approved):

adriendupuis (approved):
Not fully satisfying but it's an improvement.

PR #2695, opened 2025-04-07
Fixed broken anchor

mnocon:
Target: master, 4.6

This feels very hacky to me - but on the other hand, we won't be changing the 4.6.0 release notes that often...

### DIFF
--- a/docs/release_notes/ibexa_dxp_v4.6.md
+++ b/docs/release_notes/ibexa_dxp_v4.6.md
@@ -1097,6 +1097,9 @@ The following view matchers have been introduced in [[= product_name =]] v4.6.0:
 ### Full changelog
 
 [[% include 'snippets/release_46.md' %]]
+
+To update your application, see the [update instructions](update_from_4.6.md).
+
 [[= release_note_entry_end() =]]
 
 </div>
--- a/docs/snippets/release_46.md
+++ b/docs/snippets/release_46.md
@@ -8,4 +8,6 @@ To learn more about all the included changes, see the full release change logs:
 - [[[= product_name_exp =]] [[= version =]]](https://github.com/ibexa/experience/releases/tag/[[= version =]])
 - [[[= product_name_com =]] [[= version =]]](https://github.com/ibexa/commerce/releases/tag/[[= version =]])
 
+[[% if version != 'v4.6.0' %]]
 To update your application, see the [update instructions](../update_and_migration/from_4.6/update_from_4.6.md#[[= version_to_anchor(version) =]]).
+[[% endif %]]


docs/release_notes/ibexa_dxp_v4.6.md
```diff
@@ -1097,6 +1097,9 @@ The following view matchers have been introduced in [[= product_name =]] v4.6.0:
 ### Full changelog
 
 [[% include 'snippets/release_46.md' %]]
+
+To update your application, see the [update instructions](update_from_4.6.md).
```
adriendupuis, line 1101: let's imagine we just published 4.6.1.
Some people might still be on 4.5.x.
`update_from_4.6.md` doesn't link to `update_from_4.5.md`.
Maybe it should be introduced here. (But maybe it should be introduced in `update_from_4.6.md` and not in `ibexa_dxp_v4.6.md`. For 4.4 to 4.6, the chain is again broken.)
If introduced here, it could be something like this:

```suggestion
To update your application, see the [4.6 update instructions](update_from_4.6.md).

To update your application from v4.5.x, see the [4.5 to 4.6 update instructions](update_from_4.5.md).
```

docs/snippets/release_46.md
```diff
@@ -8,4 +8,6 @@ To learn more about all the included changes, see the full release change logs:
 - [[[= product_name_exp =]] [[= version =]]](https://github.com/ibexa/experience/releases/tag/[[= version =]])
 - [[[= product_name_com =]] [[= version =]]](https://github.com/ibexa/commerce/releases/tag/[[= version =]])
 
+[[% if version != 'v4.6.0' %]]
 To update your application, see the [update instructions](../update_and_migration/from_4.6/update_from_4.6.md#[[= version_to_anchor(version) =]]).
```
adriendupuis, line 12: Is there something preventing autolink here?
```suggestion
To update your application, see the [update instructions](update_from_4.6.md#[[= version_to_anchor(version) =]]).
```

mnocon: Yes, it doesn't work and generates a lot of:
```
WARNING -  Doc file 'release_notes/ibexa_dxp_v4.6.md' contains a link 'update_from_4.6.md#v461', but the target 'release_notes/update_from_4.6.md' is not found
           among documentation files.
```

I think autolinkining doesn't work in snippets, somehow the context of "current file" is lost

docs/release_notes/ibexa_dxp_v4.6.md
mnocon: I lean toward adding it to `update_from_4.6.md` - there should be a sentence in the introduction saying that if you're not on 4.6 yet, you should look at the instructions for the older verions.

I'd do it in a follow-up, ok for you? I don't know how far back the chain will go 😅

docs/snippets/release_46.md
adriendupuis: OK. So the relative path has to stay.

docs/release_notes/ibexa_dxp_v4.6.md
adriendupuis: You're right, let wait for this to become needed, we'll know then how it really is.

adriendupuis (approved):

julitafalcondusza (approved):

PR #2727, opened 2025-05-05
Improved wording for TaxonomyEntry

mnocon:
Target: master, 4.6

As suggested on Slack - we need to make sure that the difference between TaxonomyEntry and TaxonomyEntry assignment is understood better by the Partners.


reithor (approved):
I like it !

PR #2734, opened 2025-05-08
Add missing return types

mnocon:
Target: master only

PHPStan errors on master:
```
  Line   api/rest_api/src/Rest/InputParser/GreetingInput.php                                                                                                                    
 ------ ----------------------------------------------------------------------------------------------------------------------------------------------------------------------- 
  :12    Method App\Rest\InputParser\GreetingInput::parse() has no return type specified.                                                                                       
         🪪  missingType.return                                                                                                                                                 
  :12    Return type mixed of method App\Rest\InputParser\GreetingInput::parse() is not covariant with return type mixed of method Ibexa\Contracts\Rest\Input\Parser::parse().  
 ------ ----------------------------------------------------------------------------------------------------------------------------------------------------------------------- 

 ------ -------------------------------------------------------------------------------------------------------------------------------- 
  Line   api/rest_api/src/Rest/ValueObjectVisitor/RestLocation.php                                                                       
 ------ -------------------------------------------------------------------------------------------------------------------------------- 
  :20    Return type mixed of method App\Rest\ValueObjectVisitor\RestLocation::visit() is not covariant with return type void of method  
         Ibexa\Rest\Server\Output\ValueObjectVisitor\RestLocation::visit().              
```

Parent definitions:
https://github.com/ibexa/rest/blob/main/src/contracts/Input/Parser.php#L20
https://github.com/ibexa/rest/blob/main/src/lib/Server/Output/ValueObjectVisitor/RestLocation.php#L27

### DIFF
--- a/code_samples/api/rest_api/src/Rest/InputParser/GreetingInput.php
+++ b/code_samples/api/rest_api/src/Rest/InputParser/GreetingInput.php
@@ -9,7 +9,7 @@ use Ibexa\Rest\Input\BaseParser;
 
 class GreetingInput extends BaseParser
 {
-    public function parse(array $data, ParsingDispatcher $parsingDispatcher): mixed
+    public function parse(array $data, ParsingDispatcher $parsingDispatcher): Greeting
     {
         if (!isset($data['Salutation'])) {
             throw new Exceptions\Parser("Missing or invalid 'Salutation' element for Greeting.");


code_samples/api/rest_api/src/Rest/InputParser/GreetingInput.php
```diff
@@ -9,7 +9,7 @@
 
 class GreetingInput extends BaseParser
 {
-    public function parse(array $data, ParsingDispatcher $parsingDispatcher)
+    public function parse(array $data, ParsingDispatcher $parsingDispatcher): mixed
```
ViniTou, line 12:
```suggestion
    public function parse(array $data, ParsingDispatcher $parsingDispatcher): Greeting
```
Technicaly, you can do it like that to improve on future typehinting. As we are using `mixed` in interface, it is completly safe to narrow type.

`mixed` is fine also, so it is up to you.

ViniTou (approved):

PR #2766, opened 2025-06-03
[5.0] Added missing return type

mnocon:
Target: 5.0

PHPStan is failing on 5.0:
```
~/Desktop/repos/ibexa/documentation-developer fix-5.0-samples *1 ❯ composer phpstan                                                                                                                        10:48:44
> phpstan analyse
Note: Using configuration file /Users/marek/Desktop/repos/ibexa/documentation-developer/phpstan.neon.
 223/223 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%

 ------ ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
  Line   back_office/limitation/src/Security/Limitation/CustomLimitationType.php
 ------ ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
  79     Return type mixed of method App\Security\Limitation\CustomLimitationType::valueSchema() is not covariant with return type array|int of method Ibexa\Contracts\Core\Limitation\Type::valueSchema().
 ------ ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

 [ERROR] Found 1 error

Script phpstan analyse handling the phpstan event returned with error code 1
```

I feel `never` is the right keyword to use here: https://www.php.net/manual/en/language.types.never.php

### DIFF
--- a/code_samples/back_office/limitation/src/Security/Limitation/CustomLimitationType.php
+++ b/code_samples/back_office/limitation/src/Security/Limitation/CustomLimitationType.php
@@ -76,7 +76,7 @@ class CustomLimitationType implements Type
         throw new NotImplementedException(__METHOD__);
     }
 
-    public function valueSchema()
+    public function valueSchema(): never
     {
         throw new NotImplementedException(__METHOD__);
     }


code_samples/back_office/limitation/src/Security/Limitation/CustomLimitationType.php
```diff
@@ -76,7 +76,7 @@ public function getCriterion(Limitation $value, UserReference $currentUser): Cri
         throw new NotImplementedException(__METHOD__);
     }
 
-    public function valueSchema()
+    public function valueSchema(): never
```
adamwojs, line 79: > I feel never is the right keyword to use here: https://www.php.net/manual/en/language.types.never.php

I would stick to interface here but I guess `never` will also work in this particular example.

tbialcz (approved):

PR #2798, opened 2025-06-27
PHP API Ref v5.0.0

ezrobot:
PHP API Reference update for v5.0.0


adriendupuis:
Composer failed to download Ibexa packages.

PR #2824, opened 2025-07-14
[Object state] Change ez_lock identifier to ibexa_lock

mnocon:
Done as part of https://issues.ibexa.co/browse/IBX-10254


docs/administration/back_office/add_user_setting.md
```diff
@@ -39,7 +39,7 @@ Register the setting group as a service:
 [[= include_file('code_samples/back_office/settings/config/custom_services.yaml', 6, 9) =]]
 ```
 
-The value of the setting is accessible with `ez_user_settings['unit']`.
+The value of the setting is accessible with `ibexa_user_settings['unit']`.
```
mnocon, line 42: See https://github.com/ibexa/admin-ui/blob/main/src/bundle/Resources/views/themes/admin/ui/form_fields.html.twig#L299

PR #2836, opened 2025-07-18
Added mention about OpenAPI

mnocon:


adriendupuis (commented):
I would also mention the `/api/ibexa/v2/doc` route to access (when in `dev` mode?) to a live REST API Reference

docs/api/rest_api/rest_api_usage/rest_api_usage.md
```diff
@@ -11,6 +11,19 @@ Every interaction with the repository than you can do from back office or by usi
...
 
+## OpenAPI support
+
+The REST API is built on top of [API Platform](https://api-platform.com/docs/symfony/) and meets the [OpenAPI](https://www.openapis.org/) standard.
+
+You can downlaod the OpenAPI specification from the [REST API Reference](/api/rest_api/rest_api_reference/rest_api_reference.html), or generate it for your project by running one of the command below:
+
+``` bash
+php bin/console ibexa:openapi --output=spec.json # JSON output
+php bin/console ibexa:openapi --yaml --output=spec.yaml # YAML output
```
adriendupuis: Re-reading this, I would use the [recommended name](https://spec.openapis.org/oas/v3.1.0.html#document-structure):

```suggestion
php bin/console ibexa:openapi --output=openapi.json # JSON output
php bin/console ibexa:openapi --yaml --output=openapi.yaml # YAML output
```

Side note, I have always redirect the output to a file instead of using the `--output` option or its `-o` shortcut. The option probably ease command embedding/wrapping (like when you need to pass it to a container).

docs/api/rest_api/rest_api_usage/rest_api_usage.md
```diff
@@ -11,6 +11,23 @@ Every interaction with the repository than you can do from back office or by usi
...
+``` bash
+php bin/console ibexa:openapi --output=openapi.json # JSON output
+php bin/console ibexa:openapi --yaml --output=openapi.yaml # YAML output
+```
+
+Use the specification file with [available OpenAPI tools](https://tools.openapis.org/) to work faster with the API, for example, by generating libraries and clients for the API.
+
+!!! info
+
+    In [Symfony's `dev` environment](environments.md), you can access a REST API reference generated for your project by visiting the `/api/ibexa/v2/doc` route in the browser.
```
adriendupuis, line 29: This is when debug is turned on, which is what dev env does. I'm not sure we need to enter into this detail. For me it's OK written like that.

https://github.com/ibexa/rest/blob/main/src/bundle/Resources/config/routing.yml#L16

adriendupuis (approved):

docs/api/rest_api/rest_api_usage/rest_api_usage.md
mnocon: I'd leave it as is as well 👍

PR #2874, opened 2025-08-12
Removed mention about enabling pagination

mnocon:
Target: v5 only

This parameter is removed on v5 (and enabled by default)

### DIFF
--- a/docs/update_and_migration/from_4.6/update_to_5.0.md
+++ b/docs/update_and_migration/from_4.6/update_to_5.0.md
@@ -1044,6 +1044,14 @@ If you are using GraphQL in your project, you can generate its schema by running
 php bin/console ibexa:graphql:generate-schema
 ```
 
+### Upgrade GraphQL usage
+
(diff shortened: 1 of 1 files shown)


docs/content_management/field_types/field_type_reference/relationlistfield.md
```diff
@@ -108,32 +108,7 @@ $validators = [
...
-        }
-      }
-    }
-    }
-  }
-}
-```
+### GraphQL integration
 
+This field type supports paginating the results when queried using [GraphQL](graphql.md).
```
adriendupuis: If it's always paginating, I would say:
```suggestion
This field type is paginating the results when queried using [GraphQL](graphql.md).
```

As it can have side effects when moving from not paginated to always paginated, I would also mention this in the update process: #2891

mnocon: You're right, thank you!

adriendupuis (approved):

PR #2905, opened 2025-09-19
Reworded section for updating LTS Updates

mnocon:
As reported on Slack, people unfamiliar with the LTS Updates need more context to understand this section.

This PR:
1) Adds badges to highlight that we're talking about LTS updates (that apply to specific editions)
2) Adds more links for people not familiar with given LTS Updates

Screenshot 2025-09-19 at 10 07 20

### DIFF
--- a/docs/update_and_migration/from_4.6/update_from_4.6.md
+++ b/docs/update_and_migration/from_4.6/update_from_4.6.md
@@ -664,5 +664,5 @@ To use the [latest features](ibexa_dxp_v4.6.md) added to them, update them separ
     If you're already using it, run the following command to get the latest version of this feature:
 
     ```bash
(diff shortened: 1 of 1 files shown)


docs/update_and_migration/from_4.6/update_from_4.6.md
```diff
@@ -438,7 +438,12 @@ To use the [latest features](ibexa_dxp_v4.6.md) added to them, update them separ
 
 === "Discounts"
 
-    Run the following command to get the latest version:
+    ### Discounts [[% include 'snippets/lts-update_badge.md' %]] [[% include 'snippets/commerce_badge.md' %]]
```
mnocon, line 441: The only point of this header is to have a place where I can attach the badges

PR #2944, opened 2025-11-04
Add note on professional localization

dabrt:
| Question      | Answer |
| ------------- | --- |
| JIRA Ticket   | N/A |
| Versions      | all |
| Edition       | all |

Add note on professional localization

#### Checklist

- [x] Text renders correctly


mnocon:
My first thought was to wait with merge, but imho we could merge it right now? Even contibutions done right now might become lost.

PR #2968, opened 2025-12-03
[CI] Applied Rector suggestions and fixed PHPStan

mnocon:
Target: 5.0 only

- First commit: applies Rector suggestions
- Second commit: add Rector comment ("borrowed" from  https://github.com/ibexa/core/pull/684 )

### DIFF
--- a/docs/administration/dashboard/php_api_dashboard_service.md
+++ b/docs/administration/dashboard/php_api_dashboard_service.md
@@ -25,7 +25,7 @@ Using the `admin` account, it loads the group members, logs each one in, creates
 First argument is the `Content ID` of the dashboard to copy.
 Following arguments are the Content IDs of the user groups.
 
-``` php hl_lines="63"
+``` php hl_lines="61"
 [[= include_file('code_samples/back_office/dashboard/src/Command/DashboardCommand.php') =]]
 ```
 


code_samples/multisite/siteaccess/Configuration.php
```diff
@@ -7,6 +7,9 @@
 
 class Configuration extends SiteAccessConfiguration
 {
+    /**
+     * @phpstan-return \Symfony\Component\Config\Definition\Builder\TreeBuilder<'array'> The tree builder
```
adriendupuis, line 11: Not sure the single quotes are needed here
```suggestion
     * @phpstan-return \Symfony\Component\Config\Definition\Builder\TreeBuilder<array> The tree builder
```
Shouldn't the array structure be discribed like in https://github.com/ibexa/share/blob/main/src/lib/Permission/Mapper/UsersWithPermissionInfoMapper.php#L92 ?

Not a blocker anyway.

adriendupuis (approved):
Ooh, nice https://www.php.net/manual/en/functions.first_class_callable_syntax.php

Good catch for the php_api_dashboard_service highlight issue

code_samples/multisite/siteaccess/Configuration.php
mnocon: I've tried this and PHPStan doesn't recognize it without the quotes.
I don't know why, I've expected it to work as well - maybe @alongosz could explain this 📖 

About the array shape: I'd keep it as is for now, we can always imrpvoe this later

PR #2978, opened 2025-12-12
Added redirects for LTS updates

mnocon:
The idea is simple - very often I encounter an error page when switching between doc versions for LTS updates.

I'd like to have this redirects for 4.6 (and the reverse in v5), so that switching to 4.6 version while being on https://doc.ibexa.co/en/5.0/ai_actions/configure_ai_actions/ doesn't result in 404, but results in a redirect to https://doc.ibexa.co/en/4.6/ai_actions/install_ai_actions/

Once approved I'll create a PR for v5

### DIFF
--- a/plugins.yml
+++ b/plugins.yml
@@ -556,3 +556,7 @@ plugins:
             'getting_started/ibexa_cloud_guide.md': 'ibexa_cloud/ibexa_cloud_guide.md'
             'getting_started/install_on_ibexa_cloud.md': 'ibexa_cloud/install_on_ibexa_cloud.md'
             'infrastructure_and_maintenance/clustering/ddev_and_ibexa_cloud.md': 'ibexa_cloud/ddev_and_ibexa_cloud.md'
+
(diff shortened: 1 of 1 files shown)


plugins.yml
```diff
@@ -556,3 +556,7 @@ plugins:
             'getting_started/ibexa_cloud_guide.md': 'ibexa_cloud/ibexa_cloud_guide.md'
             'getting_started/install_on_ibexa_cloud.md': 'ibexa_cloud/install_on_ibexa_cloud.md'
             'infrastructure_and_maintenance/clustering/ddev_and_ibexa_cloud.md': 'ibexa_cloud/ddev_and_ibexa_cloud.md'
+
```
dabrt:
```suggestion
```

dabrt (approved):

adriendupuis (approved):
Genius!

Maybe we should track down all renamed/moved pages which haven't be covered by a redirection. It's probably partially scriptable.

PR #2981, opened 2025-12-15
[Discounts] Added missing events

mnocon:
Adding missing DisableDiscounts events

Target: 4.6, 5.0


adriendupuis (approved):
It seems that https://ez-systems-developer-documentation--2981.com.readthedocs.build/en/2981/api/php_api/php_api_reference/namespaces/ibexa-contracts-discounts-event.html is now fully covered in this page (excluding `DiscountStructEventInterface` which shouldn't need to).

PR #2986, opened 2025-12-15
Removed Behat testing

mnocon:
We don't use it, it's triggering false positives when looking for "ez" inside the doc - it's time to remove this.

I can 100% say it's not being used right now, but it was worth giving it a try a couple of years ago 😄


dabrt (approved):

adriendupuis (approved):
I'm for CI but this was very outdated on many aspects.

PR #2991, opened 2025-12-16
Mentioned Taxonomy Suggestions in AI Actions

mnocon:
Added mentions about Taxonomy Suggestions in the AI section, after internal feedback that people were looking for it there as well.


adriendupuis (approved):
(The card should work properly on prod.)

PR #3004, opened 2026-01-13
[4.6 only] Added messenger installation instructions

mnocon:
Doc for https://github.com/ibexa/installer/pull/202/changes

In 4.6, messenger is installed only when installing discounts - and the database schema is not created at all.
Hence, the need for manual instruction method.

### DIFF
--- a/docs/discounts/configure_discounts.md
+++ b/docs/discounts/configure_discounts.md
@@ -42,15 +42,17 @@ The worker passes the messages to the handler, which then starts the re-indexing
 
 To run discount re-indexing in the background:
 
-1\. Make sure that the transport layer is [defined properly](background_tasks.md#configure-package) in [[= product_name_base =]] Messenger configuration.
+1\. Install and configure [Ibexa Messenger](background_tasks.md#installation).
 
-2\. Make sure that the [worker starts](background_tasks.md#start-worker) together with the application to watch the transport bus:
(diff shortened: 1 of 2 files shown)


konradoboza (approved):

julitafalcondusza (approved):

docs/discounts/configure_discounts.md
```diff
@@ -42,15 +42,17 @@ The worker passes the messages to the handler, which then starts the re-indexing
...
 
 ``` bash
 php bin/console messenger:consume ibexa.messenger.transport --bus=ibexa.messenger.bus
 ```
 
-3\. Use a scheduler of your choice, for example, [cron](https://en.wikipedia.org/wiki/Cron), to periodically run the following command:
+4\. Use a scheduler of your choice, for example, [cron](https://en.wikipedia.org/wiki/Cron), to periodically run the following command:
 
 ``` bash
 php bin/console ibexa:discounts:reindex
```
adriendupuis: Somehow I would expect a CRON line there, giving me an idea of the period this command should have. Every 5 minutes??
````suggestion
``` cron
*/5 * * * * cd [path-to-ibexa]; php bin/console ibexa:discounts:reindex --quiet --env=prod
````

adriendupuis (approved):
Good for me, just a side remark.

dabrt (approved):

mnocon:
Nice suggestion, thank you!

PR #3010, opened 2026-01-15
Improved TOC header navigation for headers nested in tabs

mnocon:
Target: 4.6, 5.0

Issue reported on Slack.

Disclaimer: code is heavily AI-generated, but I've reviewed and tested it dilligently - hope I didn't miss anything.

One test scenario:
go to https://ez-systems-developer-documentation--3010.com.readthedocs.build/en/3010/update_and_migration/from_4.6/update_from_4.6/ and click on the headers (in the right TOC) in the "LTS updates" section - the tab should be activated automatically.

### DIFF
--- a/docs/js/custom.js
+++ b/docs/js/custom.js
@@ -247,4 +247,46 @@ $(document).ready(function() {
 
     // Mark higher-level nodes with "New" pill, not only the actual item
     $('.pill--new:not([hidden])').parents('.md-nav__item').children('label').children('.pill--new[hidden]').removeAttr('hidden');
+
+    function activateTabForHash() {
+        var hash = document.location.hash;
+        if (!hash) return;
+
+        var target = document.getElementById(hash.substring(1));
+        if (!target) return;
+
+        var tabbedBlock = $(target).closest('.tabbed-block');
+        if (!tabbedBlock.length) return;
+
+        var container = tabbedBlock.parent();
+        var index = container.children('.tabbed-block').index(tabbedBlock);
+        var tabbedSet = container.parent();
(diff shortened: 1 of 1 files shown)


reithor (approved):
lgtm - tested on chrome and opera.

One thing though:
May be it's semantically not 100% correct to use tabs for LTS updates in https://doc.ibexa.co/en/5.0/update_and_migration/from_4.6/update_from_4.6/#lts-updates
Tabs are fine when having a group of very similar things (like `MySql|Postgres` `Headless|Experience|Commerce`). Probably it is better to avoid Tabs for LTS Updates ?

mnocon:
The idea I had is that it's easier to quickly check the tab for each LTS update you use - having it without tabs would mean that you read through every LTS update section even when not using it.

But it's something we can discuss within the team

dabrt:
The script still hiccups - when you try switching back and forth, the focus does not follow. For example, when you click Symbol attribute tab header and then Discounts, the focus ends up in "Form Builder performance fix..."
Another example: 
obraz

reithor:
> The script still hiccups - when you try switching back and forth, the focus does not follow. For example, when you click Symbol attribute tab header and then Discounts, the focus ends up in "Form Builder performance fix..." 

seems not related to the new script - current status is already not 100% accurate - https://doc.ibexa.co/en/5.0/update_and_migration/from_4.6/update_from_4.6/#__tabbed_9_1

adriendupuis (approved):
It seems to find the right tab for any section's heading

julitafalcondusza (approved):

dabrt (approved):

PR #3028, opened 2026-01-30
Updated Commons DAM example to work - User-agent is missing

vidarl:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | N/A
| Versions      | 4.6, 5.0
| Edition       | Headless, Experience, Commerce

An [`User-Agent` header](https://foundation.wikimedia.org/wiki/Policy:Wikimedia_Foundation_User-Agent_Policy) is required, or Commons will only reply with 403 access denied now.

Also, small improvement to view template:
image

vs

image

#### Checklist

- [ ] Text renders correctly

### DIFF
--- a/code_samples/back_office/images/src/Connector/Dam/Handler/WikimediaCommonsHandler.php
+++ b/code_samples/back_office/images/src/Connector/Dam/Handler/WikimediaCommonsHandler.php
@@ -14,6 +14,7 @@ use Ibexa\Contracts\Connector\Dam\Search\Query;
 
 class WikimediaCommonsHandler implements HandlerInterface
 {
+    private const USER_AGENT = 'Ibexa DXP Commons Dam Connector';
     public function search(Query $query, int $offset = 0, int $limit = 20): AssetSearchResult
     {
         $searchUrl = 'https://commons.wikimedia.org/w/api.php?action=query&list=search&format=json&srnamespace=6'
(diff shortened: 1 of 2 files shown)


code_samples/back_office/images/templates/themes/standard/commons_asset_view.html.twig
```diff
@@ -4,7 +4,7 @@
     {{ parent() }}
     <div>
         <a href="{{ asset.assetMetadata.page_url }}">Image</a>
-        {% if asset.assetMetadata.author %} by {{ asset.assetMetadata.author }}{% endif %}
+        {% if asset.assetMetadata.author %} by {{ asset.assetMetadata.author|striptags }}{% endif %}
```
vidarl, line 7: If we trust that the URL coming from Commons cannot contain bad things. This will display the author as link:
```suggestion
        {% if asset.assetMetadata.author %} by {{ asset.assetMetadata.author|raw }}{% endif %}
```

code_samples/back_office/images/src/Connector/Dam/Handler/WikimediaCommonsHandler.php
```diff
@@ -14,6 +14,7 @@
 
 class WikimediaCommonsHandler implements HandlerInterface
 {
+    private const USER_AGENT = 'Ibexa DXP Commons Dam Connector';
```
vidarl, line 17: Maybe you want to inject Product name somehow . Wasn't sure if that worked in code snippets so hardcoded "Ibexa DXP"...

code_samples/back_office/images/templates/themes/standard/commons_asset_view.html.twig
mnocon: It's a 3rd party site so we can never be 100% sure - and this template is just an example, this will be easy to modify.

I'd keep it as it (striptags instead of raw), better safe than sorry

mnocon (approved):
Thank you Vidar!

code_samples/back_office/images/templates/themes/standard/commons_asset_view.html.twig
adriendupuis: https://commons.wikimedia.org/wiki/Commons:Reusing_content_outside_Wikimedia and https://commons.wikimedia.org/wiki/Commons:Credit_line#Credit_lines_required_by_license doesn't seem to specifically require a link on the author, I guess the link to the image page is enough. To strip the author link shouldn't be an legal issue. In short, I'm OK with this change.

adriendupuis (approved):

PR #3029, opened 2026-01-30
IBX-9846: Describe Embeddings search API

dabrt:
| Question      | Answer |
| ------------- | --- |
| JIRA Ticket   | IBX-9846 |
| Versions      | 4.6, 5.0 |
| Edition       | All |

Describe Embeddings search API

#### Checklist

- [x] Text renders correctly
- [x] PHP code samples have been fixed with PHP CS fixer


docs/search/embeddings_reference/embeddings_reference.md
```diff
@@ -0,0 +1,45 @@
...
+
+- [`Ibexa\Contracts\Core\Search\Embedding\EmbeddingProviderResolverInterface`](/api/php_api/php_api_reference/classes/Ibexa-Contracts-Core-Search-Embedding-EmbeddingProviderResolverInterface.html): Resolves the provider for a given embedding configuration
+
+## Embedding fields
+
+- [`Ibexa\Contracts\Core\Search\FieldType\EmbeddingFieldFactory`](/api/php_api/php_api_reference/classes/Ibexa-Contracts-Core-Search-FieldType-EmbeddingFieldFactory.html): Creates dedicated search fields that store embedding vectors
+
+## Validation
+
+- [`Ibexa\Contracts\Core\Repository\Values\Content\QueryValidatorInterface`](/api/php_api/php_api_reference/classes/Ibexa-Contracts-Core-Repository-Values-Content-QueryValidatorInterface.html): Validates embedding queries and configurations are validated before reaching the search engine
```
mikadamczyk: It does not validate configuration. Maybe `Validates embedding queries before they reach the search engine`

docs/search/search_api.md
```diff
@@ -378,4 +358,147 @@ $query->aggregations[] = new IntegerRangeAggregation('range', 'person', 'age',
...
     In the example all values above (and including) 60 are included in the last range.
 
-See [Agrregation reference](aggregation_reference.md) for details of all available aggregations.
+See [Aggregation reference](aggregation_reference.md) for details of all available aggregations.
+
+## Search with embeddings
+
+Embeddings are numerical representations that capture the meaning of text, images, or other content.
+Embeddings are generated by AI by converting words or documents into lists of numbers, instead of treating them as plain text.
+Such lists, aka. vectors, can then be compared to find content with similar meaning.
```
mikadamczyk:
```suggestion
Such lists, a.k.a. vectors, can then be compared to find content with similar meaning.
```

docs/search/search_api.md
```diff
@@ -378,4 +358,147 @@ $query->aggregations[] = new IntegerRangeAggregation('range', 'person', 'age',
...
+
+Searching with embeddings enables matching content based on meaning rather than exact text matches.
+Instead of comparing keywords, the system compares vectors that represent the semantic meaning of content and the query input.
+
+!!! note "Taxonomy suggestions"
+
+    Embedding queries have been introduced primarily to support the [Taxonomy suggestions](taxonomy.md#taxonomy-suggestions) feature but you use them in other scenarios.
```
mikadamczyk:
```suggestion
    Embedding queries have been introduced primarily to support the [Taxonomy suggestions](taxonomy.md#taxonomy-suggestions) feature but you can use them in other scenarios.
```

docs/search/search_api.md
```diff
@@ -378,4 +358,147 @@ $query->aggregations[] = new IntegerRangeAggregation('range', 'person', 'age',
...
+    
+Searching with embeddings can be combined with traditional search criteria and filters, which allows the semantic search to be constrained by content type, location, permissions, or other search criteria.
```
mikadamczyk: a little bit misleading. `EmbeddingQuery` doesn’t support query criteria, only filter. Maybe `can be combined with filters (Criterion) ...`

docs/search/search_api.md
```diff
@@ -378,4 +358,147 @@ $query->aggregations[] = new IntegerRangeAggregation('range', 'person', 'age',
...
+
+An embedding query is represented by the `Ibexa\Contracts\Core\Repository\Values\Content\EmbeddingQuery` value object.
+The object encapsulates the vector to search for, along with configuration such as the embedding model and similarity threshold.
```
mikadamczyk: `EmbeddingQuery` has no model or threshold fields. It only has embedding, filter, limit/offset, aggregations, performCount.

docs/search/search_api.md
```diff
@@ -378,4 +358,147 @@ $query->aggregations[] = new IntegerRangeAggregation('range', 'person', 'age',
...
+
+### Use embedding queries in search
+
+Embedding queries are executed through the search API in the same way as other search requests.
+You build an `EmbeddingQuery` instance by using a builder and pass it to the search service.
+Embedding queries can also be combined with filters and search criteria to narrow down results, such as by content type, location, or permissions.
+
+``` php
+use Ibexa\Contracts\Core\Repository\Values\Content\EmbeddingQueryBuilder;
+use Ibexa\Contracts\Core\Repository\Values\Content\Embedding;
```
mikadamczyk: no such thing like `Ibexa\Contracts\Core\Repository\Values\Content\Embedding`, maybe you were thinking about `use Ibexa\Contracts\Core\Repository\Values\Content\Query\Embedding;`

But `Embedding` is abstract and expects `float[]`, so this snippet won’t compile
```suggestion
use Ibexa\Contracts\Taxonomy\Search\Query\Value\TaxonomyEmbedding
```

docs/search/search_api.md
```diff
@@ -378,4 +358,147 @@ $query->aggregations[] = new IntegerRangeAggregation('range', 'person', 'age',
...
+use Ibexa\Contracts\Core\Repository\Values\Content\Query\Aggregation;
+
+// Create an embedding object that represents the search input
+$embedding = new Embedding('Find content similar to this text');
```
mikadamczyk: This class accepts     
```
    /**
     * @param float[] $value
     */
    public function __construct(array $value)
    {
        $this->value = $value;
    }
```

docs/search/search_api.md
```diff
@@ -378,4 +358,147 @@ $query->aggregations[] = new IntegerRangeAggregation('range', 'person', 'age',
...
+### Use embedding queries in search
+
+Embedding queries are executed through the search API in the same way as other search requests.
+You build an `EmbeddingQuery` instance by using a builder and pass it to the search service.
+Embedding queries can also be combined with filters and search criteria to narrow down results, such as by content type, location, or permissions.
+
+``` php
+use Ibexa\Contracts\Core\Repository\Values\Content\EmbeddingQueryBuilder;
+use Ibexa\Contracts\Core\Repository\Values\Content\Embedding;
+use Ibexa\Contracts\Core\Repository\Values\Content\Query\Aggregation;
```
mikadamczyk: `Aggregation` is an interface, cannot be instantiated

docs/search/search_api.md
```diff
@@ -378,4 +358,147 @@ $query->aggregations[] = new IntegerRangeAggregation('range', 'person', 'age',
...
+// Create an embedding object that represents the search input
+$embedding = new Embedding('Find content similar to this text');
+
+// Build the embedding query by using the fluent builder
+$embeddingQuery = EmbeddingQueryBuilder::create()
+    ->withEmbedding($embedding)
+    ->setLimit(10)              // maximum number of results
+    ->setOffset(0)              // result offset for pagination
+    ->setPerformCount(true)     // optionally count total matching items
+    ->setAggregations([
```
mikadamczyk: we could remove aggregations from the snippet (focus on embeddings) or:
```
      ->setAggregations([
          new ContentTypeTermAggregation('count_by_type'),
      ])
```

docs/search/search_api.md
```diff
@@ -378,4 +358,150 @@ $query->aggregations[] = new IntegerRangeAggregation('range', 'person', 'age',
...
+
+The following components are used to build and validate embedding-based queries:
+
+- [EmbeddingQuery](/api/php_api/php_api_reference/classes/Ibexa-Contracts-Core-Repository-Values-Content-EmbeddingQuery.html):
+    Represents a semantic similarity search request.
+    It contains the input vector and optional search parameters such as such as filtering, pagination, aggregations, and result counting
+
+- [EmbeddingQueryBuilder](/api/php_api/php_api_reference/classes/Ibexa-Contracts-Core-Repository-Values-Content-EmbeddingQueryBuilder.html):
+    A fluent builder for constructing `EmbeddingQuery` instances.
+    It enforces required parameters and integrates embedding queries with the search query pipeline
```
mikadamczyk:
```suggestion
    It helps construct queries consistently and integrates embedding queries with the search query pipeline, but you must still provide the required embedding value
```

docs/search/search_api.md
```diff
@@ -378,4 +358,150 @@ $query->aggregations[] = new IntegerRangeAggregation('range', 'person', 'age',
...
+
+- [QueryValidatorInterface](/api/php_api/php_api_reference/classes/Ibexa-Contracts-Core-Repository-Values-Content-QueryValidatorInterface.html):
+    Validates embedding queries before they are passed to the search engine.
+    Implementations ensure that the embedding model exists and that vector dimensions match the configured embedding field
```
mikadamczyk:
```suggestion
    Validates embedding query structure before execution. 
    Provider/model configuration and vector compatibility are resolved at runtime by the configured embedding and search engine components
```

docs/search/embeddings_reference/embeddings_reference.md
```diff
@@ -0,0 +1,45 @@
...
+## EmbeddingQuery
+
+- [`Ibexa\Contracts\Core\Repository\Values\Content\EmbeddingQuery`](/api/php_api/php_api_reference/classes/Ibexa-Contracts-Core-Repository-Values-Content-EmbeddingQuery.html): Represents a semantic similarity search request.
+It encapsulates an [Embedding](#embedding) instance and supports filtering, pagination, aggregations, and result counting through the same API as standard content queries.
+Embedding queries do not support criteria, Sort Clauses, facet builders, or spellcheck
+
+## Embedding
+
+- [`Ibexa\Contracts\Core\Repository\Values\Content\Query\Embedding`](/api/php_api/php_api_reference/classes/Ibexa-Contracts-Core-Repository-Values-Content-Query-Embedding.html): Represents the semantic input used for similarity search.
+Depending on the embedding provider, it can encapsulate text or vector data
```
mikadamczyk:
```suggestion
- [`Ibexa\Contracts\Core\Repository\Values\Content\Query\Embedding`](/api/php_api/php_api_reference/classes/Ibexa-Contracts-Core-Repository-Values-Content-Query-Embedding.html): Represents the vector input used for similarity search. 
It stores embedding values as float arrays, while providers generate those vectors from text input
```

docs/search/embeddings_reference/embeddings_reference.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Embedding providers
+
+Embedding providers generate vector representations for inputs.
```
mikadamczyk:
```suggestion
Embedding providers generate vector representations for inputs.
Out of the box, embedding search integration is provided for TaxonomyEmbedding.
If you use a custom embedding value type, implement matching embedding visitors for your search engine (Solr/Elasticsearch).
Otherwise, query execution may fail with "No visitor available".
```

docs/search/search_api.md
```diff
@@ -378,4 +358,150 @@ $query->aggregations[] = new IntegerRangeAggregation('range', 'person', 'age',
...
+// Create an Embedding instance with a float[] vector
+$embedding = new Embedding($vector);
+
+// Build the embedding query with the fluent builder
+$embeddingQuery = EmbeddingQueryBuilder::create()
+    ->withEmbedding($embedding)
+    ->setLimit(10)          // maximum number of results
+    ->setOffset(0)          // result offset for pagination
+    ->setPerformCount(true) // optionally count total matching items
+    ->build();
```
mikadamczyk:
```suggestion
use Ibexa\Contracts\Core\Repository\Values\Content\EmbeddingQueryBuilder;
use Ibexa\Contracts\Core\Repository\Values\Content\Query\Criterion\ContentTypeIdentifier;
use Ibexa\Contracts\Taxonomy\Search\Query\Value\TaxonomyEmbedding;

$vector = [
  0.0123,
  -0.9876,
  0.4567,
  ...
];

$embedding = new TaxonomyEmbedding($vector);

$embeddingQuery = EmbeddingQueryBuilder::create()
  ->withEmbedding($embedding)
  ->setFilter(new ContentTypeIdentifier('article'))
  ->setLimit(10)
  ->setOffset(0)
  ->setPerformCount(true)
  ->build();
```

docs/search/search_api.md
```diff
@@ -378,4 +358,150 @@ $query->aggregations[] = new IntegerRangeAggregation('range', 'person', 'age',
...
+
+// Execute the query via the repository
+$results = $repository->findContent($embeddingQuery);
```
mikadamczyk:
```suggestion
use Ibexa\Contracts\Core\Repository\Repository;
use Ibexa\Contracts\Core\Repository\SearchService;
use Ibexa\Contracts\Core\Repository\Values\Content\EmbeddingQueryBuilder;
use Ibexa\Contracts\Taxonomy\Search\Query\Value\TaxonomyEmbedding;

final class ExampleService
  {
      private SearchService $searchService;

      public function __construct(Repository $repository)
      {
          $this->searchService = $repository->getSearchService();
      }

      public function searchByEmbedding(array $vector): void
      {
          $query = EmbeddingQueryBuilder::create()
              ->withEmbedding(new TaxonomyEmbedding($vector))
              ->setLimit(10)
              ->setOffset(0)
              ->build();

          $result = $this->searchService->findContent($query);

          foreach ($result->searchHits as $hit) {
              // ...
          }
      }
  }

```

docs/search/search_api.md
```diff
@@ -18,7 +19,7 @@ The service should be [injected into the constructor of your command or controll
 
     `SearchService` is also used in the back office of [[= product_name =]], in components such as Universal Discovery Widget or Sub-items List.
 
-### Performing a search
+### Perform a search
```
dabrt:
```suggestion
### Perform search
```

docs/search/search_api.md
```diff
@@ -378,4 +358,150 @@ $query->aggregations[] = new IntegerRangeAggregation('range', 'person', 'age',
...
+Embeddings are generated by AI by converting words or documents into lists of numbers, instead of treating them as plain text.
+Such lists, aka vectors, can then be compared to find content with similar meaning.
+
+Searching with embeddings enables matching content based on meaning rather than exact text matches.
+Instead of comparing keywords, the system compares vectors that represent the semantic meaning of content and the query input.
+
+!!! note "Taxonomy suggestions"
+
+    Embedding queries have been introduced primarily to support the [Taxonomy suggestions](taxonomy.md#taxonomy-suggestions) feature but you can use them in other scenarios.
+
```
dabrt:
```suggestion

```

mikadamczyk (approved):

code_samples/api/public_php_api/src/Command/FindByTaxonomyEmbeddingCommand.php
```diff
@@ -0,0 +1,77 @@
+<?php
+
+/**
```
mnocon: Please remove the copyright, this code sample is meant to be used by other freely

docs/search/search_api.md
```diff
@@ -378,4 +358,147 @@ $query->aggregations[] = new IntegerRangeAggregation('range', 'person', 'age',
...
+Such lists, aka vectors, can then be compared to find content with similar meaning.
+
+Searching with embeddings enables matching content based on meaning rather than exact text matches.
+Instead of comparing keywords, the system compares vectors that represent the semantic meaning of content and the query input.
+
+!!! note "Taxonomy suggestions"
+
+    Embedding queries have been introduced primarily to support the [Taxonomy suggestions](taxonomy.md#taxonomy-suggestions) feature, therefore embedding search integration is provided for `TaxonomyEmbedding`.
+    To use them in other scenarios that use a custom embedding value type, implement matching embedding
+visitors for your search engine.
```
mnocon: Wrong formatting

Image

code_samples/api/public_php_api/src/Command/FindByTaxonomyEmbeddingCommand.php
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    protected function execute(
+        InputInterface $input,
+        OutputInterface $output
+    ): int {
+        $io = new SymfonyStyle($input, $output);
+
+        // Example embedding vector.
+        // In a real-life scenario, generate it with an embedding provider
```
mnocon: We should show people how to do this.

Can we just inject embeddingProviderResolver here, and do:

```
            $embeddingProvider = $this->embeddingProviderResolver->resolve();
            $embeddings = $embeddingProvider->getEmbedding('example_content');
```

(and `private EmbeddingProviderResolverInterface $embeddingProviderResolver`) in the constructor

code_samples/api/public_php_api/src/Service/TaxonomyEmbeddingSearchService.php
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    /**
+     * @param float[] $vector
+     *
+     * @return SearchResult<Content>
+     */
+    public function searchByEmbedding(array $vector): SearchResult
+    {
+        $query = EmbeddingQueryBuilder::create()
```
mnocon: The only difference is the lack of content type filter?
I'd combine these two into one sample, to make the maintenence easier for us in the future

docs/search/embeddings_reference/embeddings_reference.md
```diff
@@ -0,0 +1,50 @@
...
+# Embeddings search reference
+
+Embeddings provide vector representations of content or text, enabling semantic similarity search.
+Foundational abstractions are provided for embedding-based search, while embedding providers generate vector representations.
+
+## EmbeddingQuery
+
+- [`Ibexa\Contracts\Core\Repository\Values\Content\EmbeddingQuery`](/api/php_api/php_api_reference/classes/Ibexa-Contracts-Core-Repository-Values-Content-EmbeddingQuery.html): Represents a semantic similarity search request.
+It encapsulates an [Embedding](#embedding) instance and supports filtering, pagination, aggregations, and result counting through the same API as standard content queries.
+Embedding queries do not support criteria, Sort Clauses, facet builders, or spellcheck
```
mnocon: I'd mention the search engines where they are available.

Solr 9, ES, but not Legacy Search?

docs/search/embeddings_reference/embeddings_reference.md
```diff
@@ -0,0 +1,50 @@
...
+visitors for your search engine (Solr/Elasticsearch).
+Otherwise, query execution may fail with "No visitor available".
+
+### Provider contracts
+
+- [`Ibexa\Contracts\Core\Search\Embedding\EmbeddingProviderInterface`](/api/php_api/php_api_reference/classes/Ibexa-Contracts-Core-Search-Embedding-EmbeddingProviderInterface.html): Generates embeddings
+
+- [`Ibexa\Contracts\Core\Search\Embedding\EmbeddingProviderRegistryInterface`](/api/php_api/php_api_reference/classes/Ibexa-Contracts-Core-Search-Embedding-EmbeddingProviderRegistryInterface.html): Lists available embedding providers
+
+- [`Ibexa\Contracts\Core\Search\Embedding\EmbeddingProviderResolverInterface`](/api/php_api/php_api_reference/classes/Ibexa-Contracts-Core-Search-Embedding-EmbeddingProviderResolverInterface.html): Resolves the provider for a given embedding configuration
```
mnocon: Given how? Looking at https://github.com/ibexa/core/blob/d01b203204e4f9bc3bdaced92ade82eff0afbdd3/src/lib/Search/Embedding/EmbeddingProviderResolver.php#L17 , I'd say it's for the default configuration.

```suggestion
- [`Ibexa\Contracts\Core\Search\Embedding\EmbeddingProviderResolverInterface`](/api/php_api/php_api_reference/classes/Ibexa-Contracts-Core-Search-Embedding-EmbeddingProviderResolverInterface.html): Resolves the embedding provider based on system configuration
```

You can also mention the `default_embedding_model` configuration property here, as this is what's returned.

docs/search/search_api.md
```diff
@@ -378,4 +358,147 @@ $query->aggregations[] = new IntegerRangeAggregation('range', 'person', 'age',
...
+
+Searching with embeddings can be combined with filtering, which allows the semantic search results to be constrained by content type, location, permissions, or other criteria.
```
mnocon:
```suggestion
You can combine searching with embeddings with filtering, which allows the semantic search results to be constrained by content type, location, permissions, or other criteria.
```

To avoid the passive voice?
How does one filter by permissions, I don't think we support that?
It would be useful to link to the criteria that can be used here.


We're using the `findContent` method, and mention content search exclusively - can the query be used with location search as well? These would allow to constraint by location

docs/search/search_api.md
```diff
@@ -378,4 +358,147 @@ $query->aggregations[] = new IntegerRangeAggregation('range', 'person', 'age',
...
+
+An embedding query is represented by the `Ibexa\Contracts\Core\Repository\Values\Content\EmbeddingQuery` value object.
```
mnocon: PHP API link would be nice.

docs/search/embeddings_reference/embeddings_reference.md
mnocon: Faceted searched is removed in v5, and deprecated in v4.6

Let's replace it with aggregations.

mnocon: > Embedding queries do not support criteria

What does it mean? We say that it supports filtering, and filtering is defined with search criteria?

docs/search/search_api.md
```diff
@@ -378,4 +358,147 @@ $query->aggregations[] = new IntegerRangeAggregation('range', 'person', 'age',
...
+
+- [QueryValidatorInterface](/api/php_api/php_api_reference/classes/Ibexa-Contracts-Core-Repository-Values-Content-QueryValidatorInterface.html):
+    Validates embedding query structure before execution.
+    The provider/model configuration and vector compatibility are resolved at runtime by the configured embedding and search engine component
+
+### Use embedding queries in search
+
+Embedding queries are executed through the search API in the same way as other search requests.
+You build an `EmbeddingQuery` instance by using a builder and pass it to the search service.
+Embedding queries can also be combined with filters to narrow down results, such as by content type, location, or permissions.
```
mnocon: How to filter by permissions?

docs/search/search_api.md
```diff
@@ -378,4 +358,147 @@ $query->aggregations[] = new IntegerRangeAggregation('range', 'person', 'age',
...
+
+This example shows a minimal embedding query executed directly through the search service:
+
+``` php
+// ...
```
mnocon: Let's just embed the full Command here - it's easier for us to maintain, and the result is less confusing

Because it's hard to say what's happening here:
Image

docs/search/search_api.md
```diff
@@ -378,4 +358,147 @@ $query->aggregations[] = new IntegerRangeAggregation('range', 'person', 'age',
...
+[[= include_file('code_samples/api/public_php_api/src/Service/TaxonomyEmbeddingSearchService.php', 16, 18) =]]
+// ...
+[[= include_file('code_samples/api/public_php_api/src/Service/TaxonomyEmbeddingSearchService.php', 25, 36) =]]
+```
+
+The `EmbeddingQueryBuilder` ensures that the query is correctly configured before execution.
+
+!!! note "Embedding query properties"
+
+    Embedding queries do not allow standard Query properties such as `query`, `sortClauses`, `facetBuilders`, or `spellcheck`.
```
mnocon:
```suggestion
    Embedding queries do not allow standard Query properties such as `query`, `sortClauses`, or `spellcheck`.
```

docs/search/search_api.md
```diff
@@ -378,4 +358,147 @@ $query->aggregations[] = new IntegerRangeAggregation('range', 'person', 'age',
...
+
+An embedding query is represented by the `Ibexa\Contracts\Core\Repository\Values\Content\EmbeddingQuery` value object.
+The object encapsulates the embedding used for similarity search and optional search parameters such as filtering, pagination, aggregations, and result counting.
+The query is validated before being executed to ensure that the embedding configuration is consistent with the system setup.
+
+The following components are used to build and validate embedding-based queries:
+
+- [EmbeddingQuery](/api/php_api/php_api_reference/classes/Ibexa-Contracts-Core-Repository-Values-Content-EmbeddingQuery.html):
+    Represents a semantic similarity search request.
+    It contains the input vector and optional search parameters such as such as filtering, pagination, aggregations, and result counting
```
mnocon:
```suggestion
    It contains the input vector and optional search parameters such as filtering, pagination, aggregations, and result counting
```

docs/search/search_api.md
```diff
@@ -378,4 +358,147 @@ $query->aggregations[] = new IntegerRangeAggregation('range', 'person', 'age',
...
+          name: 'text-embedding-3-small'
+          dimensions: 1536
+          field_suffix: '3small'
+          embedding_provider: 'ibexa_openai'
+```
+
+For a real-life example of embedding configuration, see [Taxonomy suggestions](taxonomy.md#change-the-embedding-generation-model).
+
+Embedding providers implement the contract for generating vector representations of input data.
+At runtime, the system resolves right provider and assigns embedding generation to it.
```
mnocon:
```suggestion
At runtime, the system resolves the default provider responsible for generating embeddings.
```
> assigns embedding generation to it

just sounds too complicated to me.

Also, you can use the default provider, or request a specific provider (by using https://github.com/ibexa/core/blob/v5.0.6/src/lib/Search/Embedding/EmbeddingProviderResolver.php#L44 )

docs/search/search_api.md
```diff
@@ -378,4 +358,147 @@ $query->aggregations[] = new IntegerRangeAggregation('range', 'person', 'age',
     `null` means that a range doesn't have an end.
     In the example all values above (and including) 60 are included in the last range.
 
-See [Agrregation reference](aggregation_reference.md) for details of all available aggregations.
+See [Aggregation reference](aggregation_reference.md) for details of all available aggregations.
+
+## Search with embeddings
```
mnocon, line 363: As with Aggregation, let's mention that it's Solr/ES only

docs/search/search_api.md
```diff
@@ -378,4 +358,147 @@ $query->aggregations[] = new IntegerRangeAggregation('range', 'person', 'age',
...
+``` php
+// ...
+[[= include_file('code_samples/api/public_php_api/src/Command/FindByTaxonomyEmbeddingCommand.php', 10, 21) =]]
+// ...
+[[= include_file('code_samples/api/public_php_api/src/Command/FindByTaxonomyEmbeddingCommand.php', 26, 28) =]]
+// ...
+[[= include_file('code_samples/api/public_php_api/src/Command/FindByTaxonomyEmbeddingCommand.php', 33, 78) =]]
+```
+
+This example shows how to encapsulate an embedding query inside a reusable service class:
```
mnocon: I'd just remove this example, the reader will know that they can wrap this logic with their own helper.

docs/search/search_api.md
```diff
@@ -378,4 +358,147 @@ $query->aggregations[] = new IntegerRangeAggregation('range', 'person', 'age',
...
+``` yaml
+ibexa:
+  system:
+    default:
+      embedding_models:
+        text-embedding-3-small:
+          name: 'text-embedding-3-small'
+          dimensions: 1536
+          field_suffix: '3small'
+          embedding_provider: 'ibexa_openai'
```
mnocon: Please remember that for 4.6, connector-ai is an LTS update (optional) - and this info needs to be added.

docs/search/search_api.md
```diff
@@ -378,4 +358,147 @@ $query->aggregations[] = new IntegerRangeAggregation('range', 'person', 'age',
...
+These fields are then used by the search engine to perform vector similarity comparisons when embedding queries are executed.
+
+``` php
+use Ibexa\Contracts\Core\Search\FieldType\EmbeddingFieldFactory;
+use Ibexa\Contracts\Core\Search\Embedding\EmbeddingConfigurationInterface;
+
+// $config is an existing EmbeddingConfigurationInterface
+$factory = new EmbeddingFieldFactory($config);
+
+// Create a default embedding field (type derived from config suffix)
```
mnocon:
```suggestion
// Create a default embedding field (type derived from configuration's field suffix)
```

docs/search/search_api.md
```diff
@@ -378,4 +358,147 @@ $query->aggregations[] = new IntegerRangeAggregation('range', 'person', 'age',
...
+
+The following components are used to build and validate embedding-based queries:
+
+- [EmbeddingQuery](/api/php_api/php_api_reference/classes/Ibexa-Contracts-Core-Repository-Values-Content-EmbeddingQuery.html):
+    Represents a semantic similarity search request.
+    It contains the input vector and optional search parameters such as such as filtering, pagination, aggregations, and result counting
+
+- [EmbeddingQueryBuilder](/api/php_api/php_api_reference/classes/Ibexa-Contracts-Core-Repository-Values-Content-EmbeddingQueryBuilder.html):
+    A fluent builder for constructing `EmbeddingQuery` instances.
+    It helps construct queries consistently and integrates embedding queries with the search query pipeline, but you must still provide the required embedding value
```
mnocon:
```suggestion
    It helps construct queries consistently and integrates embedding queries with the search query pipeline.
    You must provide the required embedding value by using the `withEmbedding` method.
```

docs/search/embeddings_reference/embeddings_reference.md
```diff
@@ -0,0 +1,50 @@
...
+
+Embedding providers generate vector representations for inputs.
+Out of the box, embedding search integration is provided for TaxonomyEmbedding.
+If you use a custom embedding value type, implement matching embedding
+visitors for your search engine (Solr/Elasticsearch).
+Otherwise, query execution may fail with "No visitor available".
+
+### Provider contracts
+
+- [`Ibexa\Contracts\Core\Search\Embedding\EmbeddingProviderInterface`](/api/php_api/php_api_reference/classes/Ibexa-Contracts-Core-Search-Embedding-EmbeddingProviderInterface.html): Generates embeddings
```
mnocon:
```suggestion
- [`Ibexa\Contracts\Core\Search\Embedding\EmbeddingProviderInterface`](/api/php_api/php_api_reference/classes/Ibexa-Contracts-Core-Search-Embedding-EmbeddingProviderInterface.html): Generates embeddings for the provided text
```

docs/search/search_api.md
```diff
@@ -378,4 +358,147 @@ $query->aggregations[] = new IntegerRangeAggregation('range', 'person', 'age',
...
+          field_suffix: '3small'
+          embedding_provider: 'ibexa_openai'
+```
+
+For a real-life example of embedding configuration, see [Taxonomy suggestions](taxonomy.md#change-the-embedding-generation-model).
+
+Embedding providers implement the contract for generating vector representations of input data.
+At runtime, the system resolves right provider and assigns embedding generation to it.
+
+- [EmbeddingConfigurationInterface](/api/php_api/php_api_reference/classes/Ibexa-Contracts-Core-Search-Embedding-EmbeddingConfigurationInterface.html) defines how embedding models are configured in the system (model name, vector dimensionality, provider reference, field suffix)
```
mnocon:
```suggestion
- [EmbeddingConfigurationInterface](/api/php_api/php_api_reference/classes/Ibexa-Contracts-Core-Search-Embedding-EmbeddingConfigurationInterface.html) represents the embedding models configuration in the system (model name, vector dimensionality, provider reference, field suffix)
```

The configuration is specified (defined) in YAML, this class "just" allows to read it.

docs/search/search_api.md
```diff
@@ -378,4 +358,147 @@ $query->aggregations[] = new IntegerRangeAggregation('range', 'person', 'age',
     `null` means that a range doesn't have an end.
     In the example all values above (and including) 60 are included in the last range.
 
-See [Agrregation reference](aggregation_reference.md) for details of all available aggregations.
+See [Aggregation reference](aggregation_reference.md) for details of all available aggregations.
+
+## Search with embeddings
+
+Embeddings are numerical representations that capture the meaning of text, images, or other content.
+Embeddings are generated by AI by converting words or documents into lists of numbers, instead of treating them as plain text.
```
mnocon:
```suggestion
Embeddings are generated by converting words or documents into lists of numbers, instead of treating them as plain text.
```
Or "AI providers generate embeddings by..."

docs/search/search_api.md
```diff
@@ -378,4 +358,147 @@ $query->aggregations[] = new IntegerRangeAggregation('range', 'person', 'age',
...
+    Represents a semantic similarity search request.
+    It contains the input vector and optional search parameters such as such as filtering, pagination, aggregations, and result counting
+
+- [EmbeddingQueryBuilder](/api/php_api/php_api_reference/classes/Ibexa-Contracts-Core-Repository-Values-Content-EmbeddingQueryBuilder.html):
+    A fluent builder for constructing `EmbeddingQuery` instances.
+    It helps construct queries consistently and integrates embedding queries with the search query pipeline, but you must still provide the required embedding value
+
+- [QueryValidatorInterface](/api/php_api/php_api_reference/classes/Ibexa-Contracts-Core-Repository-Values-Content-QueryValidatorInterface.html):
+    Validates embedding query structure before execution.
+    The provider/model configuration and vector compatibility are resolved at runtime by the configured embedding and search engine component
```
mnocon: How is that related to the QueryValidator?

docs/search/search_api.md
```diff
@@ -378,4 +358,147 @@ $query->aggregations[] = new IntegerRangeAggregation('range', 'person', 'age',
...
+- [EmbeddingProviderRegistryInterface](/api/php_api/php_api_reference/classes/Ibexa-Contracts-Core-Search-Embedding-EmbeddingProviderRegistryInterface.html) lists all available embedding providers
+
+- [EmbeddingProviderResolverInterface](/api/php_api/php_api_reference/classes/Ibexa-Contracts-Core-Search-Embedding-EmbeddingProviderResolverInterface.html) determines which provider should be used for a given embedding configuration
+
+### Embedding fields
+
+Embedding vectors are stored in dedicated search fields that are created by `Ibexa\Contracts\Core\Search\FieldType\EmbeddingFieldFactory`.
+These fields are then used by the search engine to perform vector similarity comparisons when embedding queries are executed.
+
+``` php
```
mnocon: Please extract to a separate code sample

docs/search/search_api.md
```diff
@@ -378,4 +358,147 @@ $query->aggregations[] = new IntegerRangeAggregation('range', 'person', 'age',
...
+use Ibexa\Contracts\Core\Search\FieldType\EmbeddingFieldFactory;
+use Ibexa\Contracts\Core\Search\Embedding\EmbeddingConfigurationInterface;
+
+// $config is an existing EmbeddingConfigurationInterface
+$factory = new EmbeddingFieldFactory($config);
```
mnocon:
```suggestion
/** @var \Ibexa\Contracts\Core\Search\Embedding\EmbeddingConfigurationInterface $config */
$factory = new EmbeddingFieldFactory($config);
```

`EmbeddingFieldFactory` is a Symfony service, is it even meant to be instantiated directly?

docs/search/search_api.md
```diff
@@ -378,4 +358,147 @@ $query->aggregations[] = new IntegerRangeAggregation('range', 'person', 'age',
...
+
+// Create a default embedding field (type derived from config suffix)
+$embeddingField = $factory->create();
+echo $embeddingField->getType(); // for example, "ibexa_dense_vector_model_123"
```
mnocon: Should the example match one of the fields defined in:
https://github.com/ibexa/solr/blob/96022e7f6993fade128054f25cc35b3446ef75e7/src/lib/Resources/config/solr/managed-schema.xml#L157-L167

as, indicated in the comment here: https://github.com/ibexa/core/blob/d01b203204e4f9bc3bdaced92ade82eff0afbdd3/src/contracts/Search/FieldType/EmbeddingField.php#L21C75-L21C100

?

docs/search/search_api.md
```diff
@@ -378,4 +358,147 @@ $query->aggregations[] = new IntegerRangeAggregation('range', 'person', 'age',
...
+- [EmbeddingProviderInterface](/api/php_api/php_api_reference/classes/Ibexa-Contracts-Core-Search-Embedding-EmbeddingProviderInterface.html) is the runtime contract for generating vector representations from text or other inputs
+
+- [EmbeddingProviderRegistryInterface](/api/php_api/php_api_reference/classes/Ibexa-Contracts-Core-Search-Embedding-EmbeddingProviderRegistryInterface.html) lists all available embedding providers
+
+- [EmbeddingProviderResolverInterface](/api/php_api/php_api_reference/classes/Ibexa-Contracts-Core-Search-Embedding-EmbeddingProviderResolverInterface.html) determines which provider should be used for a given embedding configuration
+
+### Embedding fields
+
+Embedding vectors are stored in dedicated search fields that are created by `Ibexa\Contracts\Core\Search\FieldType\EmbeddingFieldFactory`.
+These fields are then used by the search engine to perform vector similarity comparisons when embedding queries are executed.
```
mnocon:
```suggestion
These fields can be used by the search engine to perform vector similarity comparisons when embedding queries are executed.
```

Just creating the fields does nothing, and they are not stored anywhere. To actually store them, you need to expand the index (see the other comment about this)

docs/search/search_api.md
```diff
@@ -378,4 +358,147 @@ $query->aggregations[] = new IntegerRangeAggregation('range', 'person', 'age',
...
+// Create a default embedding field (type derived from config suffix)
+$embeddingField = $factory->create();
+echo $embeddingField->getType(); // for example, "ibexa_dense_vector_model_123"
+
+// Create a custom embedding field with a specific type
+$customField = $factory->create('custom_embedding_type');
+echo $customField->getType(); // "custom_embedding_type"
+```
+
+For more information, see [Embeddings reference](embeddings_reference.md).
```
mnocon, line 399: I'd link to https://doc.ibexa.co/en/5.0/search/extensibility/index_custom_elasticsearch_data/ here (I don't see an example for Solr right now).

Once you have the field, you have to listen to the indexing event (ContentIndexCreateEvent) and actually attach it to the index

docs/search/search_api.md
```diff
@@ -378,4 +358,147 @@ $query->aggregations[] = new IntegerRangeAggregation('range', 'person', 'age',
...
+!!! note "Embedding query properties"
+
+    Embedding queries do not allow standard Query properties such as `query`, `sortClauses`, `facetBuilders`, or `spellcheck`.
+
+### Embedding configuration and providers
+
+Models used to resolve embedding queries must be configured in [system configuration](configuration.md).
+Each key defines the model's name, vector dimensionality, the field suffix used in the search index, and the embedding provider that generates vectors.
+
+``` yaml
```
mnocon: I'd mention here:
1) The `default_embedding_model` property
2) That the `field_suffix` must be unique for each model

docs/search/embeddings_reference/embeddings_reference.md
```diff
@@ -0,0 +1,50 @@
...
+
+- [`Ibexa\Contracts\Core\Search\Embedding\EmbeddingProviderRegistryInterface`](/api/php_api/php_api_reference/classes/Ibexa-Contracts-Core-Search-Embedding-EmbeddingProviderRegistryInterface.html): Lists available embedding providers
```
mnocon:
```suggestion
- [`Ibexa\Contracts\Core\Search\Embedding\EmbeddingProviderRegistryInterface`](/api/php_api/php_api_reference/classes/Ibexa-Contracts-Core-Search-Embedding-EmbeddingProviderRegistryInterface.html): Lists available embedding providers or get one by its identifier
```

docs/search/embeddings_reference/embeddings_reference.md
```diff
@@ -0,0 +1,50 @@
+---
```
mnocon, line 1: Please link to the search_api page for an example, or just include the example here as well.

In general, there's a lot of duplicated content between these two pages (this one and search_api.md), and it's hard to say which one is the primary one - search api links here for more information, but it's search_api.md that contains more information 🤔 

This one does not include:
1) Description of EmbeddingQueryBuilder
2) Examples

Do we need this page?

docs/search/search_api.md
```diff
@@ -378,4 +358,147 @@ $query->aggregations[] = new IntegerRangeAggregation('range', 'person', 'age',
...
+Searching with embeddings can be combined with filtering, which allows the semantic search results to be constrained by content type, location, permissions, or other criteria.
+
+An embedding query is represented by the `Ibexa\Contracts\Core\Repository\Values\Content\EmbeddingQuery` value object.
+The object encapsulates the embedding used for similarity search and optional search parameters such as filtering, pagination, aggregations, and result counting.
+The query is validated before being executed to ensure that the embedding configuration is consistent with the system setup.
+
+The following components are used to build and validate embedding-based queries:
+
+- [EmbeddingQuery](/api/php_api/php_api_reference/classes/Ibexa-Contracts-Core-Repository-Values-Content-EmbeddingQuery.html):
+    Represents a semantic similarity search request.
```
mnocon: Do we need the `: `? This is not a pattern that was used in the doc previously, ever.


```suggestion
- [EmbeddingQuery](/api/php_api/php_api_reference/classes/Ibexa-Contracts-Core-Repository-Values-Content-EmbeddingQuery.html) represents a semantic similarity search request.
```

mnocon (changes requested):

docs/search/embeddings_reference/embeddings_reference.md
mikadamczyk: You are right, better would be to write "Criteria are not used to express the embedding similarity part of the query." or "Embedding queries do not use criteria for similarity itself; similarity is defined by the embedded query vector, while additional filtering can still be applied through the query filter."

docs/search/search_api.md
mikadamczyk: The idea was "Embedding queries can also be combined with filters to narrow down results, for example, by content type or location. Repository search also respects the permissions of the current user.

docs/search/search_api.md
mikadamczyk: You're right. That second sentence is not really about the `QueryValidatorInterface`. QueryValidatorInterface is about validating the query structure before execution. Provider or model resolution and vector compatibility are runtime concerns handled elsewhere, so mentioning them here mixes responsibilities and makes the description less clear.
I would keep this section focused on structural validation and move the runtime and provider note to a more appropriate place, such as the execution flow or embedding runtime behavior section.

docs/search/search_api.md
mnocon: One thing that's not clear for me from this PR if embeddings can be used only for content search or for location search as well.

The example shows content search only, the taxonomy index is created only for content index (https://github.com/ibexa/taxonomy/blob/main/src/lib/Event/Subscriber/TaxonomyEmbeddingIndexDataSubscriber.php#L28) - I'm just not sure if locations are supported

mnocon: I think it's:

1) You cannot use Location search (https://github.com/ibexa/core/blob/main/src/contracts/Repository/SearchService.php#L191) because EmbeddingQuery does not inherit from LocationQuery

2) Anyway, you still can filter by location, because content search supports a couple of location criteria (https://doc.ibexa.co/en/5.0/search/criteria_reference/search_criteria_reference/#search-criteria)

docs/search/search_api.md
mikadamczyk: Good point. But for me, direct instantiation is acceptable in isolated examples or tests; in application code, `EmbeddingFieldFactory` is definitely intended to be used as a Symfony service. If we want a full example it will be something like:
```
<?php

  declare(strict_types=1);

  namespace App\Example;

  use Ibexa\Contracts\Core\Search\FieldType\EmbeddingField;
  use Ibexa\Contracts\Core\Search\FieldType\EmbeddingFieldFactory;

  final readonly class ExampleService
  {
      public function __construct(
          private EmbeddingFieldFactory $embeddingFieldFactory,
      ) {
      }

      public function createDefaultField(): EmbeddingField
      {
          return $this->embeddingFieldFactory->create();
      }

      public function createCustomField(): EmbeddingField
      {
          return $this->embeddingFieldFactory->create('custom_embedding_type');
      }
  }
```
And then describe the result separately:
```
$defaultField = $exampleService->createDefaultField();
$defaultField->getType(); // e.g. "ibexa_dense_vector_gemini_embedding_001_1536_dv"

$customField = $exampleService->createCustomField();
$customField->getType(); // "custom_embedding_type"
```
This is only a good example of how the factory works. It is not a good example of taxonomy-specific indexing logic, because taxonomy sometimes needs an explicit type based on the taxonomy model suffix.

docs/search/search_api.md
mikadamczyk: It could be:
Models used to resolve embedding queries must be configured in [system configuration](configuration.md). Each entry defines the model name, vector dimensionality, the field suffix used in the search index, and the embedding provider that generates vectors. The default model is selected through `default_embedding_model`. Each model must use a unique `field_suffix`, as it becomes part of the indexed field name.`

docs/search/search_api.md
mnocon: Note: embeddings are content search only, but allow filtering by location (see https://github.com/ibexa/documentation-developer/pull/3029#discussion_r2916660839 )

docs/search/embeddings_reference/embeddings_reference.md
dabrt: However, the constraint is still there:
https://github.com/ibexa/core/pull/536/changes#diff-dc23b4b1b530baaa1bb8a8020a0c113dfcada0d7f12e4cb4d3222fcb88462570R96

docs/search/search_api.md
dabrt: The only thing about adding custom fields to Solrcan be found here:
https://doc.ibexa.co/en/5.0/search/extensibility/solr_document_field_mappers/
where a field mapper is used to add a `parent_name` field to the index.

docs/search/search_api.md
dabrt: Does the runtime validation include validating vector dimensionality and selecting the correct indexed field for similarity search?

docs/search/embeddings_reference/embeddings_reference.md
mnocon: Yes, but it's not in v5 - https://github.com/ibexa/core/blob/main/src/contracts/Repository/Values/Content/EmbeddingQuery.php (and with this PR you're targetting v5)

Faceted search is removed in v5, deprecated in 4.6 (aggregations should be used instead)

docs/search/search_api.md
mikadamczyk: `QueryValidatorInterface` is responsible for validating the query structure before execution. Vector dimensionality and similarity field selection are runtime concerns. They are handled later in the execution path. Field selection is determined by the configured embedding model and backend specific query mapping. Vector dimensionality is only effectively validated when the query reaches the search engine. Those checks are not performed by QueryValidatorInterface.

dabrt: @mikadamczyk Thank you!

docs/search/search_api.md
```diff
@@ -378,4 +358,64 @@ $query->aggregations[] = new IntegerRangeAggregation('range', 'person', 'age',
...
 
-See [Agrregation reference](aggregation_reference.md) for details of all available aggregations.
+See [Aggregation reference](aggregation_reference.md) for details of all available aggregations.
+
+## Search with embeddings
+
+
+!!! caution "Feature support"
+
+    Search with embeddings is only available with the Solr and Elasticsearch search engines.
```
mnocon: I think it's only supported with Solr9?

V5 supports both Solr 8 and Solr 9, I think only 9 supports embedding search

Please see this:
https://doc.ibexa.co/en/5.0/content_management/taxonomy/taxonomy/#enable-taxonomy-suggestions

> [Search engine](https://doc.ibexa.co/en/5.0/search/search_engines/search_engines/): Taxonomy suggestions require a search engine that supports vector search. The feature has been tested to work with Elasticsearch or Solr 9.8.1+.

docs/search/search_api.md
```diff
@@ -378,4 +358,64 @@ $query->aggregations[] = new IntegerRangeAggregation('range', 'person', 'age',
...
+
+!!! note "Taxonomy suggestions"
+
+    Embedding queries have been introduced primarily to support the [Taxonomy suggestions](taxonomy.md#taxonomy-suggestions) feature, therefore embedding search integration is provided for `TaxonomyEmbedding`.
+
+You can narrow down the search results, for example, by content type or location.
+To do this, combine searching with embeddings with filters.
+Repository search also respects the permissions of the current user.
+
+An embedding query is represented by the [`Ibexa\Contracts\Core\Repository\Values\Content\EmbeddingQuery`](/api/php_api/php_api_reference/classes/Ibexa-Contracts-Core-Repository-Values-Content-EmbeddingQuery.html)` value object.
```
mnocon:
```suggestion
An embedding query is represented by the [`Ibexa\Contracts\Core\Repository\Values\Content\EmbeddingQuery`](/api/php_api/php_api_reference/classes/Ibexa-Contracts-Core-Repository-Values-Content-EmbeddingQuery.html) value object.
```

docs/search/search_api.md
```diff
@@ -378,4 +358,64 @@ $query->aggregations[] = new IntegerRangeAggregation('range', 'person', 'age',
...
+
+### Use embedding queries in search
+
+Embedding queries are executed through the search API in the same way as other search requests.
+You build an `EmbeddingQuery` instance by using a builder and pass it to the search service.
+
+This example shows a minimal embedding query executed directly through the search service:
+
+``` php
+// ...
```
mnocon:
```suggestion
```
There's nothing above that in the code sample, let's skip this

docs/search/search_api.md
```diff
@@ -378,4 +358,64 @@ $query->aggregations[] = new IntegerRangeAggregation('range', 'person', 'age',
...
+The object encapsulates the embedding used for similarity search and optional search parameters such as filtering, pagination, aggregations, and result counting.
+
+### Use embedding queries in search
+
+Embedding queries are executed through the search API in the same way as other search requests.
+You build an `EmbeddingQuery` instance by using a builder and pass it to the search service.
+
+This example shows a minimal embedding query executed directly through the search service:
+
+``` php
```
mnocon:
```suggestion
``` php hl_lines="39-40 42-48 50"
```

docs/search/embeddings_reference/embeddings_reference.md
```diff
@@ -0,0 +1,101 @@
...
+ibexa:
+    system:
+        default:
+        embedding_models:
+            text-embedding-3-small:
+                name: 'text-embedding-3-small'
+                dimensions: 1536
+                field_suffix: '3small'
+                embedding_provider: 'ibexa_openai'
+        default_embedding_model: text-embedding-ada-002
```
mnocon: Should this part be nested one level deeper, under `default`?

docs/search/embeddings_reference/embeddings_reference.md
```diff
@@ -0,0 +1,101 @@
...
+
+- [`Ibexa\Contracts\Core\Search\Embedding\EmbeddingProviderRegistryInterface`](/api/php_api/php_api_reference/classes/Ibexa-Contracts-Core-Search-Embedding-EmbeddingProviderRegistryInterface.html) lists available embedding providers or gets one by its identifier
+
+- [`Ibexa\Contracts\Core\Search\Embedding\EmbeddingProviderResolverInterface`](/api/php_api/php_api_reference/classes/Ibexa-Contracts-Core-Search-Embedding-EmbeddingProviderResolverInterface.html) determines the embedding provider to be used for generating embeddings based on the system configuration, or a demand passed through the `resolveByModelIdentifier` method
+
+## Configuration
+
+Models used to resolve embedding queries must be configured per SiteAccess in [system configuration](configuration.md).
+Each entry defines the model's name, vector dimensionality, the field suffix, and the embedding provider that generates vectors.
+Field suffixes assigned to the models must be unique, as they becomes part of the indexed field name.
```
mnocon:
```suggestion
Field suffixes assigned to the models must be unique, as they become part of the indexed field name.
```

docs/search/embeddings_reference/embeddings_reference.md
```diff
@@ -0,0 +1,101 @@
...
+
+### EmbeddingQuery
+
+- [`Ibexa\Contracts\Core\Repository\Values\Content\EmbeddingQuery`](/api/php_api/php_api_reference/classes/Ibexa-Contracts-Core-Repository-Values-Content-EmbeddingQuery.html) represents a semantic similarity search request.
+    It encapsulates an [Embedding](#embedding) instance and supports pagination, aggregations, and result counting through the same API as standard content queries.
+
+    !!! note "Embedding query properties"
+
+        Embedding queries do not use criteria for similarity, but for additional filtering applied through the query filter. 
+        Also, embedding queries do not allow standard Query properties supported by [search engines](search_engines.md) other than the Legacy Search, such as `query`, `sortClauses`, or `spellcheck`
```
mnocon:
```suggestion
        Also, embedding queries do not allow standard Query properties supported by [search engines](search_engines.md) other than the Legacy Search, such as `query`, `sortClauses`, or `spellcheck`.
```

docs/search/embeddings_reference/embeddings_reference.md
```diff
@@ -0,0 +1,101 @@
...
+
+Embeddings provide vector representations of content or text, enabling [semantic similarity search](search_api.md#search-with-embeddings).
+Foundational abstractions are provided for embedding-based search, while embedding providers generate vector representations.
+
+Searching with embeddings is designed for use with the [Taxonomy suggestions](taxonomy.md#taxonomy-suggestions) feature.
+The [`Ibexa\Contracts\Taxonomy\Search\Query\Value\TaxonomyEmbedding`](/api/php_api/php_api_reference/classes/Ibexa-Contracts-Taxonomy-Search-Query-Value-TaxonomyEmbedding.html) class allows embedding queries to target taxonomy data.
+
+!!! tip
+
+    Searching with embeddings isn't possible with the Legacy Search engine.
```
mnocon: Please see my other comment about the supported search engines - I believe we need to be more specific here.

mnocon (commented):
Thank you!

A couple of nitpicks, but the most important comment is about search engine support - I believe we need to mention Solr 9 explicitly to avoid confusion.

docs/search/embeddings_reference/embeddings_reference.md
dabrt: Right as always
https://github.com/ibexa/connector-openai/blob/aeef78ab854135cbf95149b136e9c22ca417e010/src/bundle/Resources/config/prepend.yaml#L19

dabrt:
@mnocon The checks are coming through, I fixed the highlighting in the code sample, everything looks ready for merge and CP to 4.6

mnocon (approved):

PR #3051, opened 2026-02-09
Removed Jira mentions from the doc

mnocon:
Target: 4.6, 5.0

As suggested on Slack.

Looking at **https://issues.ibexa.co/browse/EZP-27987** , I think the issue was solved in https://issues.ibexa.co/browse/EZP-25487 - and we can remove this mention?


mateuszdebinski (approved):

adriendupuis (approved):

adriendupuis:
I was about to complain about the many spaces between the bullet and the first word (`-   Form deletion`), like a tab 4 spaces alignment, but it's every where :p

julitafalcondusza (approved):

PR #3080, opened 2026-03-11
Added embedding generation. Avoid creating an instance of EmbeddingFieldFactory manually

mnocon:


code_samples/api/public_php_api/src/Command/FindByTaxonomyEmbeddingCommand.php
```diff
@@ -46,8 +49,11 @@ protected function execute(
             0.1111,
         ];
 
+        $embeddingProvider = $this->embeddingProviderResolver->resolve();
```
dabrt, line 52: With the vector defined in line 45 and embedding generated by the provider there are two contradicting ways of getting the vector. But vector is not used anymore in the command, so... how about I remove it completely?

code_samples/api/public_php_api/src/Service/TaxonomyEmbeddingSearchService.php
```diff
@@ -14,9 +14,9 @@
 use Ibexa\Contracts\Core\Repository\Values\Content\Search\SearchResult;
 use Ibexa\Contracts\Taxonomy\Search\Query\Value\TaxonomyEmbedding;
 
-final class TaxonomyEmbeddingSearchService
+final readonly class TaxonomyEmbeddingSearchService
```
dabrt, line 17: You suggested removing the service altogether - do you think I should keep it?

How about the two samples stay but i change their order, so that the service is presented first and then...
The command shows how it can be used?

code_samples/api/public_php_api/src/Command/FindByTaxonomyEmbeddingCommand.php
dabrt: Also, what if the command example called the service to perform the searchByEmbedding function?
This way both the command and the service could stay in the example - the command would show how to obtain the embedding and call the service to find it, the service would call the query?

But... where would I want to put the filtering - command or the service? I guess the service?

PR #3081, opened 2026-03-11
Added option to remove indent from code samples

mnocon:
Target: 4.6, 5.0

Improving readability of the code samples by reducing their indent where it's not meaningful.

It should not be enabled by default, because it can result in wrong indent when a code sample is composed of multiple include_file calls (the indentation would break then).

I've decided to keep the indent for YAML services - it might make things easier when copy/pasting (assuming someone is using 4-spaces indent)

Before:

Screenshot 2026-03-11 at 11 38 57

After:

Screenshot 2026-03-11 at 11 39 03

### DIFF
--- a/docs/administration/recent_activity/recent_activity.md
+++ b/docs/administration/recent_activity/recent_activity.md
@@ -190,7 +190,7 @@ In the following example, several actions are logged into one context group, eve
     - `complete`
 
 ``` php
-[[= include_file('code_samples/recent_activity/src/Command/ActivityLogContextTestCommand.php', 46, 66) =]]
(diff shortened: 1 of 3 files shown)


adriendupuis (approved):
I tested with https://doc.ibexa.co/en/5.0/administration/recent_activity/recent_activity/#context-group VS https://ez-systems-developer-documentation--3081.com.readthedocs.build/en/3081/administration/recent_activity/recent_activity/#context-group and it works really well!

Python code is hard to read but the way it get how many spaces it can remove with `indent = min(len(l) - len(l.lstrip()) for l in non_empty)` looks solid to me.

PR #3092, opened 2026-03-17
Replaced pagelayout with page_layout

mnocon:
As reported on Slack.

Target: 4.6, 5.0


alongosz (approved):
Archaeology - original PR for the reference: https://github.com/ezsystems/ezpublish-kernel/pull/2629. Though indeed removed completely in 5.0 only.

dabrt (approved):

PR #3105, opened 2026-03-25
IBX-10701: `ibexa_tracking_script` function described in Dev-doc

julitafalcondusza:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | [IBX-10701](https://ibexa.atlassian.net/browse/IBX-10701) ibexa/connector-raptor#5
| Versions      | 5.0
| Editions | Headless+

`ibexa_tracking_script` function described in Dev-doc.

#### Checklist

- [ ] Text renders correctly
- [ ] Text has been checked with vale
- [ ] Description metadata is up to date
- [ ] Redirects cover removed/moved pages
- [ ] Code samples are working
- [ ] PHP code samples have been fixed with PHP CS fixer
- [ ] Added link to this PR in relevant JIRA ticket or code PR

[IBX-10701]: https://ibexa.atlassian.net/browse/IBX-10701?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #3110, opened 2026-03-30
IBX-10709: Tracking modes and events - Raptor integration

julitafalcondusza:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | [IBX-10709](https://ibexa.atlassian.net/browse/IBX-10709) ibexa/connector-raptor#3 ibexa/connector-raptor#28
| Versions      | 5.0
| Editions | LTS Update for Headless+

Tracking modes and events - Raptor integration

#### Checklist

- [ ] Text renders correctly

[IBX-10709]: https://ibexa.atlassian.net/browse/IBX-10709?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/cdp/raptor_integration/tracking_event_function.md
+++ b/docs/cdp/raptor_integration/tracking_event_function.md
@@ -69,9 +69,9 @@ In this case, `context` parameter allows to override the product category by pas
 
 ## Tracking modes
 
-Tracking user interactions can be implemented on the Client side or the Server side.
+Tracking user interactions can be implemented on the client-side or the server-side.
 Each approach differs in where events are captured and how they are sent to the tracking backend.
-The `ibexa_tracking_track_event()` function works for both client-side and server-side tracking based on `tracking_type` configuration.
+The `ibexa_tracking_track_event()` function works based on `tracking_type` configuration.
 
 The tracking function outputs different content depending on the mode:
 
@@ -195,7 +195,6 @@ class ProductViewTrackingSubscriber implements EventSubscriberInterface
     }
 }
 ```
-
 
 ## Tracking events
 
--- a/docs/cdp/raptor_integration/tracking_event_function.md
+++ b/docs/cdp/raptor_integration/tracking_event_function.md
@@ -1,0 +1,124 @@
+---
+description: Raptor tracking event function.
+---
+
+# Raptor tracking event function
+
+This module introduces visit tracking functionality for collecting user interactions with products and content.
+The implementation includes product visit tracking with mapping to tracking parameters and automatic price formatting, as well as Twig functions for straightforward integration.
+It supports multi-currency setups with automatic decimal formatting (0, 2, or 3 decimals) based on configuration, and integrates with taxonomy to extract and format category paths for product categorization.
+
+## Initialize Raptor tracking script
+
+First, initialize the Raptor tracking script in your base layout template, typically within the <head> section or before the closing <body> tag:
+
+``` html+twig
+{# templates/base.html.twig #}
+<!DOCTYPE html>
+<html>
+<head>
+    {# ... other head content ... #}
+
+    {# Initialize Raptor tracking - must be called before any tracking events #}
+    {{ ibexa_tracking_script() }}
+</head>
+<body>
+    {# ... page content ... #}
+</body>
+</html>
+```
+
+## Tracking events
+
+### Product Visit Event
+
+This event tracks product page visits by users.
+This is the most common e-commerce tracking event used to capture product views for analytics, recommendation models, and user behavior processing.
+
+Required Data:
+
+- Product object (implements `ProductInterface`)
+
+Example:
+
+``` html+twig
+{# templates/product/view.html.twig #}
+{% extends 'base.html.twig' %}
+
+{% block content %}
+    <div class="product-details">
+        <h1>{{ product.name }}</h1>
+        <p>{{ product.description }}</p>
+        <div class="price">{{ product.price }}</div>
+    </div>
+
+    {# Track product visit #}
+    {{ ibexa_tracking_track_event('visit', product) }}
+{% endblock %}
+```
+
+### Basket Event
+
+This event tracks when a product is added to the shopping basket.
+It catches user interest and helps with conversion tracking and product recommendations.
+
+Required Data:
+- Product object (the product being added to cart)
+- Context array with basket information
+
+Example:
+
+``` html+twig
+{# templates/cart/add_confirmation.html.twig #}
+{% extends 'base.html.twig' %}
+
+{% block content %}
+    <div class="cart-notification">
+        <p>Product "{{ product.name }}" has been added to your cart!</p>
+        <p>Quantity: {{ addedQuantity }}</p>
+    </div>
+
+    {# Build basket content string: "SKU:quantity;SKU:quantity" #}
+    {% set basketContent = [] %}
+    {% for item in basket.items %}
+        {% set basketContent = basketContent|merge([item.product.code ~ ':' ~ item.quantity]) %}
+    {% endfor %}
+
+    {# Track basket addition #}
+    {% set basketContext = {
+        'basketContent': basketContent|join(';'),
+        'basketId': basket.id,
+        'quantity': addedQuantity
+    } %}
+
+    {{ ibexa_tracking_track_event('basket', product, basketContext) }}
+
+    <a href="{{ path('cart_view') }}">View Cart</a>
+{% endblock %}
+```
+
+Simplified example with Twig filter:
+
+``` html+twig
+{# If you have a custom Twig filter to format basket content #}
+{% set basketContext = {
+    'basketContent': basket|format_basket_content,  {# Returns "SKU-1:2;SKU-2:1;SKU-3:5" #}
+    'basketId': basket.id,
+    'quantity': addedQuantity
+} %}
+
+{{ ibexa_tracking_track_event('basket', product, basketContext) }}
+```
+
+### Custom Templates
+
+You can override the default tracking templates by providing a custom template path:
+
+``` bash
+{{ ibexa_tracking_track_event(
+    'visit',
+    product,
+    {},
+    '@MyBundle/tracking/custom_visit.html.twig'
+) }}
+```
--- a/docs/cdp/raptor_integration/tracking_event_function.md
+++ b/docs/cdp/raptor_integration/tracking_event_function.md
@@ -1,16 +1,16 @@
 ---
-description: Raptor tracking event function.
+description: Raptor tracking function.
 ---
 
-# Raptor tracking event function
+# Raptor tracking function
 
-This module introduces visit tracking functionality for collecting user interactions with products and content.
+Raptor [tracking function](https://content.raptorservices.com/help-center/introduction-to-tracking-documentation) introduces visit tracking functionality for collecting user interactions with products and content.
 The implementation includes product visit tracking with mapping to tracking parameters and automatic price formatting, as well as Twig functions for straightforward integration.
 It supports multi-currency setups with automatic decimal formatting (0, 2, or 3 decimals) based on configuration, and integrates with taxonomy to extract and format category paths for product categorization.
 
 ## Initialize Raptor tracking script
 
-First, initialize the Raptor tracking script in your base layout template, typically within the <head> section or before the closing <body> tag:
+First, initialize the `ibexa_tracking_script()` tracking script in your base layout template, typically within the <head> section or before the closing <body> tag:
 
 ``` html+twig
 {# templates/base.html.twig #}
@@ -28,16 +28,86 @@ First, initialize the Raptor tracking script in your base layout template, typic
 </html>
 ```
 
+## Tracking parameters
+
+Tracking is handled through twig function that accept following parameters:
+
+``` html+twig
+ibexa_tracking_track_event(
+    eventType,     {# string: 'visit', 'contentvisit', 'buy', 'basket', 'itemclick' #}
+    data,          {# mixed: Product, Content, or null (optional) #}
+    context,       {# array: additional context data (optional) #}
+    template       {# string: custom template path (optional) #}
+)
+```
+
+- **eventType** - type: string, defines the type of tracking event to be sent, for example, `visit`, `contentvisit`, `buy`, `basket`, `itemclick`
+- **data** (optional) - type: mixed, defines the primary object associated with the event, such as a Product or Content, can be null if not required
+- **context** (optional)- type: array, additional event data, such as quantity, basket details, or custom parameters
+- **template** (optional) - type: string, path to a custom Twig template used to render the tracking event, allows overriding the default tracking output
+
+### `context` parameter - example usage
+
+You can use `context` parameter to add additional data.
+
+During tracking, for products assigned to multiple categories, the system uses the first category.
+In this case, `context` parameter allows to override the product category by passing a category identifier:
+
+``` html+twig
+{% block content %}
+    <div class="product-details">
+        <h1>{{ product.name }}</h1>
+        {# ... product content ... #}
+    </div>
+
+    {# Track with category identifier - automatic loading and formatting #}
+    {{ ibexa_tracking_track_event('visit', product, {
+        'categoryIdentifier': 'electronics'
+    }) }}
+{% endblock %}
+```
+
+## Tracking modes
+
+Tracking user interactions can be implemented on the Client side or the Server side.
+Each approach differs in where events are captured and how they are sent to the tracking backend.
+The `ibexa_tracking_track_event()` function works for both client-side and server-side tracking based on `tracking_type` configuration.
+
+The tracking function outputs different content depending on the mode:
+
+``` yaml
+# Server-side tracking
+connector_raptor:
+    tracking_type: 'server'  # Returns HTML comments
+
+# Client-side tracking
+connector_raptor:
+    tracking_type: 'client'  # Returns <script> tags
+```
+
+- **server** - returns HTML comments, leaving placeholders without running JavaScript.
+- **client** - returns `script` tags to load the tracking script in the browser.
+
+You can switch tracking mode anytime by changing the `tracking_type` parameter.
+
+For more information on Tracking modes, see Raptor documentation:
+
+- [Client-side tracking](https://content.raptorservices.com/help-center/client-side-tracking)
+- [Server-side tracking](https://content.raptorservices.com/help-center/server-side-tracking)
+- [Client-side vs. Server-side tracking](https://content.raptorservices.com/help-center/client-side-vs.-server-side-tracking)
+
 ## Tracking events
 
-### Product Visit Event
+The following events are supported and can be triggered from Twig templates:
+
+### Product Visit event
 
 This event tracks product page visits by users.
-This is the most common e-commerce tracking event used to capture product views for analytics, recommendation models, and user behavior processing.
+It's the most common e-commerce tracking event used to capture product views for analytics, recommendation models, and user behavior processing.
 
-Required Data:
+Required data:
 
-- Product object (implements `ProductInterface`)
+- **Product object** - defines the product being tracked. It implements `ProductInterface` so the system can read its information (for example, ID, price, category).
 
 Example:
 
@@ -57,14 +127,22 @@ Example:
 {% endblock %}
 ```
 
-### Basket Event
+### Content Visit event
+
+This event tracks content page visits by users.
+It can used to check content views for analytics, personalization, and user behavior tracking.
+
+- **Content object** - defines the content being tracked.
+
+### Basket event
 
 This event tracks when a product is added to the shopping basket.
 It catches user interest and helps with conversion tracking and product recommendations.
 
-Required Data:
-- Product object (the product being added to cart)
-- Context array with basket information
+Required data:
+
+- **Product object** - defines the product being added to the basket.
+- **Context array with basket information** - provides optional data about the basket, like quantity or basket ID, to provide context for the event.
 
 Example:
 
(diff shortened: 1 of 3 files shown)


docs/cdp/raptor_integration/tracking_event_function.md
```diff
@@ -0,0 +1,302 @@
+---
+description: Raptor tracking function.
+---
+
+# Raptor tracking function
+
+Raptor [tracking function](https://content.raptorservices.com/help-center/introduction-to-tracking-documentation) introduces visit tracking functionality for collecting user interactions with products and content.
+The implementation includes product visit tracking with mapping to tracking parameters and automatic price formatting, as well as Twig functions for straightforward integration.
```
mnocon: The mention of automatic price formatting feels out of place, and is never explained - I don't know what it means to be honest

docs/cdp/raptor_integration/tracking_event_function.md
```diff
@@ -0,0 +1,302 @@
...
+It supports multi-currency setups with automatic decimal formatting (0, 2, or 3 decimals) based on configuration, and integrates with taxonomy to extract and format category paths for product categorization.
```
mnocon: 1. Add a link to Taxonomy
2. This configuration of decimal formatting is never described?

docs/recommendations/raptor_integration/tracking_event_function.md
```diff
@@ -0,0 +1,302 @@
...
+
+# Raptor tracking function
+
+Raptor [tracking function](https://content.raptorservices.com/help-center/introduction-to-tracking-documentation) introduces visit tracking functionality for collecting user interactions with products and content.
+The implementation includes product visit tracking with mapping to tracking parameters and automatic price formatting, as well as Twig functions for straightforward integration.
+It supports multi-currency setups with automatic decimal formatting (0, 2, or 3 decimals) based on configuration, and integrates with taxonomy to extract and format category paths for product categorization.
+
+## Initialize Raptor tracking script
+
+First, initialize the `ibexa_tracking_script()` tracking script in your base layout template, typically within the <head> section or before the closing <body> tag:
```
mnocon, line 12: As discussed on Slack, there' some overlap between this section and https://github.com/ibexa/documentation-developer/pull/3105/changes

docs/recommendations/raptor_integration/tracking_event_function.md
```diff
@@ -0,0 +1,302 @@
...
+
+``` html+twig
+{# templates/base.html.twig #}
```
mnocon, line 15:
```suggestion
{# templates/standard/pagelayout.html.twig #}
```
? Something for @alongosz to verify - but this is the default value we use I believe:
```
            page_layout: "@ibexadesign/pagelayout.html.twig"
```

docs/cdp/raptor_integration/tracking_event_function.md
```diff
@@ -0,0 +1,302 @@
...
+ibexa_tracking_track_event(
+    eventType,     {# string: 'visit', 'contentvisit', 'buy', 'basket', 'itemclick' #}
+    data,          {# mixed: Product, Content, or null (optional) #}
+    context,       {# array: additional context data (optional) #}
+    template       {# string: custom template path (optional) #}
+)
+```
+
+- **eventType** - type: string, defines the type of tracking event to be sent, for example, `visit`, `contentvisit`, `buy`, `basket`, `itemclick`
+- **data** (optional) - type: mixed, defines the primary object associated with the event, such as a Product or Content, can be null if not required
```
mnocon:
```suggestion
- **data** (optional) - type: mixed, accepts the primary object associated with the event, such as a Product or Content, can be null if not required
```

Looking at the code, the built-in supported types are:

Content Visit: Content
Itemclick: product id (is product id the same as product code)?
Visit, Buy, Basket: ProductInterface

and we should narrow this down (and maybe show how to implement a custom `EventMappingStrategyInterface`? or that would be an overkill?)

docs/cdp/raptor_integration/tracking_event_function.md
```diff
@@ -0,0 +1,302 @@
...
+```
+
+## Tracking parameters
+
+Tracking is handled through twig function that accept following parameters:
+
+``` html+twig
+ibexa_tracking_track_event(
+    eventType,     {# string: 'visit', 'contentvisit', 'buy', 'basket', 'itemclick' #}
+    data,          {# mixed: Product, Content, or null (optional) #}
```
alongosz:
```suggestion
    data,          {# mixed: product, content, or null (optional) #}
```
`product` and `content` are contextual variables available in views.

docs/recommendations/raptor_integration/tracking_event_function.md
```diff
@@ -0,0 +1,302 @@
...
+In this case, `context` parameter allows to override the product category by passing a category identifier:
+
+``` html+twig
+{% block content %}
+    <div class="product-details">
+        <h1>{{ product.name }}</h1>
+        {# ... product content ... #}
+    </div>
+
+    {# Track with category identifier - automatic loading and formatting #}
```
mnocon, line 62: > automatic loading and formatting

This isn't clear

docs/cdp/raptor_integration/tracking_event_function.md
```diff
@@ -0,0 +1,302 @@
...
+# Server-side tracking
+connector_raptor:
+    tracking_type: 'server'  # Returns HTML comments
+
+# Client-side tracking
+connector_raptor:
+    tracking_type: 'client'  # Returns <script> tags
+```
+
+- **server** - returns HTML comments, leaving placeholders without running JavaScript.
```
mnocon:
```suggestion
- **server** - returns HTML comments, placeholders that do not perform any action. Instead, the tracking is done server-side.
```
The current description omits what happens, it states "server side returns HTML comments (which do nothing)"

docs/recommendations/raptor_integration/tracking_event_function.md
```diff
@@ -0,0 +1,302 @@
...
+
+For more information on Tracking modes, see Raptor documentation:
+
+- [Client-side tracking](https://content.raptorservices.com/help-center/client-side-tracking)
+- [Server-side tracking](https://content.raptorservices.com/help-center/server-side-tracking)
+- [Client-side vs. Server-side tracking](https://content.raptorservices.com/help-center/client-side-vs.-server-side-tracking)
+
+## Advanced usage – direct interaction with the service
+
+### `EventMapper` method
```
mnocon, line 100: I'd consider extracting this to a separate PHP API section - called "Tracking with PHP API"

docs/cdp/raptor_integration/tracking_event_function.md
```diff
@@ -0,0 +1,302 @@
...
+- [Client-side tracking](https://content.raptorservices.com/help-center/client-side-tracking)
+- [Server-side tracking](https://content.raptorservices.com/help-center/server-side-tracking)
+- [Client-side vs. Server-side tracking](https://content.raptorservices.com/help-center/client-side-vs.-server-side-tracking)
+
+## Advanced usage – direct interaction with the service
+
+### `EventMapper` method
+
+The recommended method, providing full control over event tracking, is `EventMapper` method.
+It allows to interact directly with the service, supporting advanced configuration beyond what the Page Builder offers.
```
mnocon:
```suggestion
It allows to interact directly with the service, supporting advanced use cases not covered by default implementation.
```
Page Builder does not offer tracking at all? And it's customization, not configuration?

docs/recommendations/raptor_integration/tracking_event_function.md
```diff
@@ -0,0 +1,302 @@
...
+
+The recommended method, providing full control over event tracking, is `EventMapper` method.
+It allows to interact directly with the service, supporting advanced configuration beyond what the Page Builder offers.
+
+Check the following example:
+
+``` php
```
mnocon, line 107: Please extract the code samples to separate files and run PHPStan on them - this one will fail, because ProductInterface is not imported

docs/recommendations/raptor_integration/tracking_event_function.md
```diff
@@ -0,0 +1,302 @@
...
+Manual creation of EventData allows precise control over the events sent to the service.
+It enables to define custom event parameters, track specific user interactions, and tailor data collection to advanced use cases.
+
+Check the following example:
+
+``` php
+use Ibexa\Contracts\ConnectorRaptor\Tracking\Event\VisitEventData;
+
+$eventData = new VisitEventData(
+    productId: $product->getCode(),
```
mnocon, line 141: This code sample uses named arguments - do we actually recommend this?

I think this will be the first usage in the doc, and something that forces us to keep the BC promise on argument names (not just order).

Asking just for confirmation @alongosz

docs/recommendations/raptor_integration/tracking_event_function.md
```diff
@@ -0,0 +1,302 @@
...
+    productName: $product->getName(),
+    categoryPath: '25#Electronics;26#Smartphones',  // Build manually
```
mnocon, line 143: Does the `25#` format mean something? If we want people to build this manually, we have to explain how the format works

docs/recommendations/raptor_integration/tracking_event_function.md
```diff
@@ -0,0 +1,302 @@
...
+    itemPrice: '999.99'
+);
+
+$this->trackingDispatcher->dispatch($eventData);
+```
+
+### Example - event subscriber
+
+If you need to track events automatically based on application events, you can use Event Subscriber.
+It reacts to specific events in the application and triggers tracking logic without the need to add it manually in templates.
```
mnocon, line 154: I'd add a link to the event reference, and maybe mention that Ibexa events can be used - for example CreateOrderEvent (https://doc.ibexa.co/en/5.0/api/event_reference/order_management_events/)?

docs/recommendations/raptor_integration/tracking_event_function.md
```diff
@@ -0,0 +1,302 @@
...
+``` html+twig
+ibexa_tracking_track_event(
+    eventType,     {# string: 'visit', 'contentvisit', 'buy', 'basket', 'itemclick' #}
+    data,          {# mixed: Product, Content, or null (optional) #}
+    context,       {# array: additional context data (optional) #}
+    template       {# string: custom template path (optional) #}
+)
+```
+
+- **eventType** - type: string, defines the type of tracking event to be sent, for example, `visit`, `contentvisit`, `buy`, `basket`, `itemclick`
```
mnocon, line 43: Please link to the section where this is explained

docs/cdp/raptor_integration/tracking_event_function.md
```diff
@@ -0,0 +1,302 @@
...
+        $this->trackingDispatcher->dispatch($eventData);
+    }
+}
+```
+
+## Tracking events
+
+The following events are supported and can be triggered from Twig templates:
+
+### Product Visit event
```
mnocon: Note: if I understand correctly, in code the event type is `visit`? There's no "Product Visit"?

docs/recommendations/raptor_integration/tracking_event_function.md
```diff
@@ -0,0 +1,302 @@
...
+``` html+twig
+{# If you have a custom Twig filter to format basket content #}
+{% set basketContext = {
+    'basketContent': basket|format_basket_content,  {# Returns "SKU-1:2;SKU-2:1;SKU-3:5" #}
+    'basketId': basket.id,
+    'quantity': addedQuantity
+} %}
+
+{{ ibexa_tracking_track_event('basket', product, basketContext) }}
+```
```
mnocon, line 288: Why no buy and item clicked descriptions? Especially item clicked, I'd like to learn how this should be used - and it's the one accepting productId (product code I guess)

docs/recommendations/raptor_integration/tracking_event_function.md
```diff
@@ -0,0 +1,302 @@
...
+
+### Custom Templates
+
+You can override the default tracking templates by providing a custom template path:
```
mnocon, line 292: This should be higher I guess, together with the ibexa_tracking_track_event?

And it would be nice to explain why someone would like to do this - and what are the requirements.

Should the template call `window.raptor.push(event, params);` to actually send the events?

docs/recommendations/raptor_integration/tracking_event_function.md
```diff
@@ -0,0 +1,302 @@
+---
+description: Raptor tracking function.
```
adriendupuis, line 2:
```suggestion
description: Raptor tracking function.
month_change: true
```

docs/cdp/raptor_integration/tracking_event_function.md
julitafalcondusza: Old PR description, can be removed I guess

docs/cdp/raptor_integration/tracking_event_function.md
julitafalcondusza: This is a part of old PR description, we can remove it

docs/cdp/raptor_integration/tracking_event_function.md
julitafalcondusza: Yes, but there's a visit for product and content. I'll update with `visit`

docs/recommendations/raptor_integration/tracking_event_function.md
julitafalcondusza: we'll be resolved while creating feature-branch

docs/recommendations/raptor_integration/tracking_event_function.md
adriendupuis:
```suggestion
{# templates/themes/standard/pagelayout.html.twig #}
```

But I might introduce this path into the introduction sentence instead of the file. Something like:

> First, initialize the `ibexa_tracking_script()` tracking script in your base layout template, typically within the <head> section or before the closing <body> tag. For example, it could look like this in a `templates/themes/standard/pagelayout.html.twig` file :

alongosz: `@ibexadesign/pagelayout.html.twig` is the correct way to reference page layout template, that's the one from core actually - `vendor/ibexa/core/src/bundle/Core/Resources/views/pagelayout.html.twig` - standard design maps core templates to standard theme.

In this context extending `@ibexadesign/pagelayout.html.twig` makes more sense. 

------
Trivia: the only situation we have to be careful with is the one when we're trying to override page layout itself. If we place `templates/themes/<theme>/pagelayout.html.twig` to override the page layout from core, then we cannot do:
```twig
{% extends @ibexadesign/pagelayout.html.twig %}
```
It **has to override**:
```twig
{% extends @IbexaCore/pagelayout.html.twig %}
```
otherwise the template would reference itself which would create a loop.

But in this context we rather want to showcase that it needs to be placed in a site page layout, which can fairly often be derived from the core page layout. It all depends on a custom project I guess.

docs/cdp/raptor_integration/tracking_event_function.md
alongosz: > is product id the same as product code

Yes. What's important is that product id has to reference uniquely a @copilot specific product with an arbitrary string. So 2 different products cannot use the same ID and the ID should come from PIM (whether it's DXP PIM or a remote PIM like Quable).


> and we should narrow this down (and maybe show how to implement a custom `EventMappingStrategyInterface`? or that would be an overkill?)

Not sure, POV ping @tbialcz

docs/cdp/raptor_integration/tracking_event_function.md
alongosz: > Page Builder does not offer tracking at all? 

`ibexa_tracking_script` and `ibexa_tracking_track_event` Twig functions provide the actual tracking. They can be used in Headless, while PB for displaying recommendations is available in Experience.

> And it's customization, not configuration?

Depends on the doc language usually used. It's mostly about embedding tracking Twig function into templates or using the service, and mapping, and custom controllers. So for me it's rather customization.

docs/recommendations/raptor_integration/tracking_event_function.md
alongosz: > This code sample uses named arguments - do we actually recommend this?
> 
> I think this will be the first usage in the doc, and something that forces us to keep the BC promise on argument names (not just order).
> 
> Asking just for confirmation @alongosz

Yes and no. We use it in places where it makes sense for readability. Like this one. So if we document it, then it's an SPI and we need to keep BC. Which is fine. Even if we didn't document named arguments, we'd still need to keep SPI BC on `VisitEventData` itself (arguments order, no new non-default arguments).

Weirdly enough, in the worst case scenario, from my POV, it would be better if the code crashes on mistakenly changed argument name than silently providing wrong arguments in wrong places. Of course, we'd like to avoid making such BC breaks at all.

docs/recommendations/raptor_integration/tracking_event_function.md
alongosz: > Does the `25#` format mean something? If we want people to build this manually, we have to explain how the format works

Yes, it's very relevant. `CategoryId#CategoryName` (per Raptor docs). The only source of data is the tracking itself, so it has to provide these information. AFAIR if ID is missing, it's still fine to provide just `CategoryName`.

We also created simpler way of fetching category ID based on category name, but I'm not sure if it can be simplified easily here // POV ping @tbialcz

docs/cdp/raptor_integration/tracking_event_function.md
alongosz: >> Note: if I understand correctly, in code the event type is `visit`? There's no "Product Visit"?
>
> Yes, but there's a visit for product and content. I'll update with `visit`

Content visits should use dedicated `contentvisit` event. `visit` applies to products.

docs/recommendations/raptor_integration/tracking_event_function.md
alongosz: > Should the template call window.raptor.push(event, params); to actually send the events?

Typically no, but the system provides you with the ability to override `.js.twig` files via Design Engine if you really want. I expect that you'd typically want to override everything which surrounds JS, depending on custom project implementation. `.html.twig` files include inside `.js.twig` files.

docs/recommendations/raptor_integration/tracking_event_function.md
```diff
@@ -0,0 +1,301 @@
...
+            return;
+        }
+
+        $request = $event->getRequest();
+
+        // Example: track only if request has specific attribute
+        $product = $request->attributes->get('product');
+        if (!$product) {
+            return;
+        }
```
alongosz, line 190: Clean code:
```suggestion
        if (null === $product) {
            return;
        }
```

docs/recommendations/raptor_integration/tracking_event_function.md
```diff
@@ -0,0 +1,301 @@
...
+    {{ ibexa_tracking_track_event('visit', product) }}
+{% endblock %}
+```
+
+### Content `visit` event
+
+This event tracks content page visits by users.
+It can used to check content views for analytics, personalization, and user behavior tracking.
+
+- **Content object** - defines the content being tracked.
```
alongosz, line 234: Following one of my previous comments, I think this section lacks `contentvisit` HTML/Twig code snippet.

docs/cdp/raptor_integration/tracking_event_function.md
tbialcz: Regarding EventMappingStrategyInterface, I think including it here might be overkill for this basic reference doc. If needed, custom event mapping could be covered in a separate “Advanced customization” section with a proper implementation example.

docs/recommendations/raptor_integration/tracking_event_function.md
tbialcz: The format needs explanation per Raptor docs:
  - Format: `CategoryId#CategoryName;CategoryId#CategoryName` (e.g.,`25#Electronics;26#Smartphones`)
  - If CategoryName is missing, repeat the ID (e.g., `25#25;26#26`)
  - If CategoryId is missing, use just CategoryName (e.g., `Electronics;Smartphones`)

**HOWEVER**, the example showing manual categoryPath building is misleading. We should show the **recommended approaches first**:

  **✅ Recommended approach 1 (Twig):**
  ```twig
  {# CategoryPath is built automatically from product taxonomy #}
  {{ ibexa_tracking_track_event('visit', product) }}

  {# With category override #}
  {{ ibexa_tracking_track_event('visit', product, {
      'categoryIdentifier': 'electronics'
  }) }}

 ✅ Recommended approach 2 (PHP with EventMapper):
  // CategoryPath is extracted automatically from taxonomy via ProductEventMappingStrategy
  $eventData = $this->eventMapper->map(EventType::VISIT, $product);
  $this->trackingDispatcher->dispatch($eventData);

  // With category override
  $eventData = $this->eventMapper->map(
      EventType::VISIT,
      $product,
      [EventContext::CATEGORY_IDENTIFIER => 'electronics']
  );
  $this->trackingDispatcher->dispatch($eventData);

❌ NOT recommended (manual VisitEventData creation):
  // Only use if you have a very specific custom use case
  // You must build categoryPath manually according to Raptor format
  $eventData = new VisitEventData(
      productId: $product->getCode(),
      productName: $product->getName(),
      categoryPath: '25#Electronics;26#Smartphones',  // Manual - prone to errors
      currency: 'USD',
      itemPrice: '999.99'
  );
  $this->trackingDispatcher->dispatch($eventData);

  If we want to show how to build categoryPath manually from taxonomy, we should add it in an "Advanced" section with a proper example using TaxonomyService, but clearly mark it as NOT the recommended way.

docs/recommendations/raptor_integration/tracking_event_function.md
tbialcz: Agree on moving this higher, right after the basic ibexa_tracking_track_event examples 👍

Typically you wouldn’t call window.raptor.push(event, params) manually — it’s handled by the default templates.
If needed, overrides via Design Engine are mainly for adjusting the surrounding .js.twig / .html.twig structure rather than changing tracking logic itself.

docs/recommendations/raptor_integration/tracking_event_function.md
tbialcz:
```
{# templates/bundles/IbexaCoreBundle/default/content/full.html.twig #}
{% extends '@!IbexaCore/default/content/full.html.twig' %}

{% block content %}
    {{ parent() }}
    {{ ibexa_tracking_track_event('contentvisit', content) }}
{% endblock %}
```

docs/recommendations/raptor_integration/tracking_event_function.md
tbialcz:
```
{# templates/tracking/custom_visit.html.twig #}

  {#
   # Custom visit tracking template
   #
   # Available variables:
   # - parameters: array of Raptor tracking parameters (p1, p2, p3, etc.)
   # - debug: boolean flag to enable debug console messages
   #}

  <script type="text/javascript">
      {% autoescape 'js' %}
          (function () {
              // Custom logic before tracking
              console.log('Custom visit tracking template');
              console.log('Tracking parameters:', {{ parameters|json_encode|raw }});

              // Send the tracking event (REQUIRED for tracking to work)
              const event = 'trackEvent';
              const params = {{ parameters|json_encode|raw }};
              window.raptor.push(event, params);

              // Custom logic after tracking
              {% if debug %}
              console.log('Visit event tracked successfully');
              {% endif %}
          })();
      {% endautoescape %}
  </script>
```
Usage:
```
  {{ ibexa_tracking_track_event(
      'visit',
      product,
      {},
      '@App/tracking/custom_visit.html.twig'
  ) }}
```

docs/recommendations/raptor_integration/tracking_event_function.md
```diff
@@ -0,0 +1,301 @@
...
+Manual creation of EventData allows precise control over the events sent to the service.
+It enables to define custom event parameters, track specific user interactions, and tailor data collection to advanced use cases.
+
+Check the following example:
+
+``` php
+use Ibexa\Contracts\ConnectorRaptor\Tracking\Event\VisitEventData;
+
+$eventData = new VisitEventData(
+    productId: $product->getCode(),
```
alongosz, line 141:
```suggestion
    productCode: $product->getCode(),
```

@mnocon will say "I told you so" 🤪  
It's fine. We had an inconsistent naming in the code. Documentation helped discovering it :)

alongosz (commented):
The references to `productId` will change after ibexa/connector-raptor#48 (which will be merged before sanities, so it's safe to update it now).

Side topic: shouldn't code samples be placed in dedicated files only included in Markdown?

PR #3118, opened 2026-04-01
Fix external link icon going to new line alone

adriendupuis:
| Question      | Answer
| ------------- | ---
| JIRA Ticket   | 
| Versions      | 
| Edition       | 

Before, the icon indicating an external link can go to new line by itself.
For example, https://doc.ibexa.co/en/5.0/release_notes/ibexa_dxp_v5.0/#special-characters-in-online-editor within a 1590px width viewport.
image

With this PR, this icon can't be alone on a new line.
For example, https://ez-systems-developer-documentation--3118.com.readthedocs.build/en/3118/release_notes/ibexa_dxp_v5.0/#special-characters-in-online-editor within a 1590px width viewport.
image

Because of `:after`, the icon is not really in the link, it's like another content element following it.

The fix idea is to set the icon as a background image of the link with some padding on the right to have room for it. So, when wrapping the tag, "padding+background=icon" is attached to link content's last word. 

#### Checklist

- [ ] Text renders correctly


tools/api_refs/.phpdoc/template/css/custom.css.twig
```diff
@@ -394,13 +394,9 @@ figcaption {
...
 
-a.external:not(.card):after {
-    content: url(../../../../images/open-in-new.svg);
-    font-style: normal;
-    padding-left: 0px;
-    position: relative;
-    font-size: 11px;
-    top: -5px;
+a.external:not(.card) {
+    background: url(../../../../images/open-in-new.svg) no-repeat 100% -67%;
```
adriendupuis, line 398:
```suggestion
    background: url(../../../../images/open-in-new.svg) no-repeat 100% -40%;
```

mnocon (approved):
Thanks!

PR #3153, opened 2026-04-21
Added link checker

mnocon:
This PR makes it possible to check links in the documentation

## Problem, constraints, and assumptions:
- Broken links in the doc = bad
- we only care about checking the external links: MkDocs is doing a great job of checking all internal links within a project, including fragments
- doc.ibexa.co is protected by Cloudflare, which blocks all requests coming from IPs related to GitHub Actions
  - this means that it's not possible to verify links to doc.ibexa.co on CI environment

## Tool used

Lychee is used: https://lychee.cli.rs/

1) It's fast
2) Has a lot of configuration options (including ignoring certain links, **remapping URLS** (crucial), redirections, checking fragments and text fragments)
3) Has good CI integration
4) Full run locally takes about ~10 minutes

## How it works

1) Ignoring files

`.lycheeignore` is used for URLs that cannot be checked easily

2) Because links to doc.ibexa.co cannot be checked on CI (crawlers are blocked), and checking them is the main point, another solution is used: link remapping.

Links leading to doc.ibexa.co are remapped to local instances of the doc:
- dev doc 5.0 (by default uses 5.0 branch)
- dev doc 4.6 (by default uses 4.6 branch)
- user doc 5.0 (by default uses 5.0 branch)
- user doc 4.6 (by default uses 4.6 branch)
- connect main  (by default uses main branch)

I've checked links to lower versions of the doc (which cannot be checked on CI) manually, fixed them, and then added to the `.lycheeignore` file.

3) The branches can be overwritten when triggering the job from CI

The idea is that we can use the release_4.6.x branches, to make sure all the links are correct before the release

## Fixing the links

The rest of the PR fixes existing links:

- redirects are updated to use the new location
- broken links are fixed or removed, if removing is not possible

This PR is split into 2 commits:
- adding the tool: https://github.com/ibexa/documentation-developer/pull/3153/changes/deb36e33cc6194aaddc43d83e9aac30e4df84a10
- fixing the links: https://github.com/ibexa/documentation-developer/pull/3153/changes/ac2a7a4e43bc014ed14fc17ea16a76dfa11cc778

The job with broken links is here:
https://github.com/ibexa/documentation-developer/actions/runs/26101343148/job/76752887924

404 - external link broken
30x - redirects
`Cannot find file: File not found` - documentation link broken


docs/update_and_migration/migrate_to_ibexa_dxp/migrating_from_ez_publish.md
```diff
@@ -47,9 +47,9 @@ Alternatively you can use Null field type as a dummy implementation for the cust
 Using Null field type prevents errors from the Platform Stack, see [Null field type Reference](nullfield.md)
 - Rewrite custom web front end to use the new Platform/Symfony stack, see [Beginner Tutorial](beginner_tutorial.md)
 - Rewrite custom admin modules to use the new Platform/Symfony stack
-    - And if you do this while on 5.x, you can use several of the [available legacy migration features](https://doc.ez.no/display/EZP/Legacy+code+and+features) to make the new code appear in legacy admin
+    - And if you do this while on 5.x, you can use several of the [available legacy migration features](https://web.archive.org/web/20140911035812/https://doc.ez.no/display/EZP/Legacy+code+and+features) to make the new code appear in legacy admin
```
mnocon, line 50: These pages are no longer online, I feel linking to their archived versions is the right thing to do - in case someone is stuck on a legacy version

docs/release_notes/ibexa_dxp_v3.2.md
```diff
@@ -93,8 +93,6 @@ This way you can envision what content will be available in the future.
 
 Also, you can now apply new filters that are intended to help you declutter the calendar view.
 
-For more information, see [Calendar widget](https://doc.ibexa.co/projects/userguide/en/latest/publishing/advanced_publishing_options/#calendar-widget).
```
mnocon, line 96: This page is broken, no sense of ressurecting it right now

adriendupuis: If it's in 3.2 and https://doc.ibexa.co/projects/userguide/en/3.3/publishing/advanced_publishing_options/#calendar-widget is still there, that could be the fix.

mnocon: Thanks, added that in the latest commit

lychee.toml.dist
```diff
@@ -0,0 +1,190 @@
...
+
+# Exclude all private IPs from checking.
+exclude_all_private = true
+
+#############################  Local files  #########################
+
+# Required to resolve root-relative links (e.g. href="/…") found in every page into
+# file:// URLs. Those file:// URLs are then excluded from checking by the pattern in
+# the exclude list — internal links are validated by `mkdocs build --strict`.
+root_dir = "site"
```
adriendupuis, line 111: A little reminder here why the `root_dir` is `site` even if it becomes a bit more clear when reaching the "Remap" section below.
```suggestion
# `site` is the default `mkdocs build` output directory.
root_dir = "site"
```

If it's set here, why is it needed as command argument?
```
                  args: >-
                      --config lychee.toml
                      --cache
                      --cache-exclude-status "400.."
                      site
```

mnocon: It's a different setting - root_dir controls how relative links are resolved, the command line argument decided which directory is scanned.

Please see the usage example in https://lychee.cli.rs/recipes/root-dir/

lychee.toml.dist
```diff
@@ -0,0 +1,190 @@
...
+    # them when READTHEDOCS_VERSION_NAME is not set (e.g. in CI). Remap to the locally
+    # built site/ directory. These file:// URLs are then matched by the exclude pattern
+    # above and skipped — internal links are validated by `mkdocs build --strict`.
+    "https://doc\\.ibexa\\.co/en/latest/([^#]+\\.html)(#.*)?$ file://__BASE_DIR__/site/$1$2",
+    "https://doc\\.ibexa\\.co/en/latest/([^#/]+(?:/[^#/]+)*)/?(#.*)?$ file://__BASE_DIR__/site/$1/index.html$2",
+    "https://doc\\.ibexa\\.co/en/latest/$ file://__BASE_DIR__/site/index.html",
+    # userguide en/latest/ — map to the newest available clone (userdoc-5.0)
+    "https://doc\\.ibexa\\.co/projects/userguide/en/latest/([^#]+\\.html)(#.*)?$ file://__BASE_DIR__/repositories/userdoc-5.0/site/$1$2",
+    "https://doc\\.ibexa\\.co/projects/userguide/en/latest/([^#/]+(?:/[^#/]+)*)/?(#.*)?$ file://__BASE_DIR__/repositories/userdoc-5.0/site/$1/index.html$2",
+    "https://doc\\.ibexa\\.co/projects/userguide/en/latest/$ file://__BASE_DIR__/repositories/userdoc-5.0/site/index.html",
```
adriendupuis, line 157: I think we should fail on `latest`. We should always use `5.0`. `latest` will be `6.0` one day, and having `5.0` targeting `6.0` will be wrong. And maybe we already have `4.6` targeting `5.0` issues.
`userguide/en/latest` → `userdoc-5.0` remapping shouldn't be needed and we shouldn't have to maintain this block on 6.0 release.

.github/workflows/link_check.yaml
```diff
@@ -0,0 +1,113 @@
...
+                  args: >-
+                      --config lychee.toml
+                      --cache
+                      --cache-exclude-status "400.."
+                      site
+                  output: lychee-report.md
+                  jobSummary: false
+                  fail: true
+              env:
+                  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
adriendupuis, line 105: Why is the GitHub token needed to run lychee?

adriendupuis: Probably because of https://github.com/lycheeverse/lychee#github-token

lychee.toml.dist
mnocon: Nice, like the idea! Will implement

.github/workflows/link_check.yaml
```diff
@@ -0,0 +1,119 @@
...
+              uses: actions/cache@v4
+              with:
+                  path: .lycheecache
+                  key: lychee-${{ github.ref_name }}-${{ hashFiles('lychee.toml') }}
+                  restore-keys: |
+                      lychee-${{ github.ref_name }}-
+                      lychee-
+
+            - name: Check links
+              uses: lycheeverse/lychee-action@master
```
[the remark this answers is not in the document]
mnocon: This is temporary, to use 0.24.2.

Will change this to a released version once github.com/lycheeverse/lychee-action/issues/332 is fixed

lychee.toml.dist
mnocon: Looking at this, there are some legitimate use cases for `latest` in the doc, so I've kept the rule - but I've reviewed every "latest" link and replaced it with a specific version call where applicable.

This is done in github.com/ibexa/documentation-developer/pull/3153/changes/9ea1af8df480c1cd628c2adf882fe715c04db0b9

.lycheeignore
```diff
@@ -0,0 +1,165 @@
...
+^https://support\.qualifio\.com/
+^https://unsplash\.com/
+https://support.deepl.com/hc/en-us/articles/360020695820-API-key-for-DeepL-API
+^https?://(www\.)?npmjs\.com
+
+
+# AI provider consoles — redirect to login pages (302/307)
+https://aistudio\.google\.com/app/api-keys
+https://console\.anthropic\.com/login
+https://docs\.claude\.com/en/docs/about-claude/models/overview
```
adriendupuis: Please, remove `https://docs\.claude\.com/en/docs/about-claude/models/overview`

The redirection chain is 
302 https://docs.claude.com/en/docs/about-claude/models/overview → 200 https://platform.claude.com/docs/en/about-claude/models/overview

It's to be fixed in docs/ai/ai_actions/configure_ai_actions.md line 73

Notice that line 96 this redirection was fixed:
302 https://docs.anthropic.com/en/docs/about-claude/models/overview → 200 https://platform.claude.com/docs/en/about-claude/models/overview
Two different URLs where used to target the same content? Or two different pages were merged?

```suggestion
```

.lycheeignore
```diff
@@ -0,0 +1,165 @@
...
+^https?://127\.0\.0\.1
+
+# Placeholder domain used in examples
+^https?://example\.com
+
+# Known intentional redirects (302)
+^https://support\.ibexa\.co/
+^https://redocly\.com/redoc/
+^https://updates\.ibexa\.co
+^https://console\.cloud\.google\.com
```
adriendupuis: Please, group with `https://aistudio\.google\.com/app/api-keys`
Both are the same chain of redirections to https://accounts.google.com/v3/signin

adriendupuis (changes requested):

.lycheeignore
mnocon: Done in https://github.com/ibexa/documentation-developer/pull/3153/changes/44be7e193125da2d52f8116075bbe5749efce3ad

.lycheeignore
mnocon: Done in https://github.com/ibexa/documentation-developer/pull/3153/changes/44be7e193125da2d52f8116075bbe5749efce3ad

adriendupuis (approved):

PR #3165, opened 2026-04-23
Detecting and fixing incorrect YAML samples

mnocon:
This PR allows us to test the YAML code samples in the doc - both in the code_samples directory (.yaml files) and the ones embedded directly in Markdown (by extracting the ``` yaml code blocks).

There are two checks:
- YAML syntax check - makes sure that YAML is valid
- Bundle configuration check - `ibexa`, `security`, `ibexa_migrations` configuration keys

Not everything is checked - for example, Ibexa Migration files are not validated.

## Running the tests

To run the tests, execute:
``` php
composer update
composer check-yaml
```

Example output:
```
There was 1 failure:

1) Ibexa\Tests\Documentation\Yaml\YamlTest::testBundleConfigurationIsValid with data set "ibexa (docs/customer_management/cp_page_builder.md:119)" ('ibexa', [[['import', 'site', 'admin', 'corporate', 'custom_portal'], [['import', 'site'], ['site'], ['corporate', 'custom_portal']]], [[['eng-GB']], [['eng-GB'], [[12345, ['/media/', '/images/']]]]]], 'docs/customer_management/cp_p...der.md', 119)
Invalid configuration for "ibexa" in docs/customer_management/cp_page_builder.md:119 — The child config "default_siteaccess" under "ibexa.siteaccess" must be configured: Name of the default siteaccess

/Users/marek/Desktop/repos/mnocon/documentation-developer/test-yamls/tests/Yaml/YamlTest.php:82
```

Either fix the code sample, or add the error to the list of ignored lists by running:

```
composer yaml-update-baseline
```

Thats a baseline for samples that should be ignored - to put something in the baseline, run `composer yaml-update-baseline` and commit the modified `tests/yaml-validation-baseline.yaml` 

## Baseline

Baseline depends on a hash of the code sample (SHA-256), not the line number itself - the line number is there just for us to locate the code sample easily. This prevent raising the same issue every time the line numbers of the code sample changes (for example, when it's modified, new lines are added or deleted).

PR #3166, opened 2026-04-24
Added testing of embedded PHP snippets

mnocon:
This PR makes it possible to run code quality checks (PHPStan, CS-Fixer, Rector, Deptrac) on embedded PHP code samples.

It build upon the code samples extraction mechanism added in https://github.com/ibexa/documentation-developer/pull/3165

Thanks to this, we no longer need to extract the code samples to separate files - they can stay within the Markdown **files, making it easier to work with them, while keeping the advantage of the code quality tools.**

### How it works

Before a code quality tool is run, the code samples are extracted from Markdown files.

They are not extracted if:
- the code sample uses one of the include_file/include_code functions
- the code sample is marked with `{skip-validation}`

The rest of them are extracted to `code_samples/_inline_php` for the tools to pick up.

When the tool can automatically fix the issue (CS-Fixer, Rector), the changes are synced back to the Markdown files.

All the existing inline code samples are marked with `{skip-validation}`, a special attribute to disable the code sample checking.

This keeps this PR small and allow us using this for new content - the fixing of code samples will be done in a follow-up (https://github.com/ibexa/documentation-developer/pull/3206)


tools/extract-inline-php.php
```diff
@@ -0,0 +1,140 @@
...
+            unlink($item->getRealPath());
+        }
+    }
+
+    rmdir($dir);
+}
+
+// Clean up the output directory from the previous run to avoid stale files
+// (e.g. from docs that were renamed, moved, or had snippets removed).
+removeDirectory(OUTPUT_DIR);
```
adriendupuis: I feel like this `OUTPUT_DIR` could be used for cache but I don't see how. The code must be extracted and the hash computed to know that the extraction file already exist, so the time is lost anyway. Afterward, there the `php-cs-fixer.cache` that might remember already checked files.

adriendupuis: No error to catch there if the directory can't be removed?

adriendupuis: Change made with Copilot

> Hey,
> In @extract-inline-php.php please modify removeDirectory so it stops and returns false as soon as inside it a rmdir or unlink returns false.

```suggestion
/**
 * Recursively removes all files and directories under $dir, then removes $dir itself.
 */
function removeDirectory(string $dir): bool
{
    if (!is_dir($dir)) {
        return true;
    }

    $iterator = new RecursiveIteratorIterator(
        new RecursiveDirectoryIterator($dir, RecursiveDirectoryIterator::SKIP_DOTS),
        RecursiveIteratorIterator::CHILD_FIRST,
    );

    foreach ($iterator as $item) {
        if ($item->isDir()) {
            if (!rmdir($item->getRealPath())) {
                return false;
            }
        } else {
            if (!unlink($item->getRealPath())) {
                return false;
            }
        }
    }

    return rmdir($dir);
}

// Clean up the output directory from the previous run to avoid stale files
// (e.g. from docs that were renamed, moved, or had snippets removed).
if (!removeDirectory(OUTPUT_DIR)) {
    fwrite(STDERR, 'ERROR: Could not clean directory: ' . OUTPUT_DIR . "\n");
    exit(2);
}
```

adriendupuis (approved):

PR #3167, opened 2026-04-27
Avoid redirects from master to 5.0

mnocon:
Target: 5.0 only

Adjusting the links:
- doc doesn't have a `master` branch anymore
- after creating the stable branches last week, `5.0` corresponds to the 5.0 version (and `main/master` no longer exists)

Once we merge this, I'll regenerate the PHP API reference to fix the links there.


dabrt:
Could you use this opportunity to fix broken links to mapping references in [4.0 deprecations article](https://doc.ibexa.co/en/5.0/release_notes/ibexa_dxp_v4.0_deprecations):
https://github.com/ibexa/compatibility-layer/tree/main/src/bundle/Resources/mappings ->https://github.com/ibexa/compatibility-layer/tree/4.0/src/bundle/Resources/mappings
https://github.com/ibexa/compatibility-layer/blob/main/src/bundle/Resources/mappings/services-to-fqcn-map.php -> https://github.com/ibexa/compatibility-layer/blob/4.0/src/bundle/Resources/mappings/services-to-fqcn-map.php
https://github.com/ibexa/compatibility-layer/blob/main/src/bundle/Resources/mappings/symfony-service-tag-name-map.php -> https://github.com/ibexa/compatibility-layer/blob/4.0/src/bundle/Resources/mappings/symfony-service-tag-name-map.php
Not that we have customers for this version, but let's clear this out, too.

dabrt (approved):

mnocon:
Sure, fixed in https://github.com/ibexa/documentation-developer/pull/3167/commits/9fd16a854f5927ec9939bc0369bcc38db046295e

PR #3176, opened 2026-05-05
Added continue-on-error

mnocon:
This PR makes all the Code Quality jobs run all time, so Rector and Deptrac are not blocked by PHPStan failures - and we can see all the issues during review.

Right now (without these changes), PHPStan errors block future jobs and we don't see the full picture.

The job will still be marked as failed if PHPStan fails while Deptrac and Rector are green.


julitafalcondusza (approved):

adriendupuis (approved):
👍

PR #3180, opened 2026-05-05
Described disabling counting the results

mnocon:
Target: 5.0, 4.6

Based on feedback provided on community Slack

### DIFF
--- a/docs/search/search_api.md
+++ b/docs/search/search_api.md
@@ -67,7 +67,9 @@ By default, a search query also counts all matching results.
 If you don't need the total count, set `performCount` to `false` on [`Query`](/api/php_api/php_api_reference/classes/Ibexa-Contracts-Core-Repository-Values-Content-Query.html) or [`LocationQuery`](/api/php_api/php_api_reference/classes/Ibexa-Contracts-Core-Repository-Values-Content-LocationQuery.html) to improve performance, especially for large result sets.
 
 ``` php
-[[= include_code('code_samples/api/public_php_api/src/perform_count.php') =]]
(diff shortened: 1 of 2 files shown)


konradoboza (approved):

bnowak (approved):

docs/search/search_api.md
```diff
@@ -61,6 +61,17 @@ $output->writeln($result->getName());
     $query->limit = 100;
     ```
 
+#### Disable result count
+
+By default, a search query also counts all matching results.
+If you don't need the total count, set `performCount` to `false` on [`Query`](/api/php_api/php_api_reference/classes/Ibexa-Contracts-Core-Repository-Values-Content-Query.html) or [`LocationQuery`](/api/php_api/php_api_reference/classes/Ibexa-Contracts-Core-Repository-Values-Content-LocationQuery.html) to improve performance, especially for large result sets.
+
+``` php
+[[= include_code('code_samples/api/public_php_api/src/perform_count.php') =]]
```
adriendupuis: The example is a bit long for what it documents.

```suggestion
[[= include_code('code_samples/api/public_php_api/src/perform_count.php', 8, 10) =]]

[[= include_code('code_samples/api/public_php_api/src/perform_count.php', 14, 16) =]]
```

adriendupuis (changes requested):
The example is a bit long. We could hightlight the two lines `->performCount = false;`

But it's strange to execute an empty request, the example is a bit long for what it documents. I would focus more on the query setting and forget the surrounding. Previous and following full examples are there for that.

adriendupuis (approved):

PR #3181, opened 2026-05-05
Added a table of built-in attributes

mnocon:
Adding a table for built-in attributes.

I've noted this down when working on the doc for Quable, as the attribute type identifier is needed when overiding the templates:
see https://doc.ibexa.co/en/latest//product_catalog/customize_product_attribute_templates for an example.

And we didn't have these attribute identifiers in the doc yet.

Note: measurement is removed from the list, as it was removed in v5 (https://github.com/ibexa/measurement/pull/93).

### DIFF
--- a/docs/product_catalog/products.md
+++ b/docs/product_catalog/products.md
@@ -38,7 +38,7 @@ The following attribute types are available:
 | Name | Identifier | Description |
 |------|------------|-------------|
 | Checkbox | `checkbox` | Boolean attribute with a true/false value. |
-| Color | `color` | Stores a color value as a hex code. |
+| Color | `color` | Color value stored as a hex code. |
 | [Date and time](date_and_time.md) | `datetime` | Date and time value with configurable accuracy levels. |
 | Float | `float` | Decimal number value. |
 | Integer | `integer` | Integer number value. |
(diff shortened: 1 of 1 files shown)


dabrt:
Would it make sense to put the two tables ("Product attribute types" and "Quable attribute types") next to each other?

mnocon:
> Would it make sense to put the two tables ("Product attribute types" and "Quable attribute types") next to each other?

As discussed on Slack, I'd keep it as is for now

docs/product_catalog/products.md
```diff
@@ -35,14 +35,15 @@ Typical product attribute examples are: length, weight, color, format, and more.
...
-- [date and time](date_and_time.md)
-- float
-- integer
-- measurement (`measurement_range` and `measurement_single`)
-- selection
-- [symbol](symbol_attribute_type.md)
+| Name | Identifier | Description |
+|------|------------|-------------|
+| Checkbox | `checkbox` | Boolean attribute with a true/false value. |
+| Color | `color` | Stores a color value as a hex code. |
```
julitafalcondusza: Just a small suggestion for consistency (this is the only one that starts with a verb).
```suggestion
| Color | `color` | Color value stored as a hex code. |
```

julitafalcondusza (approved):

PR #3188, opened 2026-05-07
Document best practices when upgrading Fastly

dabrt:
| Question      | Answer |
| ------------- | --- |
| JIRA Ticket   | N/A |
| Versions      | 5.0 |
| Edition       | all |

Document best practices when upgrading Fastly

#### Checklist

- [x] Text renders correctly

### DIFF
--- a/docs/infrastructure_and_maintenance/cache/http_cache/best_practices_fastly.md
+++ b/docs/infrastructure_and_maintenance/cache/http_cache/best_practices_fastly.md
@@ -1,0 +1,156 @@
+---
+description: Recommendations for validating Fastly across development, update, and production.
+month_change: yes
+---
+
+# Best practices for testing and updating Fastly
+
+You can use Fastly in caching, performance, and request handling in [[= product_name =]] on-premise environments.
+However, it also introduces a dependency layer that can surface issues that impact authentication and caching behavior only in environments where it is enabled.
+
+Usually, Fastly isn't used during development, which means configuration or behavior mismatches are often discovered during staging or production rollout.
+To avoid production issues, ensure Fastly is included in development and staging workflows, and validate configuration changes thoroughly before moving to production.
+
+This guide explains how to:
+
+- Use Fastly safely during development and testing
+- Upgrade Fastly configuration for [[= product_name =]] v5.0
+- Identify and avoid common pitfalls
+- Validate your setup before going to production
+
+For detailed configuration steps, VCL structure, and CLI usage, [Configure and customize Fastly](https://doc.ibexa.co/en/5.0/infrastructure_and_maintenance/cache/http_cache/fastly/#configure-and-customize-fastly).
+
+## Use Fastly during development and testing
+
+Fastly behaves as an application layer in front of [[= product_name =]].
+Because of this, caching rules, cookie handling, authentication flows may behave differently compared to a direct origin setup.
+
+Introducing Fastly only in production may cause issues to surface at a late stage.
+
+### Recommended testing approaches
+
+Depending on your infrastructure, you can:
+
+- Use a dedicated Fastly service for non-production environments
+- Connect staging environments to a controlled Fastly configuration
+- Use DDEV-based setups when applicable
+
+For DDEV integration setup details, see [Fastly integration with DDEV environments](https://doc.ibexa.co/en/5.0/infrastructure_and_maintenance/clustering/clustering_with_ddev/#fastly).
+
+#### What to validate early
+
+When you enable Fastly in non-production environments, focus on testing the core authentication and caching behaviors, as outlined in the [Pre-production checklist](#preproduction-checklist) to identify configuration mismatches before they reach the staging phase.
+
+### Simulate Fastly in deployment
+
+Because Fastly can't be fully reproduced locally, you could use a simulation strategy to catch configuration mismatches early:
+
+#### Stage 1: Local smoke testing with DDEV & Varnish
+
+Fastly is built on Varnish, therefore you could use the DDEV Varnish integration to catch raw VCL, caching header, and session cookie issues locally:
+
+1. Enable the Varnish service in DDEV:
+
+    `ddev get ddev/ddev-varnish`
+
+1. Use the [[= product_name =]] v5.0 Varnish VCL [base configuration](https://github.com/ibexa/http-cache/blob/v[[= latest_tag_5_0 =]]/docs/varnish/vcl/varnish7.vcl) as your local proxy.
+1. Test authentication, back office operations, and session persistence.
+
+This stage helps identify low-level caching and cookie handling issues early.
+
+#### Stage 2: Staging with temporary Fastly domains
+
+Before you move on to production, clone your production Fastly service to validate the infrastructure layer safely:
+
+1. Clone the service: In the Fastly UI, clone your active production service to create a temporary, test version.
+1. Reconfigure origins: Point the backend to your staging environment instead of production.
+1. Configure temporary domains: Use a temporary Fastly-provided domain, for example, `staging-yourdomain.global.ssl.fastly.net`, or a dedicated test subdomain.
+1. Run validation: Access the staging site through the Fastly URL and verify header handling, cookie behavior, session persistence and back office functionality.
+
+This stage allows you to test real Fastly behavior safely before moving to production.
+
+## Update Fastly configuration safely
+
+When you upgrade [[= product_name =]], do not reuse legacy VCL files.
+Always start from the new default Fastly configuration for the target version, and carefully review your Fastly configuration customizations as you reapply them, because changes between versions may affect authentication and caching behavior.
+
+Focus on session and authentication handling, cookie-based logic, cache bypass rules, and header manipulation rules. Even small changes in these areas can affect login behavior or cache correctness.
+
+## Common Fastly issues during update
+
+The following issues may appear when Fastly is used as the view cache for [[= product_name =]] and you perform an update.
+
+| Issue | Symptoms | Common causes | Recommendation areas |
+|------|----------|---------------|-------|
+| Authentication and session issues | - Back office login fails<br>- Users are logged out unexpectedly<br>- Sessions are not persistent | - Session cookie name prefix changes break VCL rules<br>- VCL still references outdated cookie identifiers<br>- Cookie filtering removes required authentication cookies | Cookie and session handling |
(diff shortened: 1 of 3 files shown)


docs/infrastructure_and_maintenance/cache/http_cache/best_practices_fastly.md
```diff
@@ -0,0 +1,156 @@
...
+
+Because Fastly can't be fully reproduced locally, you could use a simulation strategy to catch configuration mismatches early:
+
+#### Stage 1: Local smoke testing with DDEV & Varnish
+
+Fastly is built on Varnish, therefore you could use the DDEV Varnish integration to catch raw VCL, caching header, and session cookie issues locally:
+
+1. Enable the Varnish service in DDEV:
+
+    `ddev get ddev/ddev-varnish`
```
adriendupuis: Somehow, it duplicates https://doc.ibexa.co/en/5.0/infrastructure_and_maintenance/clustering/clustering_with_ddev/#varnish

There is a deprecation message with `ddev get`

```suggestion
    `ddev add-on get ddev/ddev-varnish`
```

dabrt: It does, but here it is a different context: "How can I set up a staging environment that can simulate the real deal, for testing purposes" and there it is about setting up a real deal.

docs/infrastructure_and_maintenance/cache/http_cache/best_practices_fastly.md
```diff
@@ -0,0 +1,156 @@
...
+1. Use the [[= product_name =]] v5.0 Varnish VCL [base configuration](https://github.com/ibexa/http-cache/blob/v[[= latest_tag_5_0 =]]/docs/varnish/vcl/varnish7.vcl) as your local proxy.
+1. Test authentication, back office operations, and session persistence.
+
+This stage helps identify low-level caching and cookie handling issues early.
+
+#### Stage 2: Staging with temporary Fastly domains
+
+Before you move on to production, clone your production Fastly service to validate the infrastructure layer safely:
+
+1. Clone the service: In the Fastly UI, clone your active production service to create a temporary, test version.
```
vidarl, line 65: Fastly UI is not available for Ibexa Cloud customers, that means 99% of our customers that uses Fastly. So not sure how applicable this is..
It should be possible to create a free account and fastly and use that, but cloning existing production is not an an option for most customers

AFAIK, all cloud customers that are entitled to Fastly, also have a staging environment where Fastly is/should be enabled. One of the ideas with this staging environment is to test for instance Fastly integration

docs/infrastructure_and_maintenance/cache/http_cache/best_practices_fastly.md
```diff
@@ -0,0 +1,156 @@
...
+1. Reconfigure origins: Point the backend to your staging environment instead of production.
+1. Configure temporary domains: Use a temporary Fastly-provided domain, for example, `staging-yourdomain.global.ssl.fastly.net`, or a dedicated test subdomain.
+1. Run validation: Access the staging site through the Fastly URL and verify header handling, cookie behavior, session persistence, and back office functionality.
+
+This stage allows you to test real Fastly behavior safely before moving to production.
+
+## Update Fastly configuration safely
+
+When you upgrade [[= product_name =]], do not reuse legacy VCL files.
```
vidarl, line 74: I think we should mentioned explicitly in the upgrade docs if fastly .vcl files changes ( they don't change that often ). The burden on checking if they change should be on us, not every customer IMO

docs/infrastructure_and_maintenance/cache/http_cache/best_practices_fastly.md
```diff
@@ -0,0 +1,156 @@
...
+| Incorrect caching behavior | - Logged-in users see cached anonymous content<br>- Personalized content leaks between users | - Missing cache bypass rules for authenticated requests<br>- Incorrect cache key variation<br>- Cookies not properly considered in cache logic | Caching strategy + Cookie and session handling |
+| Over-aggressive cache bypass | - Low cache hit ratio<br>- Unexpected cache misses | - Too many cookies forwarded to backend<br>- Broad “pass” rules triggered too often | Caching strategy |
+| Environment mismatch issues | - Works in dev, fails in staging/production<br>- Issues only appear when Fastly is enabled | - Fastly not part of development setup<br>- Differences in configuration between environments | Configuration strategy |
+
+To mitigate these issues, follow these recommendations and best practices:
+
+#### Cookie and session handling
```
vidarl, line 92: In this chapter you are very specific on a problem that occurs if you don't upgrade .vcl files when upgrading from 4.6 to 5.0. And I feel the recommendations given below avoids the heart of the problem. Client needs to upgrade .vcl files.. ( they have also changed names from 4.6 to 5.0).
AFAIK, we only mentioned that one one line, here : https://doc.ibexa.co/en/5.0/update_and_migration/from_4.6/update_to_5.0/#http-cache
And we don't say how to do it... I think this would be the place to start ?

docs/infrastructure_and_maintenance/cache/http_cache/best_practices_fastly.md
```diff
@@ -0,0 +1,156 @@
...
+
+The session cookie name prefix changed from `eZSESSID{...}` in earlier versions to `IBX_SESSION_ID{...}` in v5.0, while remaining SiteAccess-specific.
+This change can break VCL rules that rely on fixed cookie name patterns.
+
+- Ensure required authentication cookies are not stripped.
+- Validate that session cookies correctly trigger cache bypass where needed.
+- Review all cookie-based conditions in VCL and update outdated cookie references.
+- Avoid relying on exact cookie names. Use pattern matching where possible to reduce the impact of cookie naming changes.
+- Consider temporary support for both old and new cookie identifiers during transition.
```
vidarl, line 101: I think I would remove all those recommendations and instead give user a way of comparing current deployed .vcls vs the one in vendor/fastly/fastly/

Unfortunately, there is no straight way for doing that, but [this doc](https://doc.ibexa.co/en/5.0/infrastructure_and_maintenance/cache/http_cache/fastly/#get-diff-between-two-versions) can help.
However, diffing the downloaded versions vs our dist vcl will never be exactly the same ( there are various snippets etc as well which will be included in the downloaded version ), but diffing the two would help a trained eye.

For an untrained-eye, I think user needs to create a new version in Fastly service, re-upload all the vcls and snippets to that version. Then download the `active` version and mentioned the newer version and then compare.

docs/update_and_migration/from_4.6/update_to_5.0.md
```diff
@@ -1091,6 +1091,14 @@ Depending on your reverse proxy, you'll find them in the following directories:
 - Varnish: `vendor/ibexa/http-cache/docs/varnish/vcl/`
 - Fastly: `vendor/ibexa/fastly/fastly/`
 
+!!! note "Fastly considerations"
+
+    Version updates may introduce changes (for example, to cookies or caching behavior) that could require modifications in Fastly/VCL configuration.
```
vidarl, line 1096: There is no "may introduce changes" here. We know for a fact that there are changes. User *must* update vcl configuration. Only reviewing them is not enough

vidarl (changes requested):

docs/infrastructure_and_maintenance/cache/http_cache/best_practices_fastly.md
vidarl: Actually, I think you may download each individally vcl using:
`fastly service vcl custom describe --service-id <SERVICE_ID> --version <VERSION_NUMBER> --name <VCL_NAME>
`

If there is something similar for snippets too, then it should be possible to compare our distributed .vcl files vs those installed. In this conntext, user should also check that the ibexa_main.vcl file is marked as `main`, ref:  https://doc.ibexa.co/en/5.0/infrastructure_and_maintenance/cache/http_cache/fastly/#list-custom-vcl-files-for-specific-version

If ibexa_user_hash.vcl is marked as `main`, the configuration will not work

vidarl:
Hi. Regarding checking if vcl and snippets in fastly service is in sync with current Ibexa version that is installed, this can be done by `curl` and `diff` : 

Set some variables used by the later commands:
```
API=https://api.fastly.com
H="Fastly-Key: $FASTLY_API_TOKEN"
V=$(curl -sf -H "$H" "$API/service/$FASTLY_SERVICE_ID/version" | jq -r '.[] | select(.active == true) | .number')
```

For the  two .vcl files:
```bash
for f in ibexa_main.vcl ibexa_user_hash.vcl; do
    curl -sf -H "$H" "$API/service/$FASTLY_SERVICE_ID/version/$V/vcl/$f" | jq -r .content > "/tmp/live_$f"
    echo "=== $f ==="
    diff -u "vendor/ibexa/fastly/fastly/$f" "/tmp/live_$f" && echo "identical"
done
```

For the snippets, first just list all of them:

```bash
curl -sf -H "$H" "$API/service/$FASTLY_SERVICE_ID/version/$V/snippet" \
    | jq -r '.[] | "\(.name)\t type=\(.type)\t priority=\(.priority)"'
```
For the `snippet_re_enable_shielding.vcl` snippet, select by name

```bash
curl -sf -H "$H" "$API/service/$FASTLY_SERVICE_ID/version/$V/snippet" \
    | jq -r '.[] | select(.name | test("shielding"; "i")) | .content' > /tmp/live_shielding.vcl
diff -u vendor/ibexa/fastly/fastly/snippet_re_enable_shielding.vcl /tmp/live_shielding.vcl
```

`ibexa_image_optimizer.vcl` needs its own check, because both its presence and its absence can be the error. The
snippet has no fixed name — the doc page for `Fastly Image Optimizer`
installs it as `Ibexa Image Optimizer` — so select it by the code it contains rather than by name. The
`x-fastly-imageopto-api` assignment is unique to it:

```bash
curl -sf -H "$H" "$API/service/$FASTLY_SERVICE_ID/version/$V/snippet" \
    | jq -r '.[] | select(.content | test("x-fastly-imageopto-api")) | .content' \
    > /tmp/live_image_optimizer.vcl

if [ -s /tmp/live_image_optimizer.vcl ]; then
    diff -u vendor/ibexa/fastly/fastly/ibexa_image_optimizer.vcl /tmp/live_image_optimizer.vcl \
        && echo "identical"
else
    echo "no Image Optimizer snippet installed"
fi
```

vidarl:
We should also mentioned to check that `MAIN` flag is set to true for `ibexa_main.vcl`:

```bash
fastly vcl custom list --version=active
SERVICE ID  VERSION  NAME                 MAIN
<id>        3        ibexa_main.vcl       true
<id>        3        ibexa_user_hash.vcl  false
```

PR #3208, opened 2026-05-17
Unified admonitions

mnocon:
Target: v5, 4.6

Our admonition usage is not consistent, we have a couple of these which compete with each other - and the `warning` and `info` ones use default MkDocs styling, not our custom one.

In this PR, I'm converting:

- `info` -> `note`
- `warning` -> `caution`

After this change, the usage is:

 - note: 350 total
 - tip: 194 tota
 - caution: 133 total

As a follow up, I'll add a vale rule checking these, to make sure we stay consistent.

### DIFF
--- a/docs/personalization/legacy_recommendation_api.md
+++ b/docs/personalization/legacy_recommendation_api.md
@@ -72,7 +72,7 @@ It fetches 8 recommendations for user Smith, who is watching the item 123 and th
 The recommendation request returns a list of item IDs that are JSON, JSONP or XML-formatted.
 The result can be integrated into any webpage by using some lines of script code. 
 
-!!! tips
+!!! tip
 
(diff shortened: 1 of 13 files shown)


docs/personalization/legacy_recommendation_api.md
dabrt, line 1: This file is not exposed in our table of contents. I know it was left behind to support some of our clients that used SS-based perso, but do we need it any more?

https://doc.ibexa.co/en/latest/personalization/legacy_recommendation_api/

dabrt (approved):

docs/personalization/legacy_recommendation_api.md
mnocon: Let's keep it for now, at some point we'll rework the whole Personalization section

PR #3209, opened 2026-05-17
[PHP] Migrated include_file to include_code. Added remove_indent for partial includes

mnocon:
Target: v5

For PHP code samples, In code blocks with only a single include_file call:

- include_code is used instead
- if the include is a partial one, remove_indent=True is added to the call

Multiple include_* calls within a single code blocks are not covered here, as they require more attention and will be done in a follow-up

**To review this, please use HTML report I've attached on Slack - this makes the review much easier.**

It contains the changes before and after, and a diff showing what has changed.
Please use the provided HTML file for reviewing - it's generated locally using the existing code samples usage job, this should make the review easier.

Screenshot 2026-05-18 at 11 55 42

### DIFF
--- a/.github/workflows/preview_comment.yaml
+++ b/.github/workflows/preview_comment.yaml
@@ -28,7 +28,7 @@ jobs:
 
             -   name: Create comment for changed files
                 run: |
-                    file_limit=100
+                    file_limit=150
                     build_url="https://ez-systems-developer-documentation--${{ github.event.pull_request.number }}.com.readthedocs.build/${{inputs.project}}en/${{ github.event.pull_request.number }}/"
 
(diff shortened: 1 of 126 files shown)


dabrt (approved):

dabrt:
The indent removal seems to be working OK.
Have you tried testing snippets that contain multiple pieces of one file? Would we need to leave `remove_indent=True` out to not break the code?

mnocon:
> Have you tried testing snippets that contain multiple pieces of one file? Would we need to leave remove_indent=True out to not break the code?

Yes, that's exactly why I didn't touch them in this PR - for them, each usage has to be carefully checked and the indent level must be adjusted manually to fit the whole code sample.

I've decided that it's not worth covering in this PR, we might do a follow-up in the future.

PR #3238, opened 2026-06-01
Replace remote provider's address

dabrt:
| Question      | Answer |
| ------------- | --- |
| JIRA Ticket   | N/A |
| Versions      | all |
| Edition       | all |

Replace ~~random cat image generator~~ remote provider's address

#### Checklist

- [x] Text renders correctly


adriendupuis (approved):
So, the kitten has grown up

I saw there is also https://www.placekittens.com/ but it sucks, the pictures are deformed while with placecats.com they're nicely cropped or something.

reithor (approved):
+1 👍

PR #3272, opened 2026-07-03
Described stamps for Ibexa Messenger

mnocon:
Target: 5.0 only for now.

Doc for:

- https://github.com/ibexa/messenger/pull/15
- https://github.com/ibexa/messenger/pull/17
- https://github.com/ibexa/messenger/pull/18

Two things:

1) I did not describe DeduplicateStamp, as I believe people should use the one from Symfony (https://symfony.com/doc/current/messenger.html#message-deduplication) and we should "hide" (and, hopefully, remove in 6.0) our copy
2) The deprecation must be mentioned in the 6.0 update guide, and the SudoStamp section must be reworked for this version - I'll create a separate PR to 6.0 next week when we have the 6.0 branches ready

Follow-ups:
- mention deprecation in update guide - done in github.com/ibexa/documentation-developer/pull/3279
- remove the sentence that SudoStamp is always attached - done in https://github.com/ibexa/documentation-developer/pull/3285
- remove the warning about the deprecation from the article - done in https://github.com/ibexa/documentation-developer/pull/3285

PR #3279, opened 2026-07-06
Added deprecations page for Cohesivo 6.0

mnocon:
Target: 4.6, 5.0, 6.0

TODO:

- Add link to DeduplicateStamp doc section once https://github.com/ibexa/documentation-developer/pull/3317 is merged

Adding deprecations page for the upcoming Cohesivo 6.0, to let Partners know about the upcoming changes ahead of time.

For now, the following changes are covered:

- https://github.com/ibexa/http-cache/pull/82 (http cache)
- https://github.com/ibexa/rector/pull/49 (new rule)
- https://github.com/ibexa/documentation-developer/pull/3272 (messenger changes)

As a new Rector rule is added, there's a need for dedicated Rector page:

I've extracted the exsiting content (from update_to_5.0.md and package's README) into a separate page we can refer people to.

As this might be the first time people hear about Cohesivo, I've added 2 links explaining the business reasoning.

Status: this needs to be merged with the upcoming release, as this is when the Rector rule becomes available.

Link checker results should be ignored, there's a known issue here (fixed in https://github.com/ibexa/documentation-developer/pull/3280 )

Companion PRs:

- https://github.com/ibexa/documentation-developer/pull/3286
- https://github.com/ibexa/documentation-developer/pull/3285


GrabowskiM (approved):

docs/release_notes/cohesivo_v6.0_deprecations.md
```diff
@@ -0,0 +1,36 @@
...
+| N/A | `\Ibexa\HttpCache\ResponseTagger\Delegator\DispatcherTagger`  | With `kernel.debug` enabled, [`DispatcherTagger`](content_aware_cache.md#dispatchertagger) will throw an exception when you pass an unsupported value instead of silently ignoring it. |
+| v5.0.7 | <nobr>`\Ibexa\Contracts\HttpCache\ResponseTagger\ResponseTagger::supports`</nobr> | Method added to the interface. All implementations must specify [the value they support for tagging](content_aware_cache.md#delegator-and-value-taggers). |
+
+### ibexa/messenger
+
+| Deprecated since | Entity                                                                          | Change                                              |
+| --- |------------------------------------------------------------------------------------|-----------------------------------------------------------|
+| v5.0.9 | [`SudoStamp`](background_tasks.md#sudostamp)     | No longer attached automatically to every dispatched message. For messages that should be processed without taking permissions into account, always attach the SudoStamp manually.                                             |
```
bnowak: All other Entity table entries use Fully Qualified  Class Names, this one is not. For consistency we could use its long version as well: `\Ibexa\Contracts\Messenger\Stamp\SudoStamp`.

docs/resources/rector.md
```diff
@@ -0,0 +1,335 @@
...
+
+[[= product_name_base =]] Rector is an optional package based on [Rector](https://getrector.com/) that comes with additional rule sets for working with [[= product_name =]] code.
+Use it to get rid of PHP and JavaScript code deprecations and prepare your project for the next major release.
+
+## Installation
+
+Add the Composer dependency:
+
+``` bash
+composer require --dev ibexa/rector:[[= latest_tag_5_0 =]]
```
bnowak: I wouldn't point it to any specific version, as that would lock/block any potential future updates of that package. If we leave only `composer require --dev ibexa/rector` - composer will install its latest available matching version. This way, the client will be able to update that package too with regular `composer update` 😉

bnowak (approved):
Shouldn't it be merged into the lowest version (4.6) and be merged-up to all higher versions? I don't know what rules do we have regarding documentation versioning?

konradoboza (approved):
Looks good! Do we have a plan to make some more noise about this page to create a habit it should be checked from time to time for the new additions?

docs/resources/rector.md
```diff
@@ -0,0 +1,335 @@
...
+vendor/bin/rector --dry-run
+```
+
+Once you're satisfied with the proposed changes, apply them by running:
+
+``` bash
+vendor/bin/rector
+```
+
+## Refactor JavaScipt code
```
wiewiurdp: I think it should be 'JavaScript' ;)

wiewiurdp (approved):

mnocon:
> Shouldn't it be merged into the lowest version (4.6) and be merged-up to all higher versions? I don't know what rules do we have regarding documentation versioning?

For versioning we don't use a flow based on git merges, instead we squash the commits into a single one when merging and then cherry-pick it manully to the appropriate versions. I know it's a bit unusual, but it works 😉 

Suggestions applied in https://github.com/ibexa/documentation-developer/pull/3279/changes/1fda0e83f550b3143cfdbfea359e9f7561615eb9, thank you!

> Looks good! Do we have a plan to make some more noise about this page to create a habit it should be checked from time to time for the new additions?

Yes, the plan is to mention this page in every release notes when we add a new deprecation to this list, to raise awareness around it.

docs/resources/rector.md
mnocon: Fixed in https://github.com/ibexa/documentation-developer/pull/3279/changes/787c6b8aac90c378c9663ecb18c7e231decf3df8 , thanks 🙈

docs/release_notes/cohesivo_v6.0_deprecations.md
```diff
@@ -0,0 +1,38 @@
...
+
+!!! note "Cohesivo v6.0 isn't released yet"
+
+    This page is published ahead of the Cohesivo v6.0 release to give you time to prepare your code for the upcoming changes.
+
+    As the work on Cohesivo 6.0 is in progress, this page **isn't exhaustive and will evolve with time**.
+
+As announced during Ibexa Summit 2026, [Ibexa DXP will be renamed to Cohesivo](https://www.ibexa.co/blog/redefining-the-dxp-from-execution-to-orchestration) to support the new [orchestration platform approach](https://www.ibexa.co/blog/the-orchestration-era).
+
+To learn more about the new brand, visit the [Cohesivo official site](https://cohesivo.com).
```
mnocon, line 21: Discussed the sentences above with Kamil - it's ok, and improvements to the cohesivo.com site will be released on 16.06 (2 days before the next planned release when I'd like to merge this)

docs/resources/rector.md
```diff
@@ -0,0 +1,335 @@
...
+
+- `"ez-form-error": "ibexa-form-error"` - changes all `ez-form-error` occurrences to `ibexa-form-error`
+
+#### Complete plugin configuration
+
+When not using the shorthand configuration, the following options are available:
+
+- `"to": "ibexa-selection-settings"` - specifies the new value
+- `"regexp": true/false` - use regexp to find the matching values. Use capture groups to reuse parts of the original value in the new value
+- `"exactMatch": true` - replace matching values only when the whole value is matched. Using the example configuration, `ez-selection-settings__field` would not be replaced as it doesn't match `ez-selection-settings` exactly
```
julitafalcondusza: For consistency:
```suggestion
- `"regexp": true/false` - uses regexp to find the matching values. Use capture groups to reuse parts of the original value in the new value
- `"exactMatch": true` - replaces matching values only when the whole value is matched. Using the example configuration, `ez-selection-settings__field` would not be replaced as it doesn't match `ez-selection-settings` exactly
```

docs/resources/rector.md
```diff
@@ -0,0 +1,335 @@
...
+##### Example output
+
+| Before | After |
+|---|---|
+| `class EzBooleanValidator extends eZ.BaseFieldValidator` | `class IbexaBooleanValidator extends ibexa.BaseFieldValidator` |
+| `const EZ_INPUT_SELECTOR = 'ezselection-settings__input';` | `const IBEXA_INPUT_SELECTOR = 'ezselection-settings__input';` |
+
+#### Rename string values
+
+This plugin changes any string value except translations. You can use it to transform selectors and other values.
```
julitafalcondusza:
```suggestion
This plugin changes any string value except translations.
You can use it to transform selectors and other values.
```

docs/resources/rector.md
```diff
@@ -0,0 +1,335 @@
...
+##### Example output
+
+| Before | After |
+|---|---|
+| `const SELECTOR_FIELD = '.ez-field-edit--ezboolean';` | `const SELECTOR_FIELD = '.ibexa-field-edit--ezboolean'` |
+| `const SELECTOR_FIELD = '.ibexa-field-edit--ezboolean';` | `const SELECTOR_FIELD = '.ibexa-field-edit--ibexaboolean';` |
+
+#### Rename translation IDs
+
+This plugin allows to change translation ids.
```
julitafalcondusza:
```suggestion
This plugin allows to change translation IDs.
```

docs/resources/rector.md
```diff
@@ -0,0 +1,335 @@
...
+
+#### Rename icons names used in getIconPath method
+
+This plugin allows you to rename any icon name that is passed as an argument to the `getIconPath` method.
+
+Identifier: `ibexa-rename-icons`
+
+##### Configuration example
+
+In this plugin, the `exactMatch` default value is set `true` when using the shorthand expression.
```
julitafalcondusza:
```suggestion
In this plugin, the `exactMatch` default value is set to `true` when using the shorthand expression.
```

julitafalcondusza (changes requested):
Please check my comments.

julitafalcondusza (approved):

PR #3292, opened 2026-07-15
Extracted Markdown converter to a separate package

mnocon:
Extracted the Markdown converter so that it can be reused in the user doc.

Follow-up to github.com/ibexa/documentation-developer/pull/3161
User doc PR: https://github.com/ibexa/documentation-user/pull/414


tools/llms_txt/hooks.py
```diff
@@ -0,0 +1,92 @@
+"""
```
mnocon, line 1: Moved from hooks.py with minimal changes

tools/llms_txt/llmstxt_preprocess.py
```diff
@@ -0,0 +1,550 @@
+"""
```
mnocon, line 1: Same here

requirements.txt
```diff
@@ -8,3 +8,4 @@ mkdocs-redirects==1.2.2
 mkdocs-autolinks-plugin==0.7.1
 Jinja2==3.1.6
 mkdocs-llmstxt
+-e .
```
mnocon, line 11: Explanation from Claude for this change:

```
Prompt: Explain the "-e ." change

The -e . line in requirements.txt (requirements.txt:11 after the diff) tells pip: install the package defined by the pyproject.toml in this directory, in "editable" mode.

Here's why it's there:

Why a package install is needed at all

The refactor made hooks.py and friends into thin shims:

from llms_txt.hooks import on_config, on_page_content

For import llms_txt to work, Python needs llms_txt to be an installed package on sys.path — it's not enough for tools/llms_txt/ to just exist as a directory in the repo, because nothing adds tools/ to sys.path by default. The old code manually did sys.path.insert(0, ...); the new code relies on the package being properly installed instead.

pyproject.toml declares where that package physically lives:

[tool.setuptools]
package-dir = {"" = "tools"}
packages = ["llms_txt"]

So pip install . (run from the repo root, where pyproject.toml is — that's what the bare . means) installs the tools/llms_txt/ directory as the top-level importable module llms_txt.

Why -e (editable) instead of a plain install

A normal pip install . copies the package files into your Python environment's site-packages. If you then edit tools/llms_txt/hooks.py, your installed copy is stale until you reinstall.

pip install -e . instead installs a pointer back to tools/llms_txt/ in the repo. Editing the source is picked up immediately — no reinstall needed. That matters here because this is source code actively being edited (docs build hooks), not a third-party dependency.

Why it's in requirements.txt (not just requirements-dev.txt)

hooks.py is required for every mkdocs build, not just for running tests — so whatever installs dependencies for a real doc build (local, CI, ReadTheDocs) also needs llms_txt importable. Since requirements-dev.txt does -r requirements.txt, adding it to the base file covers both.
```

tools/llms_txt/update_llmstxt_config.py
```diff
@@ -0,0 +1,131 @@
+#!/usr/bin/env python3
```
mnocon, line 1: Same here

adriendupuis (approved):

PR #3299, opened 2026-07-20
Fixed cli.ibexa.cloud

reithor:
- in theory it should have been cli.ibexa.cloud for a long time
- [cli.ibexa.co](http://cli.ibexa.co/) was only working because upsun had some "leftover"  DNS settings
- those DNS setting have been removed last week


mnocon (approved):
Thank you!

PR #3325, opened 2026-07-28
Fixed card link when using MkDocs development server

mnocon:
Target: 5.0, 6.0 + same in userdoc

Stopped always forcing the "https" scheme when generating the card links.

This changes nothing for production (where https is used), but always using https didn't work with MkDocs development server (mkdocs serve) which uses HTTP.

Example: https://ez-systems-developer-documentation--3325.com.readthedocs.build/en/3325/discounts/discounts/

### DIFF
--- a/main.py
+++ b/main.py
@@ -7,8 +7,8 @@ from mkdocs.structure.pages import Page
 from mkdocs.utils import meta
 from typing import List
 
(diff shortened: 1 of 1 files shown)


adriendupuis (approved):
🤩

PR #3341, opened 2026-08-05
Removed title and descritpion extraction from HTML

mnocon:
Target: 4.6, 5.0, 6.0, user doc

If you look at https://github.com/ibexa/documentation-developer/tags , you might see that a tag is created every day.

That's not supposed to happen, the tag should be created only when there's new content in the documentation.

The tags were created (incorrectly) because there was an actual difference in documentation content between them - turns out that our builds were not reproducible and relied on doc.ibexa.co access.

If requests to doc.ibexa.co succeeded: title and description were extracted, all good
If requests to doc.ibexa.co failed: they failed silently, title and description were set to empty

And we were not notified at all that it happens!

This has happened a couple of times when building the "Composer package" version of the doc (Cloudflare likes to respond with 429 for requests to doc.ibexa.co coming from GitHub Actions) and caused the content to fluctuate every day.

TBH it's possible that it was broken on the live site and we've never noticied.
And making requests whenever `mkdocs build` is run feels weird.

This PR:

1) Adds the titles and descriptions as they were extracted every time
2) Removes the "Extract title and description" from  HTML pages (both external and internal)  - it still works for Markdown files

### DIFF
--- a/docs/commerce/shopping_list/shopping_list.md
+++ b/docs/commerce/shopping_list/shopping_list.md
@@ -21,7 +21,7 @@ A user can have several shopping lists, including a default one named "My Wishli
 [[= cards([
 "commerce/shopping_list/shopping_list_design",
 "commerce/shopping_list/shopping_list_api",
-("api/php_api/php_api_reference/namespaces/ibexa-contracts-shoppinglist.html", "PHP API Reference", "<code>Ibexa\Contracts\ShoppingList</code>"),
+("api/php_api/php_api_reference/namespaces/ibexa-contracts-shoppinglist.html", "PHP API Reference", "<code>Ibexa\\Contracts\\ShoppingList</code>"),
 "api/event_reference/shopping_list_events/",
 "search/shopping_list_search_reference/shopping_list_criteria/",
 "search/shopping_list_search_reference/shopping_list_sort_clauses/",
--- a/main.py
+++ b/main.py
@@ -1,8 +1,6 @@
 import os
 import pprint
 import re
-import urllib.error
-import urllib.request
 from mkdocs.structure.pages import Page
 from mkdocs.utils import meta
 from typing import List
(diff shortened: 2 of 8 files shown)


docs/commerce/shopping_list/shopping_list.md
```diff
@@ -21,7 +21,7 @@ A user can have several shopping lists, including a default one named "My Wishli
 [[= cards([
 "commerce/shopping_list/shopping_list_design",
 "commerce/shopping_list/shopping_list_api",
-("api/php_api/php_api_reference/namespaces/ibexa-contracts-shoppinglist.html", "PHP API Reference", "<code>Ibexa\Contracts\ShoppingList</code>"),
+("api/php_api/php_api_reference/namespaces/ibexa-contracts-shoppinglist.html", "PHP API Reference", "<code>Ibexa\\Contracts\\ShoppingList</code>"),
```
mnocon, line 24: The `\C` was responsible for:

`INFO    -  DeprecationWarning: invalid escape sequence '\C'`

visible in the builds (https://github.com/ibexa/documentation-developer/actions/runs/30985847130/job/92240292415)

So it's not related to this PR, but small enoguh that IMHO can be included here.

adriendupuis: It reminds me the battle against `\E` ESC character in preview_comment.yaml

main.py
```diff
@@ -104,63 +101,31 @@ def cards(pages, columns=1, style="cards", force_version=False):
...
-                html = True
-                content = open("docs/%s" % path, "r").read()
-                page = _absolute_page_url(scheme, site, language, version, page)
+            is_external = path.startswith(("http://", "https://"))
+            if is_external or path.endswith(".html"):
+                if not (custom_title and custom_description):
+                    raise ValueError(
+                        "cards(): card %r must supply a title and a description - only "
+                        "Markdown pages carry front matter to read them from." % path
+                    )
```
adriendupuis, line 110: Nice

adriendupuis (approved):

julitafalcondusza (approved):

PR #3342, opened 2026-08-05
Bumped Composer version

mnocon:
Target: 4.6, 5.0, 6.0

Feedback received on Slack - we should promote the latest Composer version (2.10 is available)

### DIFF
--- a/docs/getting_started/requirements.md
+++ b/docs/getting_started/requirements.md
@@ -251,11 +251,17 @@ For production setups it's recommended that you use Varnish/Fastly, Redis/Valkey
 
 === "[[= product_name =]] v5.0"
 
-    - Composer: recent 2.8 version
+    - Composer: 2.8+
+
+    If you see a "+" next to the product version, it indicates a recommended version or higher within the same major release.
+    For example, "1.18+" means any 1.x version equal to or higher than 1.18, but not 2.x.
 
 === "[[= product_name =]] v4.6"
 
-    - Composer: recent 2.7 version
+    - Composer: 2.7+
+
+    If you see a "+" next to the product version, it indicates a recommended version or higher within the same major release.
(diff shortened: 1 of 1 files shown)


docs/getting_started/requirements.md
```diff
@@ -251,11 +251,17 @@ For production setups it's recommended that you use Varnish/Fastly, Redis/Valkey
 
 === "[[= product_name =]] v5.0"
 
-    - Composer: recent 2.8 version
+    - Composer: 2.8+
+
+    If you see a "+" next to the product version, it indicates a recommended version or higher within the same major release.
+    For example, "1.18+" means any 1.x version equal to or higher than 1.18, but not 2.x.
```
adriendupuis, line 257: I know this sentence is all over the place and if I ask for change here, it should be updated elsewhere, sorry for that. Maybe an included snippet like docs/snippets/update/db/db_backup_warning.md and others?

This "recommended version" then "or higher" bothers me. It might sounds like higher is possible but not the one recommended.

```suggestion
    If you see a "+" next to the product version, it indicates a recommended version range, a minimal version or higher within the same major release.
    The higher is more recommended.
    For example, "1.18+" means any 1.x version equal to or higher than 1.18, but not 2.x.
    So the latest 1.x is recommended but 1.18 fits the requirement.
```

adriendupuis (approved):
It's already OK and my comment is a bit out of scope so it could be for another time.
So I approve while letting my thought for a general improvement of this sentence that is used in a lot of other places in this page.

julitafalcondusza (approved):

docs/getting_started/requirements.md
mnocon: Let's keep it like it is currently, and revisit this when we release 6.0 - as we can rework the requirements page for that release

PR #3378, opened 2026-09-02
First batch of removal for SaaS

mnocon:
Target: SaaS only

Removes:

- Content related to Ibexa Commerce
- Release notes
- Update guides
- Requirements page
- Ibexa Cloud
- Notification types (won't be configurable on SaaS)


adriendupuis (approved):
It works as a first batch. #3381 removes things that are updated here but didn't worth keeing (like docs/resources/phpstorm_plugin.md or docs/resources/contributing/).
The Experience flavor and its pills should be removed in the future as SaaS comes into only one flavor, its own SaaS flavor. See you in next batches.

PR #3404, opened 2026-09-08
Removed PHP API reference

mnocon:
Removed PHP API reference files.

Note: I'm removing these files to show (in the next PR) that they can be safely removed and that the build is not broken.

Changes to the PHP API Reference script that removes generating them will be extracted to the next PR.


adriendupuis (approved):
I found nothing among the deleted files that should have been kept.

PR #3405, opened 2026-09-08
More deletions

mnocon:
This PR doesn't build - it contains only deletions.
It's on purpose, to make it easier to review.

Deletions include:
1) Code samples, PHP API reference files, and the tooling around them (total of 509 files)
2) Customization articles, event references, extend X articles, GraphQL API (GraphQL won't be avaialble on SaaS)

To review them, I recommend looking at the `docs` dir in https://diffshub.com/ibexa/documentation-developer/pull/3405 - it displays the content that was deleted by default

2) Tutorials - they will have to be reworked for SaaS, with Headless Page Builder being one of them. In their current state, they're not usable for SaaS.


docs/administration/back_office/add_user_setting.md
mnocon, line 1: This part is not extensible on SaaS

adriendupuis (approved):
I found nothing among the deleted files that should have been kept.

